PR #751, opened 2016-02-17
MERC-251 update card-image to use product preset size

sherrybc:
The cause of this issue is from `card` css class which expands the image based on browser size, such as 265x456 px, but the current `gallery` preset is at 190x250 px, hence the blurry image. 

One way to fix this issue is by using `product` preset which is bigger than 265x456 px, at 500x659 px. 

@mcampa @mickr @bc-ejoe

### DIFF
--- a/templates/components/products/card.html
+++ b/templates/components/products/card.html
@@ -1,7 +1,7 @@
 <article class="card {{#if alternate}}card--alternate{{/if}}">
     <figure class="card-figure">
         <a href="{{url}}">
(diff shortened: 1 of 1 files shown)


bc-julianleong:
Hey. Isn't there a side-effect to this. Could we hold this PR while I chat with Eddie? 
Given that we have released the Image Resizing functionality in 1.3.1, this issue is also solved by that.

bc-julianleong:
good to go

PR #765, opened 2016-03-09
MERC-404 Fix dropdown menus

mcampa:
@hegrec @mickr

### DIFF
--- a/assets/scss/components/foundation/dropdown/_dropdown.scss
+++ b/assets/scss/components/foundation/dropdown/_dropdown.scss
@@ -9,12 +9,15 @@
     position: absolute;
 }
 
-.dropdown-menu.is-loading {
-
-    .loadingOverlay {
+.dropdown-menu {
+    &.is-loading .loadingOverlay {
         background: none;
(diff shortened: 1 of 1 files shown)


assets/scss/components/foundation/dropdown/_dropdown.scss
```diff
@@ -18,6 +18,11 @@
     }
 }
 
+.dropdown-menu.is-open {
```
hegrec: Can you move the is-open inside the block like `&.is-open {`

mcampa: I'm living as is because if I do this:
```scss
.dropdown-menu
    &.is-open {
         display: block;
    }
}

the scss linter starts whining with
```
  21: [W] MergeableSelector:  Merge rule `.dropdown-menu` with rule on line 12
```
And if I merge it with L2 I'll have to add one more nested level and disable `NestingDepth` in scss-lint

I don't like having a limit of 2 on `NestingDepth`. Any reasons for that?

hegrec:
:+1:

PR #768, opened 2016-03-10
 MERC-341 fix state validation in shipping estimator

sherrybc:
- when default country does not have states, show input field 
- update state validation; only validate when it is a dropdown

@mcampa @mickr @bc-ejoe

### DIFF
--- a/assets/js/theme/cart/shipping-estimator.js
+++ b/assets/js/theme/cart/shipping-estimator.js
@@ -2,6 +2,7 @@ import $ from 'jquery';
 import stateCountry from '../common/state-country';
 import nod from '../common/nod';
 import utils from 'bigcommerce/stencil-utils';
+import { classifyForm, Validators } from '../common/form-utils';
 
 export default class ShippingEstimator {
     constructor($element) {
@@ -35,6 +36,7 @@ export default class ShippingEstimator {
         });
 
         this.bindValidation();
+        this.bindStateValidation();
         this.bindUPSRates();
     }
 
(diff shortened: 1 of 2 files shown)


assets/js/theme/cart/shipping-estimator.js
```diff
@@ -73,6 +78,34 @@ export default class ShippingEstimator {
         ]);
     }
 
+    setStateValidation() {
+
+        let $last;
+
+        stateCountry(this.$state, this.context, { useIdForStates: true }, (err, field) => {
+            if (err) {
+                throw new Error(err);
```
mcampa: Is `err` a string? if not, you should just throw err

mickr:
:+1:

templates/components/cart/shipping-estimator.html
```diff
@@ -34,14 +34,18 @@
...
             <dd class="estimator-form-input">
-                <select class="form-select" id="shipping-state" name="shipping-state" data-field-type="State">
-                    <option>{{lang 'cart.shipping_estimator.select_a_state'}}</option>
-                    {{#each states}}
-                        <option value="{{id}}" {{#if selected}}selected="selected"{{/if}}>
-                            {{name}}
-                        </option>
-                    {{/each}}
-                </select>
+                {{#if selected_state}}
```
mcampa: I think you need to check for `states.length` instead of `selected_state`

mcampa: Yeah I just check the context. `selected_state` could be null while `states` is an array with all the states

mcampa:
:+1:  LGTM

PR #769, opened 2016-03-11
MERC-409: Enable theme specific builds

hegrec:
### What
Removing JSPM in favor of Webpack. Also, enable developers to utilize their own stencil-cli integrated build process if Webpack is not a favored tool for the developer of a theme.

### Why
* Remove the need to install JSPM as a global tool and requiring `jspm install`. 
* Speed up the page load time by downloading a single bundle rather than many files over multiple ajax connections.
* Enable the build to be completely theme driven and not forced by the platform.

### Testing
Fetch and checkout this branch and then `npm install`
Clone stencil utils and checkout this branch https://github.com/bigcommerce/stencil-utils/pull/56
`npm link` stencil-utils and `npm link @bigcommerce/stencil-utils` here in the theme
Fetch and checkout this branch https://github.com/bigcommerce/stencil-cli/pull/201
`npm install` within stencil-cli
`stencil start` and develop like normally.

See `stencil.conf.js` for the development and production build steps.


webpack.conf.js
```diff
@@ -0,0 +1,27 @@
...
+            }
+        ]
+    },
+    plugins: [
+        new webpack.ProvidePlugin({
+            $: "jquery",
+            jQuery: "jquery",
+            'window.jQuery': 'jquery'
+        }),
+        new webpack.optimize.DedupePlugin()
```
davidchin: Cool plugin :) Do we want to only run this for prod build? In the doc (https://webpack.github.io/docs/list-of-plugins.html), it says 'Note: Don’t use it in watch mode. Only for production builds'. Probably because it slows down the build in watch mode?

hegrec: Good callout, I'll move it to a production bundling step.

mickr: This looks great so far! I'll look more tomorrow

stencil.conf.js
```diff
@@ -0,0 +1,66 @@
...
+webpackConfig.entry = './assets/js/app.js';
+webpackConfig.output = {
+    path: __dirname + '/assets/js',
+    filename: 'bundle.js'
+};
+
+/**
+    Hook into the stencil-cli browsersync instance for rapid development of themes.
+    Watch any custom files and trigger a rebuild
+*/
```
mcampa: Can you add a description to the parameter `Bs`?
```
* @param browser-sync ...
```

hegrec: fixed

.eslintignore
```diff
@@ -0,0 +1 @@
+/assets/js/bundle.js
```
mickr: Can you add the config files here as well, ESLint is operating under ES2016 guidelines

hegrec: Done

stencil.conf.js
```diff
@@ -0,0 +1,75 @@
...
+var bundleLocation = '/assets/js/bundle.js';
+
+webpackConfig.context = __dirname;
+webpackConfig.entry = './assets/js/app.js';
+webpackConfig.output = {
+    path: __dirname + '/assets/js',
+    filename: 'bundle.js'
+};
+
+/**
```
bc-jstoffan: Looks like a stray comment block.

webpack.conf.js
```diff
@@ -0,0 +1,26 @@
+var webpack = require('webpack');
+
+module.exports = {
+    devtool: "inline-source-map",
```
bc-jstoffan: It'd be cool to (some day) have two bundles for production, one that includes sourcemaps and one that doesn't, that you could switch between using a query parameter or something. It'd make debugging live stores a lot easier. Not really in the scope of this PR, though.

templates/layout/base.html
```diff
@@ -32,25 +32,12 @@
...
-            <script src="{{cdn '/assets/js/bundle.js'}}"></script>
-            <script>
-                startupApp(stencilBootstrap);
-            </script>
-        {{else}}
-            <script src="{{cdn '/assets/jspm_packages/system.js'}}"></script>
-            <script src="{{cdn '/assets/config.js'}}"></script>
-            <script src="{{cdn '/assets/js/dependency-bundle.js'}}"></script>
-            <script>
-                System.import("js/app").then(function(){ startupApp(stencilBootstrap)});
```
bc-jstoffan: JSPM and SystemJS serve two distinct purposes. SystemJS is basically a client-side module loader polyfill. It'll just load whatever dependency tree you give it, deferring execution however you define. The tree could be every file (like the old JSPM build) or modularized bundles (common, app-specific, etc.). Removing it here should be fine, though, since themes are simple enough that dependency sharing and module-specific cache busting probably aren't necessary.

hegrec:
Build here is failing because it depends on a new version of stencil-utils.

hegrec:
Anything before I merge this?

mickr:
:+1: Do we need QE to test this?

PR #773, opened 2016-03-16
MERC-363 Show "init" image from a product or option rule

mcampa:
If an option has an image change rule attached to it and it's set as the default option, the image change rule does not trigger upon loading of the product details page. 

https://jira.bigcommerce.com/browse/MERC-363
@bc-ejoe @sherrybc @mickr @hegrec


mickr:
:+1:

PR #777, opened 2016-03-21
MERC-111 - Fix text cutoff in search dropdown

mickr:
MERC-111 - Fix text cutoff in search dropdown

* Unset height rule for the dropdown search bar, the results with going with the default height instead of being explicitly sized.

### DIFF
--- a/assets/scss/components/stencil/navUser/_navUser.scss
+++ b/assets/scss/components/stencil/navUser/_navUser.scss
@@ -254,6 +254,7 @@
 
     .form-input {
         font-size: fontSize("small");
+        height: unset;
     }
 
     .productGrid {


mickr:
Before using Firefox
screen shot 2016-03-21 at 3 16 49 pm

After using Firefox
screen shot 2016-03-21 at 3 17 06 pm

@bc-chris-roper @bc-miko-ademagic

hegrec:
:+1:

indominus:
I'm cool with this, but make sure that it's tested in all major browsers including IE. I have a feeling that it would've obviously been set for a reason, and typically form elements are annoying little dudes cross-browser. One of the browsers might need an explicit height set.

mickr:
I tested on Chrome, Safari, and Firefox... I don't currently have IE but maybe @bc-arun can test?

bc-chris-roper:
What that issue looks like based on the screenshot isn't that the height is the problem on its own, but that the specified height combined with the font-size and padding of the box doesn't leave enough space for the text.

So another approach to fix it would be to just increase the height to what it ideally should be.

bc-arun:
@mickr I verified it on IE 13 & 12. Looks good.
screen shot 2016-03-22 at 10 14 19 am

mickr:
@bc-arun can you give this the green heart?

bc-arun:
:green_heart:

PR #778, opened 2016-03-21
MERC-116 - Price fields should be of number type

mickr:
MERC-116 - Price fields should be of number type

* Change field type to `number` this will prevent unwanted strings from being submitted for price.


hegrec:
Will this support floats?

mickr:
It will with the `step=any` attribute

hegrec:
:+1:

mickr:
Really wish they would remove the close and comment button

PR #785, opened 2016-04-06
MERC-108 - Add 'X' to close full screen quick search dropdown.

mickr:
MERC-108 - Add 'X' to close full screen quick search dropdown.

#### Why
Once there are search results it is difficult and unintuitive to close the search results drop down.

#### How
Just added the standard 'X' that we use in modals to display when there are search results. This is bound to the listener with the `data-drop-down-close` data attribute.

### DIFF
--- a/assets/js/theme/global/stencil-dropdown.js
+++ b/assets/js/theme/global/stencil-dropdown.js
@@ -66,6 +66,9 @@ export default class StencilDropdown {
             modalOpened = true;
         }).on('close.fndtn.reveal', '[data-reveal]', () => {
             modalOpened = false;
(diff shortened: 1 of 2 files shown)


mickr:
screen shot 2016-04-06 at 10 58 39 am

mickr:
Not sure why travis is failing... `grunt eslint` passes locally.

hegrec:
We have a strange use of `stencilDropdown`. I've changed it in my build PR so that it uses a const

mickr:
@hegrec I have rebased this PR against master now.

hegrec:
:+1: LGTM

PR #787, opened 2016-04-07
MERC-456 Update the continue url to point to a new specific url instead of home

lord2800:
## What

Change the url for the "continue" button to point to a specific url instead of to the home url.

## Why

Currently the url doesn't take redirects during account signup into account, which means adding to wishlist while not signed in doesn't work when you create an account (among other things).

Ping @hegrec @mickr @mcampa


mickr:
:+1:

PR #788, opened 2016-04-08
MERC-581 - adding support for hidden throttle token to product review modal

bc-bfenton:
Without this, the throttle token won't be sent to BC App, and it will consider all "throttled" requests to be invalid.


hegrec:
Is the data already being provided?

bc-bfenton:
@hegrec unless I'm misreading things the data should be provided here https://github.com/bigcommerce/bigcommerce/blob/master/includes/resources/ProductReviews.php#L203

hegrec:
:+1:

PR #790, opened 2016-04-13
MERC-454 Configurable Fields on Cart and Cart Preview

mickr:
MERC-454 Configurable Fields on Cart and Cart Preview

* Configurable fields were not visible unless there are options on the cart page
* Configurable fields are now visible in the cart preview modal.


mickr:
@sherrybc

hegrec:
:+1:

PR #795, opened 2016-05-03
MERC-410: Improve gallery image size

hegrec:
## What
Use the gallery image size and also bump it up slightly to prevent blurry images

## Why
The `product` preset is much larger than the css limited size of card images. They are never larger than 300px across all responsive widths so this appears to be an appropriate size.

### After

@sherrybc @bc-ejoe @mickr @mcampa

### DIFF
--- a/config.json
+++ b/config.json
@@ -75,8 +75,8 @@
         "height": 659
       },
       "gallery": {
-        "width": 190,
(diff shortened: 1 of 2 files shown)


mcampa:
👍

PR #796, opened 2016-05-05
MERC-662: Transpile Object.assign

hegrec:
## What
Babel 6 does not transpile object.assign automatically so we need to include this plugin.

@mcampa @bc-ejoe @mickr @sherrybc


mcampa:
LGTM 👍

PR #797, opened 2016-05-05
MERC-616, MERC-608 logo and product image resize functionality

sherrybc:
update schema.json for Logo
update schema.json for product images
add subtext about Desktop sizing
update the display order of the settings in TE 

@mcampa @mickr @bc-ejoe 

MERC-616 also has changes in stencil-paper and stencil-editor

### DIFF
--- a/config.json
+++ b/config.json
@@ -61,28 +61,13 @@
     "show_accept_visa": false,
     "product_list_display_mode": "grid",
     "logo-position": "center",
-    "_images": {
-      "logo": {
-        "width": 250,
-        "height": 100
-      },
-      "thumb": {
-        "width": 100,
-        "height": 100
-      },
-      "product": {
-        "width": 500,
-        "height": 659
-      },
-      "gallery": {
-        "width": 300,
-        "height": 300
-      },
-      "zoom": {
-        "width": 1280,
-        "height": 1280
-      }
-    },
+    "logo_size": "250x100",
+    "gallery_size": "300x300",
+    "productgallery_size": "300x300",
+    "product_size": "500x659",
+    "productthumb_size": "100x100",
+    "thumb_size": "100x100",
+    "zoom_size": "1280x1280",
     "default_image_brand": "/assets/img/BrandDefault.gif",
     "default_image_product": "/assets/img/ProductDefault.gif",
     "default_image_gift_certificate": "/assets/img/GiftCertificate.png",
--- a/schema.json
+++ b/schema.json
@@ -670,6 +670,27 @@
     "name": "Logo",
     "settings": [
       {
+        "type": "imageDimension",
+        "label": "Logo size",
+        "id": "logo_size",
+        "description": "Please specify a maximum image size for desktop display. We'll automatically scale down the image for smaller devices.",
+        "force_reload": true,
+        "options": [
+          {
+            "value": "original",
+            "label": "Original (as uploaded)"
+          },
+          {
+            "value": "250x100",
+            "label": "Optimized for theme"
+          },
+          {
+            "value": "custom",
+            "label": "Specify dimensions"
+          }
+        ]
+      },
+      {
         "type": "select",
         "label": "Logo position",
         "id": "logo-position",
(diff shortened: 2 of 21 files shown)


config.json
```diff
@@ -61,6 +61,8 @@
     "show_accept_visa": false,
     "product_list_display_mode": "grid",
     "logo-position": "center",
+    "logoSize": "250x100",
+    "productImageSize": "300x300",
```
hegrec: Does this work by default? I notice you are using these in the getImage helper, but the values here do not exist in `_images` as an object with `width` and `height`

sherrybc: yes will tag you in the PR (;

mickr:
Are there any dependent PR's?

schema.json
```diff
@@ -1344,6 +1375,183 @@
...
+      {
+        "type": "paragraph",
+        "content": "Please specify a maximum image size for desktop display. We'll automatically scale down the image for smaller devices.",
+        "hideExpression": "model.productSize != 'product'"
+      },
+      {
+        "type": "select",
+        "label": "Main product images",
+        "id": "productSize",
+        "force_reload": true,
```
mcampa: `force_reload` is not needed here. Stencil cli scan all templates and automatically injects the `force_reload` for you

schema.json
```diff
@@ -1344,6 +1375,183 @@
...
+            "label": "Specify dimensions"
+          }
+        ]
+      },
+      {
+        "type": "number",
+        "id": "_images.product.width",
+        "force_reload": true,
+        "label": "Max width",
+        "className": "formly-field-dimensions",
```
mcampa: I'm not sure why designers need to now what class names are used in theme editor. I think we should remove this attribute and use one style in theme-editor

sherrybc:
@mickr yes depends on stencil-editor (still open) and paper (merged)

config.json
```diff
@@ -61,6 +61,11 @@
     "show_accept_visa": false,
     "product_list_display_mode": "grid",
     "logo-position": "center",
+    "logoSize": "250x100",
```
mcampa: We are using snake_case for all our settings, can you change that to keep consistency?

sherrybc: angular doesn't recognize snake_case

config.json
```diff
@@ -78,6 +87,10 @@
         "width": 300,
         "height": 300
       },
+      "productGallery": {
```
hegrec: You shouldn't need to define `productGallery` with the same size as `gallery`

schema.json
sherrybc: ideally, I would like width x height to be one type, this is definitely hacky, I can move the classNames to theme-editor for now, but it's just as hacky, we should definitely do a tech debt on this, but fyi this needs be in asap.

config.json
sherrybc: because `gallery` applies to more than just products, but since we are currently allowing users to change product gallery size, but we made a new one just for products.

sherrybc: post images uses `gallery` for example

sherrybc:
one more go 💃  @mcampa @mickr @bc-ejoe

sherrybc:
The reason why `_images` is changed to dot notation is because nested key does not work with custom templates, one of the work around is to use the model property, but that would change the way stencil-editor is structured, see https://github.com/formly-js/angular-formly/blob/v7.3.2/src/run/formlyNgModelAttrsManipulator.js#L46-L50

sherrybc:
@bc-jstoffan @mcampa @mickr @bc-ejoe @bc-ranji

config.json
```diff
@@ -61,28 +61,13 @@
...
-        "height": 1280
-      }
-    },
+    "logoSize": "250x100",
+    "gallerySize": "300x300",
+    "productGallerySize": "300x300",
+    "productSize": "500x659",
+    "productThumbSize": "100x100",
+    "thumbSize": "100x100",
+    "zoomSize": "1280x1280",
```
mcampa: I still think we need to change this to snake_case to be consistent

hegrec:
👍  Nice, it's very clean this way.

bc-ejoe:
👍

PR #799, opened 2016-05-11
MERC-691: revert back to slick-carousel v1.5.5

bc-ejoe:
## What?

Fix for missing previous slide arrow on the carousel by reverting back to slick-carousel v1.5.5.  

## Why?

Prior to 328cf24, v1.5.5 was being pulled in.  v1.5.6 of the lib introduced a change to the dom location of the arrow, causing the z-index to be lower than the carousel image.

## Testing/Proof

before:
screen shot 2016-05-11 at 1 19 42 am

after:
screen shot 2016-05-11 at 1 19 16 am

@hegrec @mickr @sherrybc @mcampa

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "lodash": "^3.5.0",
     "nod-validate": "^2.0.12",
     "pace": "hubspot/pace",
-    "slick-carousel": "^1.5.9",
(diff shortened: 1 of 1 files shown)


hegrec:
👍  LGTM

PR #806, opened 2016-05-21
MERC-724: Use es2015 loose mode for ie10 scripts compatibility

hegrec:
Run using babel es2015 loose preset to stick to more traditional es5 style bundles.
This is caused by yet another change from babel 5 to babel 6 where babel 6 assumes it can follow es6 more closely, for our use case, this will not work because we cater to many customers in different environments around the world.

Babel Loose Mode: http://www.2ality.com/2015/12/babel6-loose-mode.html

I tested this in browserstack against IE10, IE11, MS Edge, Safari, and Chrome for production bundled and cli driven themes and all seem to be working great.

@bc-ejoe @mickr @mcampa @sherrybc @bc-ranji


mcampa:
👍

bc-ejoe:
can you open this PR against the v1.2.1 branch?

PR #809, opened 2016-05-24
MERC-721 - Consistent `required` labeling

mickr:
MERC-721 - Consistent `required` labeling

* Required fields for options should be clearly marked as required, and consistent.


mickr:
@bc-ejoe

hegrec:
👍

PR #819, opened 2016-06-30
MERC-888 update orders and cart product images

sherrybc:
screen shot 2016-06-29 at 7 23 03 pm
orders

screen shot 2016-06-29 at 7 29 59 pm
orders detail

screen shot 2016-06-29 at 5 30 43 pm
cart

screen shot 2016-06-29 at 5 23 08 pm
cart preview

@mcampa @mickr @bc-ejoe @deepa2083

### DIFF
--- a/assets/scss/components/stencil/account/_account.scss
+++ b/assets/scss/components/stencil/account/_account.scss
@@ -41,6 +41,10 @@
     img {
         width: 100%;
     }
+
+    .account-product-image {
+        width: auto;
+    }
 }
 
 .account-listShipping {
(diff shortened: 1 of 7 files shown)


deepa2083:
The default product thumbnail size is set to 10 x 10 now instead of optimized to theme. Please revert that 
screen shot 2016-06-30 at 10 21 23 am

mcampa:
`scss-lint` is whining because `SelectorDepth` & `NestingDepth` max depth are too strict. 
We should increase the`max_depth` to 3. What do you think @sherrybc @hegrec?

hegrec:
I think it is generally better for the compiler when you have specific class names that you can bring up a level rather than having a deep tree.

mcampa:
👍

PR #820, opened 2016-07-08
MERC-893: update the view in quick view modal if product has options

bc-ejoe:
## What?

Fix for 'Add to Cart' button in quick view modal being disabled.  

## Why?

`window.BCData` does not get set for the quick view modal, so the call to `updateView` would disable the 'Add to Cart' button.  Updated so that product attributes are updated when the product has options, and then the view gets updated.  (related to MERC-678)

## Testing/Proof

screen shot 2016-07-08 at 4 08 55 pm
screen shot 2016-07-08 at 11 25 55 am

@mcampa @hegrec @sherrybc @bc-ranji

### DIFF
--- a/assets/js/theme/common/product-details.js
+++ b/assets/js/theme/common/product-details.js
@@ -35,8 +35,23 @@ export default class Product {
         this.imageGallery.init();
         this.listenQuantityChange();
         this.initRadioAttributes();
-        this.updateProductAttributes(productAttributesData);
-        this.updateView(productAttributesData);
+
+        const $form = $('form[data-cart-item-add]', $scope);
+        const hasOptions = $('[data-product-option-change]', $form).html().trim().length;
+
+        // Update product attributes. If we're in quick view and the product has options, then also update the initial view in case items are oos
+        if (_.isEmpty(productAttributesData) && hasOptions) {
+            const $productId = $('[name="product_id"]', $form).val();
+
+            utils.api.productAttributes.optionChange($productId, $form.serialize(), (err, response) => {
+                const attributesData = response.data || {};
+
+                this.updateProductAttributes(attributesData);
+                this.updateView(attributesData);
+            });
+        } else {
+            this.updateProductAttributes(productAttributesData);
+        }
 
         previewModal = modalFactory('#previewModal')[0];
         productSingleton = this;


assets/js/theme/common/product-details.js
```diff
@@ -35,8 +35,24 @@ export default class Product {
         this.imageGallery.init();
         this.listenQuantityChange();
         this.initRadioAttributes();
-        this.updateProductAttributes(productAttributesData);
-        this.updateView(productAttributesData);
+
+        const $form = $('form[data-cart-item-add]', $scope);
+        const $hasOptions = $('[data-product-option-change]', $form).html().trim().length;
```
mcampa: This is is a number not a jquery object. Remove the `$`

assets/js/theme/common/product-details.js
```diff
@@ -35,8 +35,24 @@ export default class Product {
...
+        const $isQuickView = $scope.hasClass('quickView');
```
mcampa: same here

assets/js/theme/common/product-details.js
```diff
@@ -35,8 +35,24 @@ export default class Product {
...
+
+        // Update product attributes. If we're in quick view and the product has options, then also update the initial view in case items are oos
+        if ($isQuickView && $hasOptions && window.FormData !== undefined) {
```
mcampa: Why not just check if `window.BCData.product_attributes` is defined or not?

bc-ejoe: I'll change to checking if `window.BCData.product_attributes` is undefined, but will still need to check if there are options because the `product-attributes` endpoint only returns data if the product has options.

bc-ejoe:
👀  🙏

sherrybc:
👍

assets/js/theme/common/product-details.js
```diff
@@ -35,8 +35,23 @@ export default class Product {
...
         this.listenQuantityChange();
         this.initRadioAttributes();
-        this.updateProductAttributes(productAttributesData);
-        this.updateView(productAttributesData);
+
+        const $form = $('form[data-cart-item-add]', $scope);
+        const hasOptions = $('[data-product-option-change]', $form).html().trim().length;
+
+        // Update product attributes. If we're in quick view and the product has options, then also update the initial view in case items are oos
+        if (_.isEmpty(productAttributesData) && hasOptions && window.FormData !== undefined) {
```
mcampa: Do we need `window.FormData !== undefined` check?

mcampa: seems like jquery.serialize() does not use FormData. So you can remove that check

bc-ejoe: done

mcampa:
👍

PR #821, opened 2016-07-22
MERC-930 update gift certificate img

sherrybc:
temporary fix for enlarged gift certificate image in IE10, 11, and firefox 41

@mcampa @bc-ong @bc-miko-ademagic

### DIFF
--- a/templates/components/cart/content.html
+++ b/templates/components/cart/content.html
@@ -12,7 +12,7 @@
             <tr class="cart-item" data-item-row>
                 <td class="cart-item-block cart-item-figure">
                     {{#if type '==' 'GiftCertificate'}}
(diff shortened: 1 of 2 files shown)


bc-ong:
:+1:

templates/components/cart/content.html
```diff
@@ -133,7 +133,6 @@ <h4 class="cart-item-name"><a href="{{url}}">{{name}}</a></h4>
                                 <svg><use xlink:href="#icon-close"></use></svg>
                             </a>
                         {{/if}}
-                        <span class="cart-remove cart-remove--spacer"></span>
```
sherrybc: not sure why that is there but it creates an extra space which misaligns with the other items

PR #833, opened 2016-08-11
MERC-999 update feature product style

sherrybc:
`column-count` breaks layout to top to bottom arrangement rather than left to right; use pattern-lab columns 

@bc-ong

### DIFF
--- a/assets/scss/layouts/products/_productGrid.scss
+++ b/assets/scss/layouts/products/_productGrid.scss
@@ -143,7 +143,12 @@
 
     .csscolumns & {
         @include breakpoint("medium") {
-            column-count: 3;
+            @include grid-column(12);
+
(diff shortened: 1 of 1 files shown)


assets/scss/layouts/products/_productGrid.scss
```diff
@@ -143,7 +143,13 @@
 
     .csscolumns & {
         @include breakpoint("medium") {
-            column-count: 3;
+            @include grid-column(12);
+        }
+
+        .product {
```
bc-ong: Would it make sense to include this in the breakpoint block started online 145?

sherrybc:
@bc-ong one more go 💃

bc-ong:
:+1:

PR #837, opened 2016-08-18
CHECKOUT-919: add main content uco customizations

jordanwink201:
ability to change background color
Ability to update Typography via existing style editor features under 'Backgrounds & Fonts"
*Page Text	
*Heading	
*Heading Text
update form fields color

screen shot 2016-08-23 at 6 57 07 pm

@bigcommerce/checkout @sherrybc @mcampa @bc-ong

### DIFF
--- a/assets/scss/checkout.scss
+++ b/assets/scss/checkout.scss
@@ -1,3 +1,73 @@
 // =============================================================================
 // Stencil Checkout  - Customize the checkout experience
 // =============================================================================
+
+// Main content
+.ng-checkout-container {
+    background-color: stencilColor("uco_main_backgroundColor");
+}
+
+// Order Summary
+.cart {
+    background-color: stencilColor("uco_cart_backgroundColor");
+}
+
+// Typography
+
+$fontFamily-headings:           stencilFontFamily("uco_headings_font"), Arial, Helvetica, sans-serif;
+$fontFamily-sans:               stencilFontFamily("uco_body_font"), Arial, Helvetica, sans-serif;
+
+// Headings
+h1,
+h2,
+h3,
+h4,
+h5,
+h6 {
+    font-family: $fontFamily-headings;
+}
+
+// Body
+body {
+    font-family: $fontFamily-sans;
+}
+
+// Links
+a {
+    color: stencilColor("uco_links_color");
+}
+
+// Buttons
+.button {
+    background-color: stencilColor("uco_buttons_color");
+}
+
+// Errors
+.form-field--error {
+
+    .form-input,
+    .form-select,
+    .form-checkbox + .form-label::before,
+    .form-radio + .form-label::before,
+    .form-prefixPostfix-label {
+        border-color: stencilColor("uco_errors_color");
+    }
+
+    .form-field-error,
+    .form-inlineMessage {
+        color: stencilColor("uco_errors_color");
+    }
+
+    // Depth need to style svg with validation state
+    // scss-lint:disable SelectorDepth
+    .form-input-indicator > svg {
+        fill: stencilColor("uco_errors_color");
+    }
+    // scss-lint:enable SelectorDepth
+}
+
+// Form Fields
+.form-legend,
+.form-body {
+    background-color: stencilString("uco_form-field_backgroundColor");
+}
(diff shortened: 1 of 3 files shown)


PascalZajac:
Looks fine to me but Travis hates your SASS.

bc-edlou:
Please tag @bigcommerce/ui-platform for all HTML/CSS/SASS changes.

assets/scss/checkout.scss
```diff
@@ -1,3 +1,85 @@
 // =============================================================================
 // Stencil Checkout  - Customize the checkout experience
 // =============================================================================
+
+// Main content
+.ng-checkout-container {
```
bc-miko-ademagic: Are there any other classes to use for this container? `.ng-` tells me it's a class scoped for js. If there's another class to use, or a different container altogether, that'd probably be better.

assets/scss/checkout.scss
```diff
@@ -1,3 +1,85 @@
...
+}
+
+// Font Overrides
+$fontFamily-headings:           stencilFontFamily("uco_headings_font"), Arial, Helvetica, sans-serif;
+$fontFamily-sans:               stencilFontFamily("uco_body_font"), Arial, Helvetica, sans-serif;
+
+// Font Headings
+h2 {
+    font-family: $fontFamily-headings;
+}
```
bc-miko-ademagic: Can this live in your typography section of this page?

assets/scss/checkout.scss
```diff
@@ -1,3 +1,85 @@
...
+.orderedList-item {
+    &::before {
+        font-family: $fontFamily-headings;
+    }
+}
+
+// Font Body
+body {
+    font-family: $fontFamily-sans;
+}
```
bc-miko-ademagic: This can probably go to typography too. I'm just suggesting that because you're not using classes to style this stuff, so it might be good to group it together early on for structure and readability.

assets/scss/checkout.scss
```diff
@@ -1,3 +1,85 @@
...
+
+.accordion {
+    .accordion-navigation, dd {
+        // scss-lint:disable SelectorDepth
+        > a {
+            font-family: $fontFamily-sans;
+        }
```
bc-miko-ademagic: Is the nesting/selector depth necessary here? You're 3 levels deep when you get to the `a`, I'm wondering if you can avoid the second level altogether?

assets/scss/checkout.scss
```diff
@@ -1,3 +1,85 @@
...
+    }
+}
+
+// Font Body
+body {
+    font-family: $fontFamily-sans;
+}
+
+.accordion {
+    .accordion-navigation, dd {
```
bc-miko-ademagic: This needs to go to new lines.
Also why do you have to target the `dd` directly?

bc-miko-ademagic:
This is more of an abstract question related to the project, and not so much a comment on your code... 
Do we want to allow people to customize their checkout experience separately to their theme in the theme editor? Wondering if we should just be using variables available to the theme in things like typography so they just carry through to the checkout?

assets/scss/checkout.scss
jordanwink201: There are not other classes to use for this container.
```
<div class="ng-checkout-container" id="micro-app-ng-checkout">
  <div class="ng-checkout-view" data-ui-view></div>
</div>
```
a class can be added specifically for this purpose

jordanwink201:
@bc-miko-ademagic for this ticket we are allowing them to style UCO independently of their theme. Next step will be inheriting styles from their current theme into UCO, setting those as the default values, and then they can further customize just UCO if they wish.

assets/scss/checkout.scss
bc-miko-ademagic: Ah, right. I think it's meant to be an extension of `.appContainer`, it's just a badly formed class.

jordanwink201: @bc-miko-ademagic are we happy to use it for this purpose or would you suggest something else?

bc-miko-ademagic: No this is fine for now. Should be camelcased, `ngCheckout-container` but 🍹

bc-miko-ademagic:
Once you resolve your conflicts, 👍

PR #848, opened 2016-09-23
How to inject the product ids from the templates to js

mcampa:
@bc-AlyssNoland


assets/js/theme/category.js
```diff
@@ -4,6 +4,12 @@ import $ from 'jquery';
 import FacetedSearch from './common/faceted-search';
 
 export default class Category extends CatalogPage {
+    constructor(context) {
+        super();
+
+        console.log(context.productIds);
```
mcampa: Here `context.productIds` has an array of product ids

ex. 
```js
[93, 86, 88, 94, 97]
```

PR #849, opened 2016-09-27
Revert "MERC-1175 the fix for the dropdown bug using blueprint method"

mcampa:
Reverts bigcommerce/stencil#847

Reverting because @bc-ranji found a bug, and we are not fixing it until the next release


sherrybc:
👍

bc-ranji:
💚

PR #851, opened 2016-09-28
Use CSS to position maintenance banner

davidchin:
## What?
* Use CSS only to position maintenance banner.

## Why?
* Currently, we use both JS and CSS to position the maintenance banner. They conflict with each other. Therefore, the banner goes out of position when you scroll.
* For your reference, these are the original commits. As you can see, JS solution was implemented first. And then few days later, a different engineer introduced the CSS change.:
  * https://github.com/bigcommerce/stencil/commit/4ab7648598622ec8b0ac4a379e102f185b350b4c
  * https://github.com/bigcommerce/stencil/commit/44a0757fc05d6b0c7412ad5236ae7f97588e3f90
* Existing CSS here: https://github.com/bigcommerce/stencil/blob/master/assets/scss/components/stencil/maintenanceNotice/_maintenanceNotice.scss

## Testing / Proof
|Before|After|   
|---|---|
|screen shot 2016-09-28 at 12 26 22 pm|screen shot 2016-09-28 at 12 26 39 pm|

@mcampa @sherrybc @icatalina @PascalZajac

### DIFF
--- a/assets/js/theme/global/maintenanceMode.js
+++ b/assets/js/theme/global/maintenanceMode.js
@@ -8,8 +8,6 @@ export default function (maintenanceMode = {}) {
     const header = maintenanceMode.header;
     const notice = maintenanceMode.notice;
 
-    let scrollTop = 0;
-
     if (!(header && notice)) {
         return;
     }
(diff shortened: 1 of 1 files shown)


icatalina:
LGTM 👍

PR #856, opened 2016-10-01
Sale badges

briansilvey:
Merging branch of sale badges

### DIFF
--- a/assets/scss/layouts/products/_productSaleBadges.scss
+++ b/assets/scss/layouts/products/_productSaleBadges.scss
@@ -66,9 +66,9 @@ span.sale-text-burst {
     font-weight: 700;
     height: rem-calc(20px);
     line-height: rem-calc(16px);
-    padding-top: spacing("eighth") / 2;
     padding-left: spacing("third");
     padding-right: spacing("half");
+    padding-top: spacing("eighth") / 2;
     position: absolute;
     transition: 800ms ease;
     z-index: zIndex("low");
--- a/assets/scss/layouts/products/_productSaleBadges.scss
+++ b/assets/scss/layouts/products/_productSaleBadges.scss
@@ -13,9 +13,7 @@
     height: 50px;
     overflow: visible;
     position: relative;
-     -webkit-transform: scaleX(1) scaleY(1) scaleZ(1);
     transform: scaleX(1) scaleY(1) scaleZ(1);
-    -webkit-transform-origin: 50% 50% 0;
     transform-origin: 50% 50% 0;
     width: 50px;
     z-index: auto;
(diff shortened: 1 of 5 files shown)


briansilvey:
@bc-adamferenzi @johnhebron @cristycarpenter

assets/scss/layouts/products/_productSaleBadges.scss
```diff
@@ -0,0 +1,112 @@
...
+    border-radius: 0 50px 50px 0;
+    color: stencilColor("color_text_product_sale_badges");
+    float: left;
+    font-size: 15px;
+    font-weight: 700;
+    height: 20px;
+    line-height: 16px;
+    margin-bottom: 4px;
+    padding: 2px 10px;
+    position: absolute;
```
bc-miko-ademagic: Can you please use `remCalc()` to convert the values for height, width, and line height into rems? E.g. `height: remCalc(20px);`. You should be able to find examples in Cornerstone. Also, the padding and margins should ideally use our [spacing map] (https://github.com/bigcommerce/citadel/blob/fc1958d9f5731e228d02926a207f8cc393947c5b/src/settings/global/layout/_layout.scss#L51).

assets/scss/layouts/products/_productSaleBadges.scss
```diff
@@ -0,0 +1,112 @@
...
+
+span.sale-text-burst {
+    color: stencilColor("color_text_product_sale_badges");
+    font-weight: 600;
+    line-height: 1rem;
+    position: absolute;
+    text-align: center;
+    top: 20%;
+    width: 100%;
+    z-index: 999;
```
bc-miko-ademagic: We try to stay away from arbitrary values for z-index so that they don't interfere with other parts of the UI, e.g. Modals, flyouts. Can you use something from [our z-index map](https://github.com/bigcommerce/citadel/blob/fc1958d9f5731e228d02926a207f8cc393947c5b/src/settings/global/z-index/_z-index.scss) instead?

assets/scss/layouts/products/_productSaleBadges.scss
briansilvey: @bc-miko-ademagic thanks for looking this over and for all the tips! I have a question... I tried using the spacing map for padding and I think I am going to have to get a little more specific for padding on some of these. Is it ok to use ```rem-calc``` in the instances where the space mapping might not work best?

bc-miko-ademagic: To make it easier to follow our spacing rhythm we prefer people try to follow the spacing map. You can do things like `spacing("single") + spacing("quarter")` etc to get more specific values out of it. You've got tiny values - is there a reason `spacing("eighth")` won't work instead of `2px` for example?

briansilvey: I just saw that you can add, multiply etc. and got it to work. Thanks for the quick response, learning some new stuff so appreciate all the tips! :)

mjschock:
Closing as it's been reopened as #896

PR #858, opened 2016-10-01
As low as

bc-adamferenzi:
Added "As low as" text to price on layout pages if product has options, along with toggle in Theme Editor
https://jira.bigcommerce.com/browse/MERC-1165


mjschock:
New PR opened:  https://github.com/bigcommerce/stencil/pull/886

PR #865, opened 2016-11-08
PROJECT-1134: Customizing UCO in Stencil Theme Editor

jordanwink201:
## What?
- CHECKOUT-917: add header styles
- CHECKOUT-919: add main content uco customizations
- CHECKOUT-929: allow for custom header bg and custom logo
- CHECKOUT-1191: add classes for customizations
## Why?

Customizing UCO in Stencil Theme Editor
## Testing / Proof

https://store-tl7pvkadws.my-integration.zone/

@bigcommerce/checkout @bc-miko-ademagic @bc-anna @mcampa @lord2800


schema.json
```diff
@@ -1462,6 +1462,587 @@
     ]
   },
   {
+    "name": "Optimized Checkout",
```
mcampa: I wonder if we should hide this tab if you are not using UCO. Otherwise it will be confusing when users try to change a setting and nothing changes.

We could add an attribute that shows/hide depending on a store configuration. Something like this:

```json
{
    "name": "Optimized Checkout",
    "enabled": "ucoEnabled",
    "settings": [ 
        ...
    ]
}
```

jordanwink201: Yes the section should be hidden if UCO is not enabled. At the moment the condition is getting sent through `BC_SEED_DATA` https://github.com/bigcommerce-labs/ng-stencil-editor/pull/162/files#diff-607a41934429c0248a0a01f2eb030febR88

mcampa:
The test store url in the description is not working, can you fix that?

schema.json
jordanwink201: Yeah it could replace `this.formatId(name) === 'optimizedcheckout'`
Would it be much cleaner (and worth changing the theme `schema.json`), rather than just checking if `"name": "Optimized Checkout"` exists in the theme ?

assets/scss/optimized-checkout.scss
```diff
@@ -0,0 +1,180 @@
...
+// Background, Logo Position
+// -----------------------------------------------------------------------------
+
+.optimizedCheckout-header {
+    background-color: stencilColor("optimizedCheckout-header-backgroundColor");
+    // scss-lint:disable StringQuotes
+    background-image: url(stencilImage('optimizedCheckout-backgroundImage', 'optimizedCheckout-backgroundImage-size'));
+    // scss-lint:enable StringQuotes
+    background-size: cover;
+
```
jordanwink201: What background css properties should we use for this background image since we currently also have the ability to change the background image's size? Or should we remove the ability to change the size of the background image?
@bc-miko-ademagic @bc-annavu @icatalina @davidchin

bc-annavu: I would prefer to keep `background-size: cover;` since we're using the same image for all responsive sizes and this guarantees that there won't be any blank spaces at the edges of the header.

As for if we should be allowing for different image sizes - I think having one size makes more sense. If we allow them to upload smaller image sizes, we run the risk that the images will be stretched in different screen sizes. Currently in the code, we don't do anything special with the design at different breakpoints for us to allow for variable sizes.

assets/scss/optimized-checkout.scss
```diff
@@ -0,0 +1,174 @@
...
+    font-family: stencilFontFamily("optimizedCheckout-headingSecondary-font"), Arial, Helvetica, sans-serif;
+    font-weight: stencilFontWeight("optimizedCheckout-headingSecondary-font");
+}
+
+//
+// Body
+// Background & Shipping Method Overlay
+// -----------------------------------------------------------------------------
+
+body {
```
jordanwink201: would it be better to have a body class instead of targeting the element?
@davidchin @icatalina

icatalina: I don't think it'll make much of a different, tbh... Do you think it'll be easier / more consistent for documenting it?

assets/scss/optimized-checkout.scss
```diff
@@ -0,0 +1,174 @@
...
+    font-family: stencilFontFamily("optimizedCheckout-contentSecondary-font"), Arial, Helvetica, sans-serif;
+    font-weight: stencilFontWeight("optimizedCheckout-contentSecondary-font");
+}
+
+//
+// Primary Button
+// Background & Border & Text Color & Typography
+// -----------------------------------------------------------------------------
+
+.optimizedCheckout-button--primary {
```
davidchin: Should this be `optimizedCheckout-buttonPrimary`?

assets/scss/optimized-checkout.scss
```diff
@@ -0,0 +1,174 @@
...
+        color: stencilColor("optimizedCheckout-button-primary-colorActive");
+    }
+}
+
+//
+// Secondary Button
+// Background & Border & Text Color & Typography
+// -----------------------------------------------------------------------------
+
+.optimizedCheckout-button--secondary {
```
davidchin: Should this be `optimizedCheckout-buttonSecondary`?

assets/scss/optimized-checkout.scss
```diff
@@ -0,0 +1,174 @@
...
+    font-weight: stencilFontWeight("optimizedCheckout-contentSecondary-font");
+}
+
+//
+// Primary Button
+// Background & Border & Text Color & Typography
+// -----------------------------------------------------------------------------
+
+.optimizedCheckout-button--primary {
+    background-color: stencilColor("optimizedCheckout-button-primary-backgroundColor");
```
davidchin: Should this be `optimizedCheckout-buttonPrimary-backgroundColor`, similar to `optimizedCheckout-contentPrimary-color`?

assets/scss/optimized-checkout.scss
```diff
@@ -0,0 +1,174 @@
...
+// =============================================================================
+
+//
+// Header Section
+// Background, Logo Position
+// -----------------------------------------------------------------------------
+
+.optimizedCheckout-header {
+    background-color: stencilColor("optimizedCheckout-header-backgroundColor");
+    // scss-lint:disable StringQuotes
```
davidchin: Just curious, what's the reason for using single quote?

assets/scss/optimized-checkout.scss
jordanwink201: @bc-annavu agreed to keep the button classes to use `--`:`.optimizedCheckout-button--primary`, but change the stencil settings to use `-`:`optimizedCheckout-button-primary-backgroundColor`

bc-annavu: In settings, we don't typically denote modifiers with `--` which is why we are only using one `-` for it. 

As  for this particular class name - I'm indifferent. I see it as an actual modifier to the button class, so in theory we could just leave the `button--primary` to it. But if we want consistency changing it to `buttonPrimary` would be best. I think in this case, figuring what is the easier to document.

assets/scss/optimized-checkout.scss
bc-miko-ademagic: I think body is one of the few elements we style without using classes. Others I can think of are `a`, and to a small extent the `ul`s and `ol`s. For the purpose of this file, we should try to keep the selectors as close to (if not identical to) what they currently are.

bc-miko-ademagic:
Looks 👍  to me, but i'd love a sanity check from @icatalina

assets/scss/optimized-checkout.scss
bc-miko-ademagic: tl;dr this is fine imo

icatalina:
Checked, all seems ok. Just one thing, I don't know how the release process works. Is it safe to merge, or will it be better to wait for the rest of `PROJECT-1134` to have been merged?

jordanwink201:
merging all of PROJECT-1134 PR's now

PR #870, opened 2016-11-13
MERC-1025, MERC-1477: HTML refinements.

PascalZajac:
## What?
* Demonstrate usage of the `locale_name` variable for specifying the document language.
* Add `title` attributes to all images.

## Why?
For accessibility and SEO.

## Testing / Proof
screen shot 2016-11-13 at 7 24 22 pm
screen shot 2016-11-13 at 8 03 36 pm

Ping @mcampa @lord2800 @bc-AlyssNoland

### DIFF
--- a/templates/components/account/order-contents.html
+++ b/templates/components/account/order-contents.html
@@ -21,11 +21,12 @@
             </div>
             <figure class="account-product-figure">
                 {{#if type '===' 'giftcertificate'}}
-                    <img src="{{cdn ../../theme_settings.default_image_gift_certificate}}" alt="GiftCertificate">
+                    <img src="{{cdn ../../theme_settings.default_image_gift_certificate}}" alt="Gift Certificate" title="Gift Certificate">
                 {{else}}
                     <img class="account-product-image"
                          src="{{getImage image 'productthumb_size' (cdn ../../theme_settings.default_image_product)}}"
-                         alt="{{image.alt}}">
+                         alt="{{image.alt}}"
+                         title="{{image.alt}}">
                 {{/if}}
                 {{#unless refunded}}
                     {{#if download_url}}
--- a/templates/components/page/contact-us-form.html
+++ b/templates/components/page/contact-us-form.html
@@ -50,7 +50,7 @@
     {{#if forms.contact.captcha_url}}
         <div class="form-field">
             <label class="form-label" for="captcha">{{lang 'forms.contact_us.captcha'}}</label>
-            <img src="{{forms.contact.captcha_url}}" alt="img">
+            <img src="{{forms.contact.captcha_url}}" alt="Verification image" title="Verification image">
             <p>{{lang 'forms.contact_us.captcha_desc'}}</p>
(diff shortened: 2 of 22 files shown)


davidchin:
I think `alt` should be sufficient. I'd only use `title` if the text is different to `alt` and you want it to show as a tooltip. I think major search engines use `alt` text for indexing. If you use both and they are the same, you might even get penalised for keyword spamming. Screen readers also prefer `alt` over `title` for images. But MERC-1477 seems to suggest I'm wrong 😞

templates/components/account/order-contents.html
```diff
@@ -21,11 +21,12 @@ <h5 class="account-listShipping-title">{{lang 'account.orders.details.ship_to_mu
             </div>
             <figure class="account-product-figure">
                 {{#if type '===' 'giftcertificate'}}
-                    <img src="{{cdn ../../theme_settings.default_image_gift_certificate}}" alt="GiftCertificate">
+                    <img src="{{cdn ../../theme_settings.default_image_gift_certificate}}" alt="GiftCertificate" title="GiftCertificate">
```
davidchin: `alt` and `title` should be readable by humans. So "Gift certificate" instead of "GiftCertificate". `title` text should ideally be different to `alt`.

templates/components/page/contact-us-form.html
```diff
@@ -50,7 +50,7 @@
     {{#if forms.contact.captcha_url}}
         <div class="form-field">
             <label class="form-label" for="captcha">{{lang 'forms.contact_us.captcha'}}</label>
-            <img src="{{forms.contact.captcha_url}}" alt="img">
+            <img src="{{forms.contact.captcha_url}}" alt="img" title="img">
```
davidchin: Same as above, `alt` and `img` should be meaningful. So it should probably be something about "captcha" instead of "img".

templates/components/account/order-contents.html
PascalZajac: Yeah I wondered why the current value was meaningless. I'm happy to update it.

templates/components/page/contact-us-form.html
PascalZajac: What would you suggest for this? `CAPTCHA`? `Verification Image`?

PascalZajac:
> I think alt should be sufficient.
Historically I've used both just since browsers tended to prefer one or the other on mouseover, but I believe those days are past. Regardless, the discussion on Slack and the JIRA implies there's value in doing this, so...

templates/components/page/contact-us-form.html
davidchin: Yeah I'm happy with "Verification image".

templates/components/account/order-contents.html
PascalZajac: 🔧

templates/components/page/contact-us-form.html
PascalZajac: 🔧

davidchin:
Yeah, I'm not a SEO expert so I'm going to take his recommendation. :)

PR #872, opened 2016-11-15
Malformed json

mcampa:
I'm not sure how this got merged. Seems like the tests we have in place don't check for this.

@johnhebron @bc-adamferenzi Please make sure you test your code before opening a PR

@lord2800


lord2800:
👍

bc-adamferenzi:
Hrm...not sure how that got through either. Looks like it was an extra comma on my end.  It was def tested, but I'm wondering if in my late night switch of PR I ended up on the wrong one :/  My fault, won't happen again.

PR #878, opened 2016-11-19
Revert "CHECKOUT-1308: re-position optimized checkout settings "

jordanwink201:
Reverts bigcommerce/stencil#877

@aleachjr


aleachjr:
If there is already an exception to get this out then it should stay. I know there was talk this week trying to get it out but wasnt sure that it was happening. If you know it is happening then lets leave it for Monday.

PascalZajac:
Given the nature of the changes, I don't care if it stays in.

PR #879, opened 2016-11-23
MERC-1272 Replace history.js with html5 history api shim

mcampa:
# What?
MERC-1272: Use of URLs with a hashbang (#!) removes the hashbang from the URL and adds it back to window.location without the #

# How?
This PR replaces [history.js](https://github.com/browserstate/history.js/) which is abandoned and buggy with [HTML5-History-API](https://github.com/devote/HTML5-History-API) which operates according to W3C specification.

@lord2800 @sherrybc

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ node_modules/
 .DS_Store
 assets/css-artifacts
 assets/js/bundle.js
+assets/js/bundle.js.map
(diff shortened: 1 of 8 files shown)


sherrybc:
LGTM 👍

nickdengler:
💚  Tested on integration on this store: https://store-uq4oft7hn6.my-integration.zone/manage

PR #886, opened 2016-12-03
As low as

mjschock:
* replaces https://github.com/bigcommerce/stencil/pull/858

* shows 'As low as' label on the non-product pages when the `price_as_low_as` flag is set to `true`; defaults to `false` for *all* Cornerstone themes.  flag is set with toggle in Theme Editor.

toggle

`price_as_low_as = false`

off

`price_as_low_as = true`

on

### DIFF
--- a/config.json
+++ b/config.json
@@ -234,7 +234,8 @@
     "optimizedCheckout-step-text": "#ffffff",
     "optimizedCheckout-form-text": "#666666",
     "optimizedCheckout-formField-backgroundColor": "white",
-    "optimizedCheckout-formField-borderColor": "#989898"
+    "optimizedCheckout-formField-borderColor": "#989898",
+    "price_as_low_as": false
   },
   "read_only_files": [
     "/assets/scss/components/citadel",
--- a/templates/components/products/price.html
+++ b/templates/components/products/price.html
@@ -1,37 +1,83 @@
-{{#if price.with_tax}}
-    <div class="price-section price-section--withTax" {{#if schema_org}}itemprop="offers" itemscope itemtype="http://schema.org/Offer"{{/if}}>
-        {{#if price.rrp_with_tax}}
-            <span data-product-rrp-with-tax class="price price--rrp">{{price.rrp_with_tax.formatted}}</span>
-        {{/if}}
-        <span data-product-price-with-tax class="price">{{price.with_tax.formatted}}</span>
-        {{#if price.without_tax}}
-            {{lang 'products.price_with_tax' tax_label=price.tax_label}}
-        {{/if}}
-    </div>
-{{/if}}
-
-{{#if price.without_tax}}
-    <div class="price-section price-section--withoutTax {{#if price.with_tax}}price-section--minor{{/if}}"  {{#if schema_org}}itemprop="offers" itemscope itemtype="http://schema.org/Offer"{{/if}}>
-        {{#if price.rrp_without_tax}}
-            <span data-product-rrp-without-tax class="price price--rrp">{{price.rrp_without_tax.formatted}}</span>
-        {{/if}}
-        {{#if schema_org}}
-            <meta itemprop="price" content="{{price.without_tax.value}}">
-            <meta itemprop="priceCurrency" content="{{currency_selector.active_currency_code}}">
-        {{/if}}
-        <span data-product-price-without-tax class="price price--withoutTax">{{price.without_tax.formatted}}</span>
-        {{#if price.with_tax}}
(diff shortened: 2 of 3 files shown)


templates/components/products/price.html
```diff
@@ -1,15 +1,20 @@
...
+        {{else}}
         <span data-product-price-with-tax class="price">{{price.with_tax.formatted}}</span>
+        {{/if}}
         {{#if price.without_tax}}
             {{lang 'products.price_with_tax' tax_label=price.tax_label}}
         {{/if}}
     </div>
 {{/if}}
-
 {{#if price.without_tax}}
```
mjschock: wondering if we should use an `{{#else}}` here - can (or should) both `price.with_tax` and `price.without_tax` every be falsey simultaneously?

mjschock: i changed it to an else but perhaps there's a special case i'm missing here?

templates/components/products/price.html
```diff
@@ -1,16 +1,20 @@
+<!--products have options-->
+{{#if product.options.length}}
```
mjschock: `has_options` is always undefined here because it's not available on the product model (https://stencil.bigcommerce.com/docs/product-resources).  instead it's on the product card model (https://stencil.bigcommerce.com/docs/common-product-card-model)

templates/components/products/price.html
```diff
@@ -1,16 +1,20 @@
+<!--products have options-->
```
mjschock: should we include the comments?

templates/components/products/price.html
mjschock: actually, i'm confused now - what is the meaning of the following for example:

```
{{#if price.with_tax}}
    <div class="price-section price-section--withTax" {{#if schema_org}}itemprop="offers" itemscope itemtype="http://schema.org/Offer"{{/if}}>
        ...
        {{#if price.without_tax}}
            {{lang 'products.price_with_tax' tax_label=price.tax_label}}
        {{/if}}
    </div>
{{else}}
```

templates/components/products/price.html
mcampa: fine by me

templates/components/products/price.html
```diff
@@ -1,16 +1,20 @@
...
+{{#if product.options.length}}
 {{#if price.with_tax}}
     <div class="price-section price-section--withTax" {{#if schema_org}}itemprop="offers" itemscope itemtype="http://schema.org/Offer"{{/if}}>
         {{#if price.rrp_with_tax}}
             <span data-product-rrp-with-tax class="price price--rrp">{{price.rrp_with_tax.formatted}}</span>
         {{/if}}
+        {{#if theme_settings.price_as_low_as}}
+        <span data-product-price-with-tax class="price">As low as {{price.with_tax.formatted}}</span>
```
mcampa: indent properly please

templates/components/products/price.html
mjschock: nevermind - not thinking of with_tax/without-tax in the right way

templates/components/products/price.html
```diff
@@ -35,3 +42,42 @@
         </div>
     {{/if}}
 {{/if}}
+<!--END products with options-->
+{{else}}
```
mcampa: Looks like there is some indentation problems above

templates/components/products/price.html
```diff
@@ -1,37 +1,83 @@
...
-        {{#if price.with_tax}}
-            <abbr title="{{lang 'products.excluding_tax'}}">{{lang 'products.price_without_tax' tax_label=price.tax_label}}</abbr>
+{{#or has_options product.options.length}}
+    {{#if price.with_tax}}
+        <div class="price-section price-section--withTax" {{#if schema_org}}itemprop="offers" itemscope itemtype="http://schema.org/Offer"{{/if}}>
+            {{#if price.rrp_with_tax}}
+                <span data-product-rrp-with-tax class="price price--rrp">{{price.rrp_with_tax.formatted}}</span>
+            {{/if}}
+            {{#if theme_settings.price_as_low_as}}
+                <span translate>As low as </span>
```
mjschock: pulled this out from within the span (and below too) because updatePriceView in product-details.js will overwrite everything inside the span

config.json
```diff
@@ -277,7 +278,9 @@
         ]
       },
       "images": {},
-      "settings": {}
+      "settings": {
+        "button--primary-color": "#bf4141"
```
mjschock: will delete - didn't mean to add

bc-ranji:
💚

mjschock:
@mcampa or @nboliver - is it ok to merge this now?

mcampa:
👍

PR #888, opened 2016-12-07
MERC-1560 Updating rating to links instead of select list options.

junedkazi:
#  Do not merge

Currently we are showing ratings facets as select list but it works more like radio buttons. So this will help move it away from select list options & use it as single select links.

### Before
ratings-before

### After
ratings-after

ping @mcampa @bc-ranji @bc-julianleong


mcampa:
test are failing. you need to use a class for styling instead

junedkazi:
@mcampa test have been fixed. Can you give this one a quick pass ?

mcampa:
👍 LGTM

shanth100:
💚

PR #894, opened 2016-12-12
add yarn.lock

mjschock:
see https://yarnpkg.com/en/docs/yarn-lock#toc-check-into-source-control


mjschock:
figured i'll probably inadvertently include it in a PR at some point so i might as well do it know and get rid of the overhead in making sure not to commit it.

mcampa:
I agree on adding `yarn.lock` to the repo, although is not documented, users would want to use `yarn` instead of `npm`.

@lord2800 what do you think?

lord2800:
My only concern is https://github.com/yarnpkg/yarn/issues/761, but that may not be a problem for this. Should we add a note that while we have a yarn lockfile, we don't regularly test with it and it may be incorrect/out of date at any time?

mjschock:
@lord2800 - actually, i've just noticed that sometimes yarn doesn't seem to install phantomjs-prebuilt correctly.  it's a dependency of karma-phantomjs-launcher.  as a result, the `karma:unit` task hangs for me when using yarn.  :(  but it does speed up the npm install locally :)  i think a note would be the best approach.  should we add that to the readme?

mcampa:
I'm going to vote to not include `yarn.lock` because of these issues

mjschock:
@mcampa ok, i'll just create an alias for `yarn install --no-lockfile` and reinvestigate later

PR #900, opened 2017-01-03
MERC-1619

cristycarpenter:
MERC-1619: Adding a margin top to the carousel on mobile view. This is to prevent the carousel's images from being cutoff near the top on mobile devices. Here's a screenshot of before (left) and after (right):

before-after-merc-1619

**Why**
Hi Cristy Carpenter and Pascal Zajac, I cannot remember anything . But looking at the code, I think a positive margin of 2.5rem is added to `.body`. in order to visually override the negative 2.5rem margin added to `.heroCarousel`. We need to do to this because we want a gap between the header and the body for content pages. But for the homepage, we want the carousel to sit flush against the header.
It looks like when Chris adjusted the body margin to 1.5rem for mobile, he forgot to also reduce it in `.heroCarousel`. So, instead of removing breakpoint adjustment in `_body.scss`, I think the proper fix is to change `_heroCarousel.scss` to have `margin-top` like the following:

```
.heroCarousel {
    margin-top: -spacing("single");
 
        margin-top: -(spacing("single") + spacing("base"));
    }
}
```

### DIFF
--- a/assets/scss/components/stencil/heroCarousel/_heroCarousel.scss
+++ b/assets/scss/components/stencil/heroCarousel/_heroCarousel.scss
@@ -15,7 +15,11 @@
 
 .heroCarousel {
     margin-bottom: (spacing("double") + spacing("single"));
-    margin-top: -(spacing("single") + spacing("base"));
+    margin-top: -(spacing("single"));
+
+    @include breakpoint("medium") {
+        margin-top: -(spacing("single") + spacing("base"));
+    }
 
     .js & { // 2
         max-height: remCalc(0);
--- a/assets/scss/components/stencil/heroCarousel/_heroCarousel.scss
+++ b/assets/scss/components/stencil/heroCarousel/_heroCarousel.scss
@@ -18,6 +18,9 @@
     margin-top: -(spacing("single"));
 
     @include breakpoint("medium") {
+        // Added to visually override the top margin for _body.scss on mobile.
+        // The '.body' top margin creates space between the header and page content.
+        // However, on the homepage, we want the carousel flush with the header.
         margin-top: -(spacing("single") + spacing("base"));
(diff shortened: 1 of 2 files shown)


assets/scss/components/stencil/heroCarousel/_heroCarousel.scss
```diff
@@ -15,7 +15,11 @@
 
 .heroCarousel {
     margin-bottom: (spacing("double") + spacing("single"));
-    margin-top: -(spacing("single") + spacing("base"));
+    margin-top: -(spacing("single"));
+
+    @include breakpoint("medium") {
```
PascalZajac: I think it would be worth adding a comment explaining _why_ this is necessary (basically copy the explanation David gave on the JIRA). Likewise it's probably worthwhile to add a comment to the `.body` definition to say "If you change the spacing here, please update the heroCarousel too".

davidchin:
Hey @cristycarpenter, nice work! 👏  I have a small comment to make. In this project, we follow a specific format when we leave comments in CSS files. Can you please have a look at this [section](https://github.com/bigcommerce/sass-style-guide#commenting) of the style guide? If you scroll to the top of `_heroCarousel.scss` file, you can see an [example](https://github.com/bigcommerce/stencil/blob/master/assets/scss/components/stencil/heroCarousel/_heroCarousel.scss#L11) which you can follow. Thanks!

cristycarpenter:
Oh, that makes sense now @davidchin. I like this process for comments  - it's very clean. Thanks for the guide! I can update it in the morning :)

PascalZajac:
Huh, TIL. That's cool, we should do that more widely.

PascalZajac:
Tested locally, looks fine to me.

PR #901, opened 2017-01-05
MERC-1637 Switch contact captcha to use google captcha.

junedkazi:
Switch contact captcha to use google captcha instead of internal version of captcha.

ping @mcampa @lord2800 @PascalZajac

### DIFF
--- a/templates/components/page/contact-us-form.html
+++ b/templates/components/page/contact-us-form.html
@@ -47,15 +47,24 @@
         {{/if}}
     </div>
 
-    {{#if forms.contact.captcha_url}}
-        <div class="form-field">
-            <label class="form-label" for="captcha">{{lang 'forms.contact_us.captcha'}}</label>
-            <img src="{{forms.contact.captcha_url}}" alt="Verification image" title="Verification image">
-            <p>{{lang 'forms.contact_us.captcha_desc'}}</p>
-            <input class="form-input" type="text" id="captcha" name="captcha">
+    {{#if forms.contact.recaptcha.enabled}}
+        <div class="panel">
+            <header class="panel-header">
+                <h3 class="panel-title">{{lang 'forms.contact_us.captcha'}}</h3>
+            </header>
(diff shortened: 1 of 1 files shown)


templates/components/page/contact-us-form.html
```diff
@@ -47,15 +47,24 @@
...
+                <h3 class="panel-title">{{lang 'forms.contact_us.captcha'}}</h3>
+            </header>
+            <div class="panel-body">
+                <script type="text/javascript"
+                        src="https://www.google.com/recaptcha/api/challenge?k={{forms.contact.recaptcha.public_key}}">
+                </script>
+                <noscript>
+                    <iframe src="https://www.google.com/recaptcha/api/noscript?k={{forms.contact.recaptcha.public_key}}" height="300" width="500" frameborder="0"></iframe>
+                    <br>
+                    <textarea name="recaptcha_challenge_field" rows="3" cols="40">
```
mcampa: Why is this a textarea instead of an input? and why 3x40?

junedkazi: This is google markup https://developers.google.com/recaptcha/old/docs/customization & we use the same markup for create account page template as well https://github.com/bigcommerce/stencil/blob/master/templates/pages/auth/create-account.html#L35

lord2800: Eventually I'd love to get this out of the template and into an object instead. This kind of markup belongs on our end, so it can't be accidentally broken.

junedkazi: I agree & I do have plans to do it part of captcha v2 upgrade.

mcampa:
Can you post a screenshot on how this looks? with js enabled and disabled

mcampa:
👍

junedkazi:
#### With JS
stencil-after-with-js
#### With NO JS
stencil-after-with-nojs

shanth100:
💚

PR #903, opened 2017-01-11
STENCIL-2598 - Add GeoTrust SSL Seal Toggle

mjschock:
## what (STENCIL-2598, formerly MERC-1160)
* adds a toggle for the GeoTrust SSL Seal
* implements the the instructions <a href="https://www.geotrust.com/support/seal/agreement/installation-instructions/">here</a>
* the domain and size can be set

@bigcommerce/stencil-team

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Prevent carousel images from being cut off on large screens by adding a new setting to theme editor schema [#909](https://github.com/bigcommerce/stencil/pull/909)
 - Add schema description specifying that social media icons must be set up to see them [#920](https://github.com/bigcommerce/stencil/pull/920)
 - Show account creation links only if it is enabled in store settings [#917] (https://github.com/bigcommerce/stencil/pull/917)
+- Add GeoTrust SSL Seal Toggle [#903] (https://github.com/bigcommerce/stencil/pull/903)
 
 ## 1.5.1 (2017-02-07)
 - Fix an issue with a horizontal scroll bar showing in theme editor [#915](https://github.com/bigcommerce/stencil/pull/915)
--- a/config.json
+++ b/config.json
@@ -245,7 +245,10 @@
     "restrict_to_login": false,
     "swatch_option_size": "22x22",
     "social_icon_placement_top": false,
-    "social_icon_placement_bottom": "bottom_none"
+    "social_icon_placement_bottom": "bottom_none",
+    "show_geotrust_ssl_seal": false,
+    "geotrust_ssl_seal_hostname": "",
+    "geotrust_ssl_seal_size": "M"
   },
   "read_only_files": [
     "/assets/scss/components/citadel",
--- a/config.json
+++ b/config.json
@@ -247,7 +247,6 @@
     "social_icon_placement_top": false,
     "social_icon_placement_bottom": "bottom_none",
     "show_geotrust_ssl_seal": false,
-    "geotrust_ssl_seal_hostname": "",
     "geotrust_ssl_seal_size": "M"
   },
   "read_only_files": [
--- a/templates/components/common/ssl-seal.html
+++ b/templates/components/common/ssl-seal.html
@@ -1,0 +1,8 @@
+<table width="135" border="0" cellpadding="2" cellspacing="0" title="Click to Verify - This site chose GeoTrust SSL for secure e-commerce and confidential communications.">
+    <tr>
+        <td width="135" align="center" valign="top">
+            <script type="text/javascript" src="https://seal.geotrust.com/getgeotrustsslseal?host_name={{theme_settings.geotrust_ssl_seal_hostname}}&amp;size={{theme_settings.geotrust_ssl_seal_size}}&amp;lang={{locale_name}}"></script><br />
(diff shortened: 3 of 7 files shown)


templates/components/common/ssl-seal.html
```diff
@@ -0,0 +1,8 @@
+<table width="135" border="0" cellpadding="2" cellspacing="0" title="Click to Verify - This site chose GeoTrust SSL for secure e-commerce and confidential communications.">
+    <tr>
+        <td width="135" align="center" valign="top">
+            <script type="text/javascript" src="https://seal.geotrust.com/getgeotrustsslseal?host_name={{theme_settings.geotrust_ssl_seal_hostname}}&amp;size={{theme_settings.geotrust_ssl_seal_size}}&amp;lang={{locale_name}}"></script><br />
+            <a href="http://www.geotrust.com/ssl/" target="_blank"  style="color:#000000; text-decoration:none; font:bold 7px verdana,sans-serif; letter-spacing:.5px; text-align:center; margin:0px; padding:0px;"></a>
+        </td>
+    </tr>
+</table>
```
willfaught: Missing NL

config.json
```diff
@@ -590,4 +593,4 @@
       }
     }
   ]
-}
+}
```
willfaught: Missing NL

willfaught:
🍹 Keep in mind https://github.com/bigcommerce/standards/blob/master/docs/Git.md. Can we add the ticket (MERC-1160?) to the summary and use a verb (add?) and sentence case (Add GeoTrust SSL...)?

templates/components/common/ssl-seal.html
mjschock: hmmm...  yes, i see. the linter doesn't look at html or json.  wondering if we should add something to do the check.

mjschock: i've added a new line at the end but it doesn't seem to show up here, although you can see that there are now 9 lines:  https://github.com/mjschock/stencil/blob/f80cdfc06862a9668db449697261c8d9a07eb4a2/templates/components/common/ssl-seal.html

config.json
mjschock: it's been added.  see https://github.com/mjschock/stencil/blob/f80cdfc06862a9668db449697261c8d9a07eb4a2/config.json.  
screen shot 2017-01-25 at 10 19 04 am

.eslintrc
```diff
@@ -7,7 +7,7 @@
         "complexity": 1,
         "no-alert": 0,
         "consistent-return": 0,
-        "max-len": 0,
+        "max-len": 0
```
mcampa: Unnecessary change. `.eslintrc` is not real json format 🍹 . 
You can have trailing commas and add comments in this file

mjschock: yeah, just removed it because vscode doesn't like it.   it's been added back :)

config.json
```diff
@@ -244,7 +244,10 @@
     "restrict_to_login": false,
     "swatch_option_size": "22x22",
     "social_icon_placement_top": false,
-    "social_icon_placement_bottom": "bottom_none"
+    "social_icon_placement_bottom": "bottom_none",
+    "show_geotrust_ssl_seal": false,
+    "geotrust_ssl_seal_hostname": "",
```
junedkazi: why not use `shopPath` instead ? Is there a reason for not using the store domain instead of offering a separate field ?

mcampa: Good question, however `shopPath` or `ShopPathSSL` return a url `http://...` not the hostname. But we could return the hostname in the `Setting` resource and use that

junedkazi: We do have `base_url` & `secure_base_url` for both http & https links. We also have the `home` in the `urls` section of the `context`. All the urls are with `http{s}://` so we will need to strip them out. The question is should we pass it as resource on the bcapp side or just add a helper in pencil to strip `http{s}://`.

mcampa: +1 It would be useful to have helper for that. Like `getHostname`

But the question is if merchants really need to enter their own domain

junedkazi: Since the trust ssl seal is for the store domain name. I don't see a reason why we should expose a field for the user to enter it unless there is some historic reason which I am not aware of.

mcampa:

```
## Draft
- <here>

## 1.5.1 (2017-02-07)
- ...

```

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - Prevent carousel images from being cut off on large screens by adding a new setting to theme editor schema [#909](https://github.com/bigcommerce/stencil/pull/909)
 - Add schema description specifying that social media icons must be set up to see them [#920](https://github.com/bigcommerce/stencil/pull/920)
 - Show account creation links only if it is enabled in store settings [#917] (https://github.com/bigcommerce/stencil/pull/917)
+- Add GeoTrust SSL Seal Toggle [#903] (https://github.com/bigcommerce/stencil/pull/903)
```
mjschock: @junedkazi - updated

junedkazi:
@nickdengler can we test this one before we merge this one.

nickdengler:
LGTM 💚

PR #909, opened 2017-01-24
STENCIL-2455: Prevents carousel images from being cut off on large sc…

cristycarpenter:
…reens.

**JIRA Ticket** 
https://jira.bigcommerce.com/browse/STENCIL-2455

**What**
Switches carousel background-size from "cover" to "100% 100%" which fits the entire image into the slide's viewing area. 

**Why**
Carousel images are currently becoming cut off on the top and bottom at browser widths greater than 1261px.

**Caveats**
Images may look a bit stretched. 
- We could increase the height set by the hero carousel? 
- We could remove "visibility: hidden" on .heroCarousel-image (although, the image wouldn't scale :/)

**Screenshots**
Before Change:
stencil-2455-broken

After Change:
stencil-2455-fix

### DIFF
--- a/assets/scss/components/stencil/heroCarousel/_heroCarousel.scss
+++ b/assets/scss/components/stencil/heroCarousel/_heroCarousel.scss
@@ -78,17 +78,19 @@
         }
     }
 
+    &.heroCarousel-slide-stretch .heroCarousel-slide {
+        @include breakpoint("large") { // 4
+            background-size: 100% 100%;
+        }
+    }
 }
+
 
 .heroCarousel-slide {
     background-position: 50%;
     background-repeat: no-repeat;
     background-size: cover;
     position: relative;
-
-    @include breakpoint("large") { // 4
-        background-size: 100% 100%;
-    }
 
     a {
         text-decoration: none;
--- a/assets/scss/components/stencil/heroCarousel/_heroCarousel.scss
+++ b/assets/scss/components/stencil/heroCarousel/_heroCarousel.scss
@@ -15,6 +15,9 @@
 // 3. Visually overrides the top margin for '.body' defined in _body.scss.
 //    The '.body' top margin creates space between the header and page content.
 //    However, on the homepage, we want the carousel to be flush with the header.
+//
+// 4. Allows image to scale on large screens while preventing the top and bottom
+//    from becoming cut off.
 // -----------------------------------------------------------------------------
 
 .heroCarousel {
@@ -82,6 +85,10 @@
     background-repeat: no-repeat;
     background-size: cover;
     position: relative;
+
+    @include breakpoint("large") { // 4
+        background-size: 100% 100%;
+    }
 
     a {
         text-decoration: none;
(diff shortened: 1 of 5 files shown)


mcampa:
We could add a toggle in the Theme Editor to enable this feature 🍹. Something like:
- [ ] Allows image to scale on large screens

This way people won't complain about their images being too stretched. To add the checkbox add a new setting to the `schema.json` file under `Carousel` section

cristycarpenter:
I actually had that same thought earlier @mcampa! I forgot to post it as a possible solution. I am comfortable making those code changes if that's what we think is best :)

mcampa:
Added two checkboxes:

image

image

@mjschock @junedkazi @mattolson

schema.json
```diff
@@ -745,6 +745,18 @@
     "name": "Carousel",
     "settings": [
       {
+        "type": "checkbox",
```
junedkazi: How is this being used ? I do not see any corresponding template change to support it.

templates/components/carousel.html
```diff
@@ -9,7 +9,8 @@
     }'
 >
     {{#each carousel.slides}}
-    <div class="heroCarousel-slide" style="background-image: url({{image}})">
+    <div
+        class="heroCarousel-slide" style="background-image: url({{image}})">
```
junedkazi: Is there a reason to move this to the next line ?

schema.json
mcampa: here https://github.com/bigcommerce/stencil/pull/909/files#diff-06b2d3b23dce96e1619d2b53d6c947ecR43

templates/components/carousel.html
mcampa: no, that was a mistake. Fixing

assets/scss/components/stencil/heroCarousel/_heroCarousel.scss
```diff
@@ -75,8 +78,14 @@
         }
     }
 
+    &.heroCarousel-slide-stretch .heroCarousel-slide {
```
davidchin: If you want to modify the appearance of a CSS selector, you should use `--` instead of `-`. Also I think this is unneccessarily nested. You can lower its specificity by moving it out. i.e.:

```css
.heroCarousel-slide--stretched {
    @include breakpoint("large") {
        background-size: 100%;
    }
}
```

And apply the modifier class directly to `.heroCarousel-slide` :)

mcampa: done

assets/scss/components/stencil/heroCarousel/_heroCarousel.scss
```diff
@@ -83,6 +85,12 @@
     background-size: cover;
     position: relative;
 
+    &.heroCarousel-slide--stretch {
```
davidchin: Thanks for doing the change! `heroCarousel-slide--stretch` doesn't need to be nested in `heroCarousel-slide`. Just move the former below the latter.

mcampa:

```
## Draft
- <here>

## 1.5.1 (2017-02-07)
- ...

```

nickdengler:
Tested on an integration store 💚

PR #913, opened 2017-01-25
STENCIL-2950 - Hide cart actions on compare page when restrict_to_log…

mjschock:
…in is on and user is not logged in


nickdengler:
Tested locally 💚

PR #914, opened 2017-01-31
STENCIL-2461 Check if gift certificates are enabled before rendering.

junedkazi:
Only show the gift certificate section if gift certificates are enabled.

@mcampa @mjschock @bigcommerce/stencil-team

### DIFF
--- a/templates/components/cart/totals.html
+++ b/templates/components/cart/totals.html
@@ -71,25 +71,27 @@
             {{> components/cart/coupon-input}}
         </li>
     {{/if}}
-    {{#each cart.gift_certificates}}
+    {{#if settings.gift_certificates_enabled}}
+        {{#each cart.gift_certificates}}
+            <li class="cart-total">
+                <div class="cart-total-label">
+                    <strong>{{lang 'cart.gift_certificates.code_label' code=code}}</strong>
+                    <span>
+                        {{remaining.formatted}} | <a href="{{remove_url}}">{{lang 'cart.gift_certificates.remove'}}</a>
+                    </span>
+                </div>
+                <div class="cart-total-value">
+                    {{used.formatted}}
+                </div>
+            </li>
+        {{/each}}
         <li class="cart-total">
             <div class="cart-total-label">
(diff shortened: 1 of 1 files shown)


lord2800:
👍

templates/components/cart/totals.html
```diff
@@ -71,25 +71,27 @@
...
-        </li>
-    {{/each}}
-    <li class="cart-total">
-        <div class="cart-total-label">
-            <strong>{{lang 'cart.gift_certificates.gift_certificate'}}:</strong>
-        </div>
-        {{> components/cart/gift-certificate-input}}
-    </li>
+    {{#if settings.gift_certificates_enabled}}
+      {{#each cart.gift_certificates}}
```
mcampa: 4 spaces indention

templates/components/cart/totals.html
```diff
@@ -71,25 +71,27 @@
...
-            </div>
-        </li>
-    {{/each}}
-    <li class="cart-total">
-        <div class="cart-total-label">
-            <strong>{{lang 'cart.gift_certificates.gift_certificate'}}:</strong>
-        </div>
-        {{> components/cart/gift-certificate-input}}
-    </li>
+    {{#if settings.gift_certificates_enabled}}
```
mcampa: You don't need this flag because when gift certificates are disabled `cart.gift_certificates` should be empty.

if `cart.gift_certificates` is not empty is because a bcapp bug.

mcampa: Plus, If you fix the issue in  bcapp, you fix it for all themes

junedkazi: We still need the check to hide the form element.

junedkazi: `cart.gift_certificates` empty does not mean gift certificates are disabled. It just means no gift certificates are applied to cart.

mcampa: No, but if GC are disabled `cart.gift_certificates` should be empty

nickdengler:
Checked out and tested locally 💚

PR #918, opened 2017-02-06
STENCIL-2491 Fix mobile search dropdown style

mcampa:
Don't need to show the close button because there is one in the top left corner already

https://jira.bigcommerce.com/browse/STENCIL-2491

## Before:
image

## After:
image

@bc-AlyssNoland @mjschock @bc-ong


bc-ong:
:+1:

EdFromBJCS:
Would be nice if we could clear the search results, but perhaps users don't do that anyway.

PR #920, opened 2017-02-07
STENCIL-2946 - Add message specifying that social media icons must be…

mjschock:
… set up

# WHY
* a user may not have set up there social icons yet, so they might wonder why they aren't showing up (see https://jira.bigcommerce.com/browse/STENCIL-2946)


mcampa:

```
## Draft
- <here>

## 1.5.1 (2017-02-07)
- ...

```

PR #925, opened 2017-02-17
STENCIL-2439 File upload button does not properly display in IE

mcampa:
File upload button does not properly display in IE if font-size is 0

## Before:
image

## After:
image

@bigcommerce/stencil-team


mjschock:
LGTM 👍  - after fixing the tests, that is - looks like it just needs to be moved up a level

PR #926, opened 2017-02-18
STENCIL-3084 - fix z-index on product sale badges

bookernath:
Sale badges can appear on top of the megamenu if it comes down far
enough. This is due to a bad z-index.

Proof:
Star Sale Badge
image
Diagonal Sale Badge
image
Top Left Sale Badge
image


bc-miko-ademagic:
👍  as long as it continues to appear above the product cards

PR #927, opened 2017-02-18
STENCIL-2629 - Remove option for "disable" on number of products per page

bookernath:
We have this weird option for "disabling" the number of products per page, which makes no sense and is causing confusion.

In the future we may evaluate some kind of best practice for infinite scrolling, but for now we should disable this.

Proof:
image


junedkazi:
Also we should evaluate the delta between 2 numbers at-least for cornerstone. Right now between 6-8 it is 2, then between 8-9 it is 1 & the next one is 12 where it is 3.

mcampa:
hmm, what if a merchant does not want to display one of these sections?
I don't see why this is needed, I actually think is necessary.

Why does not make sense? and why is confusing?
Maybe we don't hide the title, and that's why is confusing?

mcampa:
Disregard my message, I thought we were talking about "feature products" and "new products" in the home page

nickdengler:
The only weird thing is that if a user has this value already set, it will become set as a blank value when they upgrade. This doesn't appear to cause any problems (the storefront shows 20 items when blank, just as it did when it was "disabled"), but calling it out as weird.

💚

PR #928, opened 2017-02-18
STENCIL-3009 - automatically open review link if hash is included

bookernath:
Our product review emails link to the PDP with the hash #write_review
included.

We need to respond to this hash and automatically pop the review form.

*Testing/Proof*
Visit any product page with #write_review at the end of your URL.

### DIFF
--- a/assets/js/theme/product.js
+++ b/assets/js/theme/product.js
@@ -63,7 +63,7 @@ export default class Product extends PageManager {
     }
 
     productReviewHandler() {
-        if (this.url.indexOf('#writeReview') !== -1) {
+        if (this.url.indexOf('#write_review') !== -1) {
             this.$reviewLink.click();
         }
     }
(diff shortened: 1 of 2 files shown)


assets/js/theme/product.js
```diff
@@ -53,6 +53,10 @@ export default class Product extends PageManager {
             return false;
         });
 
+        if (window.location.hash.indexOf('write_review') === 1) {
```
willfaught: Just curious: why indexOf instead of ===?

mcampa:
I think we already have this logic here https://github.com/bigcommerce/stencil/blob/80d48965453f513751f949b498e9c3e6579612c9/assets/js/theme/product.js#L65

We need to fix it there

bookernath:
Nice catch! I'll update there.

PR #929, opened 2017-02-18
STENCIL-2493 - Do not display "none" option for required pick lists

bookernath:
Correctly hiding the “none” option for required pick lists (without
photo)

I've also verified that this works with and without a default selection on the option.

Proof:

image

### DIFF
--- a/templates/components/products/options/product-list.html
+++ b/templates/components/products/options/product-list.html
@@ -43,14 +43,15 @@
             {{/each}}
         </ul>
     {{else}}
+        {{#unless required}}
         <input class="form-radio"
                type="radio"
                name="attribute[{{id}}]"
                value="0"
                id="attribute_0_{{id}}"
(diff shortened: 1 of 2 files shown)


templates/components/products/options/product-list.html
```diff
@@ -51,6 +52,7 @@
                {{#if defaultValue '==' 0}}checked{{/if}}
                {{#if required}}required{{/if}}>
```
bc-miko-ademagic: is this if statement redundant then?

bookernath: Suppose so! I'll update.

bookernath:
@bc-miko-ademagic updated.

mcampa:
I think we should test this before merging

nickdengler:
Tested on integration store 💚

PR #934, opened 2017-02-19
STENCIL-2457 - fix incorrect suggestion warning

bookernath:
Don’t suggest that nothing was found when things are found.

Proof:

image

### DIFF
--- a/templates/pages/search.html
+++ b/templates/pages/search.html
@@ -73,13 +73,15 @@ product_results:
                             <a href="{{forms.search.suggested_query.url}}" class="search-refine">{{lang 'forms.search.refine'}}</a>
                         </div>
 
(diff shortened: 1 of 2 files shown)


templates/pages/search.html
```diff
@@ -73,13 +73,15 @@
...
                         </div>
 
-                        <div class="search-suggestion">
-                            <p>
-                                {{lang 'forms.search.your_search_for'}}
-                                "<strong>{{forms.search.query}}</strong>"
-                                {{lang 'forms.search.no_match'}}
-                            </p>
-                        </div>
+                        {{#if result_count '===' 0}}
```
junedkazi: can we just use `unless` here ?

PR #938, opened 2017-02-21
STENCIL-2445: Remove unused variable causing js error

bc-ong:
Unused event variable was causing js errors in Firefox and preventing execution of the rest of the function.

@mcampa @mjschock


mcampa:
LGTM 👍 

Can you add a line to the changelog?

PR #939, opened 2017-02-21
STENCIL-2558 - Update menu depth in navigation

mjschock:
## WHAT
* Provides an option in theme editor to select the menu design type for categories.
* Has the legacy behavior when the design is set to Simple.
* Is responsive to small and medium+ screen sizes.
* Simple setting (3 and below; default, existing behavior):  http://g.recordit.co/K67orI3sgY.gif
* Alternate setting (up to the maximum display depth value):  http://g.recordit.co/nH5xoqTtr3.gif (note that the animation for the arrow looks a little strange in the GIF)

## WHY
* https://jira.bigcommerce.com/browse/STENCIL-2558

ping @bigcommerce/stencil-team, @bc-ong

### DIFF
--- a/assets/js/theme/global/mobile-menu-toggle.js
+++ b/assets/js/theme/global/mobile-menu-toggle.js
@@ -29,8 +29,10 @@ export class MobileMenuToggle {
     } = {}) {
         this.$body = $('body');
         this.$menu = $(menuSelector);
+        this.$navList = $('.navPages-list.navPages-list-depth-max');
         this.$header = $(headerSelector);
         this.$scrollView = $(scrollViewSelector, this.$menu);
+        this.$subMenus = this.$navList.find('.navPages-action');
         this.$toggle = $toggle;
         this.mediumMediaQueryList = mediaQueryListFactory('medium');
 
@@ -38,6 +40,7 @@ export class MobileMenuToggle {
         this.onToggleClick = this.onToggleClick.bind(this);
         this.onCartPreviewOpen = this.onCartPreviewOpen.bind(this);
         this.onMediumMediaQueryMatch = this.onMediumMediaQueryMatch.bind(this);
+        this.onSubMenuClick = this.onSubMenuClick.bind(this);
 
         // Listen
         this.bindEvents();
@@ -56,6 +59,7 @@ export class MobileMenuToggle {
     bindEvents() {
         this.$toggle.on('click', this.onToggleClick);
         this.$header.on(CartPreviewEvents.open, this.onCartPreviewOpen);
+        this.$navList.on('click .navPages-action', this.onSubMenuClick);
 
         if (this.mediumMediaQueryList && this.mediumMediaQueryList.addListener) {
             this.mediumMediaQueryList.addListener(this.onMediumMediaQueryMatch);
@@ -92,6 +96,8 @@ export class MobileMenuToggle {
 
         this.$header.addClass('is-open');
         this.$scrollView.scrollTop(0);
+
+        this.resetSubMenus();
     }
 
     hide() {
@@ -106,6 +112,8 @@ export class MobileMenuToggle {
             .attr('aria-hidden', true);
 
         this.$header.removeClass('is-open');
+
+        this.resetSubMenus();
     }
 
     // Private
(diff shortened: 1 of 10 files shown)


bc-ong:
LGTM 👍

config.json
```diff
@@ -596,4 +597,4 @@
       }
     }
   ]
-}
+}
```
mcampa: nl

schema.json
```diff
@@ -696,6 +696,17 @@
     ]
   },
   {
+    "name": "Categories",
+    "settings": [
+      {
+        "type": "checkbox",
```
mcampa: Can you change this to a dropdown or radios? The idea is to have multiple designs for the navigation, if in the future we add a new style we can add a line to the dropdown.

So, instead of a boolean, we should use a string setting with 2 possible values

templates/components/common/navigation-menu.html
```diff
@@ -2,15 +2,26 @@
     <div class="navPages-quickSearch">
         {{> components/common/quick-search}}
     </div>
-    <ul class="navPages-list">
+    <ul class="navPages-list{{#if theme_settings.menu_display_depth_max}} navPages-list-depth-max{{/if}}">
```
mcampa: It would be more readable if we have something like:

```
{{#if theme_settings.navigation_design '==' 'dropdown'}}
  {{> {{> components/common/navigation-dropdown}}
{{/if}}
{{#if theme_settings.navigation_design '==' 'columns'}}
  {{> {{> components/common/navigation-columns}}
{{/if}}
```

This way it would be easier to add a third design in the future

config.json
mjschock: done, this always creeps back out when making changes in theme editor, maybe we can adjust stencil cli to add a nl here...

schema.json
mjschock: done - it's a dropdown with multiple designs now

templates/components/common/navigation-menu.html
mjschock: it's been changed - navigation_design can be 'simple' or 'alternate' where 'simple' is the default, existing design.

templates/components/common/navigation-list-alternate.html
```diff
@@ -0,0 +1,8 @@
+{{#if children}}
+    <a class="navPages-action navPages-action-depth-max has-subMenu is-root" href="{{url}}" data-collapsible="navPages-{{id}}">
+        {{name}} <i class="icon navPages-action-moreIcon" aria-hidden="true"><svg><use xlink:href="#icon-chevron-down" /></svg></i>
+    </a>
+    {{> components/common/navigation-dropdown}}
+{{else}}
+    <a class="navPages-action" href="{{url}}">{{name}}</a>
+{{/if}}
```
mcampa: You can configure you editor to always add a new line

mjschock: i tried configuring my editor to remove it but it's saved by stencil-cli and it removes the nl.  i think it's the jsonlint module:  https://github.com/bigcommerce/stencil-cli/blob/master/lib/json-lint.js#L8.  it gets rid of the nl.  should we add it back when writing to file?

mcampa: yes you should :)

mjschock: haha, yes, it's been added to navigation-list-alternate.html.  was mixing up config.json.

assets/js/theme/global/mobile-menu-toggle.js
```diff
@@ -53,9 +56,14 @@ export class MobileMenuToggle {
         return this.$menu.hasClass('is-open');
     }
 
+    get subMenuIsOpen() {
```
junedkazi: `isSubMenuOpen` 🍹

junedkazi: Also we just use it in once place. Can we just add the check directly instead of a separate method ?

mjschock: sure, the code's been moved directly inside the calling method

nickdengler:
LGTM 💚

PR #940, opened 2017-02-22
STENCIL-3190 add customized_checkout feature to theme feature list

jordanwink201:
## What
as per title ^

screen shot 2017-02-22 at 2 38 13 pm

## Why
to serve checkout.html from the stencil template itself https://github.com/bigcommerce/stencil/blob/master/templates/pages/checkout.html.
Related PR https://github.com/bigcommerce/stencil/pull/880

## Testing
https://store-vbf2yxgg1v.my-integration.zone/
theme bundle -> [Cornerstone-1.5.2.zip](https://github.com/bigcommerce/stencil/files/792175/Cornerstone-1.5.2.zip)

@bigcommerce/checkout @bigcommerce/stencil-team

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 - Remove unused variable causing js error with search in the nav [#938](https://github.com/bigcommerce/stencil/pull/938)
 - Add settings to theme editor schema to customize Optimized Checkout discount banners [#924] (https://github.com/bigcommerce/stencil/pull/924)
 - Update Optimized Checkout discount banners default values per theme variation [#942] (https://github.com/bigcommerce/stencil/pull/942)
+- Add customized checkout feature to features array [#880] (https://github.com/bigcommerce/stencil/pull/880)
 
 ## 1.5.2 (2017-02-14)
 - Added a setting to theme editor schema to show/hide the homepage carousel [#909](https://github.com/bigcommerce/stencil/pull/909)
--- a/config.json
+++ b/config.json
@@ -28,6 +28,7 @@
       "product_showcase",
       "persistent_cart",
       "one_page_check_out",
+      "customized_checkout",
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,8 @@
 - Fix an issue where required product list options would display an invalid "none" choice [#929](https://github.com/bigcommerce/stencil/pull/929)
 - Remove unused variable causing js error with search in the nav [#938](https://github.com/bigcommerce/stencil/pull/938)
 - Add settings to theme editor schema to customize Optimized Checkout discount banners [#924] (https://github.com/bigcommerce/stencil/pull/924)
+- Add optimized checkout feature [#880] (https://github.com/bigcommerce/stencil/pull/880)
```
mcampa: This sounds like you are adding the entire functionality in this PR, but is actually just a flag and this could confuse some people. Please elaborate or just add `flag` at the end

davidchin:
Please hold off merging this until https://github.com/bigcommerce/bigcommerce/pull/17621 is tested and merged, thanks!

snaderiBC:
The reason that why this feature was not added in my original PR is mainly due to the thank you page. Separating thank you page was not meant to take as long as it actually took to complete. The original plan was to add the thank you page as stencil resource before 10 Feb. Anyways, now that thank you page is a separate page but not stencil resource; if we add customized_checkout feature it will result in showing JSON content as HTML in thank you because there is no thank you resource in stencil.
The thank you page as Stencil resource needs to be added before adding customized_checkout feature flag. An alternative (but not desired) solution would be to introduce another feature for thank you page.

PascalZajac:
I don't think that's true since @icatalina addressed that issue: https://github.com/bigcommerce/bigcommerce/pull/17621

PR #949, opened 2017-03-06
CHECKOUT-1189 : Adding support to Order Confirmation page in Stencil

snaderiBC:
#### What?

1. Adding Order Confirmation page to Themes so that along with Checkout page it can also be customized.
2. ~Adding the feature `customized_checkout` to activate the feature for themes.~
I've excluded above to prevent any racing conditions when merging related PRs.
After below PRs are merged, I'll create another PR with just the theme feature to prevent any unwanted issue in production.

#### Related PRs
https://github.com/bigcommerce-labs/ng-checkout/pull/394
https://github.com/bigcommerce/bigcommerce/pull/17826

screen shot 2017-03-09 at 12 13 35 pm


snaderiBC:
Ping: @lord2800 @mcampa @icatalina @jordanwink201 for review

mcampa:
Please add a line to `CHANGELOG.md` file describing your change

snaderiBC:
Also, @mcampa I've named and renamed everywhere for the final page of check out to be `Order Confirmation`.
Looks like there have been already a page called `order-complete.html`. So, I suggest to remove that and replace it with what I've added to this PR so that it is consistent everywhere.
Also, is dash a convention? `order-confirmation.html` instead of `orderconfirmation.html` ?

PR #951, opened 2017-03-07
STENCIL-2934 Add recaptcha v2 support to cornerstone.

junedkazi:
#### What?
Adding support for google recaptcha v2.

#### Tickets / Documentation

- https://jira.bigcommerce.com/browse/STENCIL-2934

### Do Not Merge
It depends on https://github.com/bigcommerce/bigcommerce/pull/17715. 

zip: 
[Cornerstone-recaptcha-v2-1.5.3.zip](https://github.com/bigcommerce/cornerstone/files/829184/Cornerstone-recaptcha-v2-1.5.3.zip)

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -236,7 +236,6 @@
     },
     "create_account": {
         "heading": "New Account",
-        "captcha_heading": "Security Question",
         "created": {
             "heading": "Your account has been created",
             "intro": "Thank you for creating your account at {store_name}. Your account details have been emailed to <strong>{email}</strong>",
@@ -469,8 +468,6 @@
             "phone": "Phone Number",
             "order": "Order Number",
             "rma": "RMA Number",
-            "captcha": "Are you human?",
-            "captcha_desc": "Please copy the characters from the image into the text field below. Doing this helps us prevent automated submissions.",
             "question": "Comments/Questions",
             "submit": "Submit Form",
             "successful": "We've received your feedback and will respond shortly if required. <a href=\"{shopPath}\">Continue</a>."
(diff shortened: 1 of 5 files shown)


lang/en.json
```diff
@@ -581,12 +578,10 @@
         "new": "New Products",
         "warranty": "Warranty Information",
         "reviews": {
-            "captcha_heading": "Security Question",
             "hide": "Hide Reviews",
             "new": "Write a Review",
             "show": "Show Reviews",
             "header": "{total, plural, =0{0 Reviews} one {# Review} other {# Reviews}}",
-            "rating": "Rating",
```
junedkazi: Duplicate key.

templates/components/page/contact-us-form.html
```diff
@@ -47,33 +47,17 @@
...
-    {{/if}}
-
     <div class="form-field">
         <label class="form-label" for="contact_question">{{lang 'forms.contact_us.question'}}
             <small>{{lang 'common.required' }}</small>
         </label>
         <textarea name="contact_question" id="contact_question" rows="5" cols="50" class="form-input"></textarea>
     </div>
 
+    {{#if forms.contact.recaptcha.enabled}}
```
junedkazi: Moved the field down after the question description field since it is a practice that it should be the last element before the submit.

junedkazi:
@bigcommerce/stencil-team

templates/components/page/contact-us-form.html
mcampa: do we need to wrap it in the `if` block?
I think `forms.contact.recaptcha.markup` will be empty when is disabled. like in https://github.com/bigcommerce/cornerstone/pull/951/files#diff-ecbae6e2b7d5bbf5c950d68878e79d99R24

junedkazi: I would like to keep the captcha enabled check in place irrespective of what is returned in the markup field.

mcampa: That's not how we do this kind of things in Stencil, we removed most of all boolean Blueprint variables because we can check agains the data

lang/en.json
Jaymarkf: please have a v3

PR #955, opened 2017-03-10
STENCIL-2944 Load amp iframe only when we have product options.

junedkazi:
#### What?
Load amp-iframe script only when we have product options.

ping @bigcommerce/stencil-team @mcampa @bookernath


bookernath:
👍

PR #957, opened 2017-03-11
STENCIL-3262 - only load AMP-analytics if there's an ID

bookernath:
Currently we load the AMP-analytics module even if no ID is specified.

Although we want everyone to have an ID, I don't want there to be
errors on the storefront, so let's fix that.

https://jira.bigcommerce.com/browse/STENCIL-3262

### DIFF
--- a/templates/layout/amp.html
+++ b/templates/layout/amp.html
@@ -20,7 +20,9 @@
         {{{snippet 'htmlhead'}}}
         <script async custom-element="amp-form" src="https://cdn.ampproject.org/v0/amp-form-0.1.js"></script>
         <script async custom-element="amp-sidebar" src="https://cdn.ampproject.org/v0/amp-sidebar-0.1.js"></script>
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
 - Print a readable error instead of dumping the whole error object to the console [#950](https://github.com/bigcommerce/cornerstone/pull/950)
 - Fixed homepage featured products floating left and unecessarily wrapping to next row [#948](https://github.com/bigcommerce/cornerstone/pull/948)
 - Add google recaptcha v2 support to cornerstone. [#951](https://github.com/bigcommerce/cornerstone/pull/951)
+- Only load amp-analytics module when an analytics ID is specified. [#957](https://github.com/bigcommerce/cornerstone/pull/957)
```
junedkazi: Not sure we need new change log entry for for changes to a feature which is not yet released. We should just let all of this roll into the parent entry.

PR #958, opened 2017-03-13
STENCIL-3194: Adds theme editor setting for product display mode (grid vs list).

cristycarpenter:
#### What?
Adds a "Product display mode" setting to the theme editor so merchants can choose between a grid or list view for their products on category, brand, and search results pages.  Product list view functionality already exists in the theme. This change allows merchants to actually use it :)

#### Tickets / Documentation
https://jira.bigcommerce.com/browse/STENCIL-3194

Grid Mode (Default):
grid-view

List Mode:
list-view


nickdengler:
At larger screen sizes I'm seeing some clipping on the side of the button:
screen shot 2017-03-13 at 12 45 51 pm
It flickers on and off when the screen is large enough that the buttons appear on the side.
Not a showstopper, but if it's an easy fix, I'm still doing some testing on this.

bookernath:
@nickdengler I don't think this should block the release but let's create a JIRA for a fast-follow fix.

nickdengler:
Sounds good, logged a ticket. Otherwise looks 💚

PR #965, opened 2017-03-16
Add clearfix to product view details for AMP

bc-annavu:
#### What?
There is a bug where the image carousel in AMP will be tiny when there are additional product info added to the product information causes the carousel to resize itself aligned right caused by the floats not being cleared inside the above frame.

image

@mcampa @junedkazi

### DIFF
--- a/templates/components/amp/css/product.html
+++ b/templates/components/amp/css/product.html
@@ -119,6 +119,11 @@
 .productView-details {
     margin-bottom: 2rem
 }
+.productView-details:after {
+    content: "";
+    display: table;
+    clear: both;
+}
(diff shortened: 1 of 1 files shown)


bookernath:
Upon testing, this didn't fix the issue for me. Sent @bc-annavu an example.

junedkazi:
Do we still need this PR since we have this fixed in https://github.com/bigcommerce/cornerstone/pull/982

bookernath:
I don't think so. @bc-annavu ?

bc-annavu:
Isn't needed anymore. I'll close this PR.

PR #968, opened 2017-03-16
Load amp social share only when we have icons to share.

junedkazi:
#### What?
Load amp social share only when we have social icons.

##### Before
share_before
##### After
after


bookernath:
👍

junedkazi:
Just a rebase to resolve changelog conflicts.

PR #982, opened 2017-03-27
Fix AMP Carousel on Product Page

bc-annavu:
### What?
Restyled how product details are implemented on the product page.

### Why?
When there are additional details on the product page, the amp carousel
shrinks because of the way that the product details are styled.

### Proof/Testing

image

ping: @junedkazi @bookernath


bookernath:
💚

PR #983, opened 2017-03-27
STENCIL-3328 - Remove footer scripts from amp-iframe

bookernath:
Removing the footer.scripts helper from the amp-iframe so footer
scripts will NOT render in the amp product options iframe. This was
causing weird display of product options on some stores.

Ping @junedkazi


bookernath:
Tested on my store with a JS alert. Confirmed it works as intended.

PR #989, opened 2017-04-04
Fix invalid icon HTML in AMP templates

bc-annavu:
#### What?
AMP is throwing validation errors based on some of the markup we use for our alert messages. We need to change the markup to use valid HTML tags and add in the styles for them since they were missing.

image

@bookernath @junedkazi


bookernath:
💚

PR #996, opened 2017-04-13
STENCIL-3010 - Repopulate review form fields after error

mjschock:
#### What?

Fixes an issue where submitting an invalid form (usually because of the captcha) would result in the loss of the form data.  For example, the comments might need to be re-entered.

#### Tickets / Documentation

- [STENCIL-3010](https://jira.bigcommerce.com/browse/STENCIL-3010)

#### GIF showing Fix

http://g.recordit.co/MhD8igg03S.gif

@bigcommerce/stencil-team 

P.S. i opted for `if/else`'s here for the sake of readability - i'm open to using an `if` subexpression in-line if people prefer that

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Draft
+- Repopulate review form fields after error [#996](https://github.com/bigcommerce/cornerstone/pull/996)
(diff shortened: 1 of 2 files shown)


mjschock:
@junedkazi - changelog added

PR #999, opened 2017-04-21
Fix latest node-sass issues with Citadel upgrade and conditional impo…

mjschock:
…rt swap with mixin

#### What?

* Fixes the two node-sass issues that were preventing upgrading node-sass:
  * Points to the latest version of Citadel which includes a [fix](https://github.com/bigcommerce/citadel/commit/0e4cc29c9bcb3781cdcc109d6f62598556f2ca5a) to `_tabs.scss`.
  * Moves the conditional import of _clear-sans.scss to a conditional mixin inclusion instead

cc @bigcommerce/stencil-team

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Draft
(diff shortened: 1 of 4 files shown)


mcampa:
Add a changelog entry before merging it

mjschock:
changelog added, merging

Issue #1000, opened 2017-04-26
Category page Sidebar issue 


minankmori:
Hello,

We have noticed one issue with cornerstone theme category page.
When any category has no subcategories and shop by price filter available in the left sidebar. category products right bar section will be moved to the left side and show white space on the right sidebar.
Please check the issue on attach screenshot.
![2017-04-26_2021](https://cloud.githubusercontent.com/assets/26482645/25440764/02f38828-2abe-11e7-8c70-0af23c33485c.png)

You can fix this issue by adding more condition on the sidebar so when sidebar has no section available you can make the category products section full width and show 4 products per row to fill that white space.

Thanks,
Minank Mori

sacr3dc0w:
Using the {{or}} operator made for a very simple solution. 👍

sacr3dc0w:
#1001 didn't consider 'shop by price', #1018 does.

PreciselyAlyss:
Closing based on @sacr3dc0w's fixes. Please re-open if necessary.

PR #1001, opened 2017-04-29
Category sidebar display & copyright year in footer

sacr3dc0w:
When neither faceted search is enabled, or a subcategory is not present, the main content area remains at 75% width. This is due to the styling of .page-sidebar + .page-content. Another issue is three products display display in a row.

Using the {{or}} operator, verify if child category exists or faceted
search is enabled. If neither is true, page content will stretch 100%
and display four products by default.

Instead of using an extra JavaScript call on the client side, let's use {{moment}} to display the copyright year.

[https://github.com/bigcommerce/cornerstone/issues/1000](https://github.com/bigcommerce/cornerstone/issues/1000)
[https://stencil.bigcommerce.com/docs/handlebars-helpers-reference](https://stencil.bigcommerce.com/docs/handlebars-helpers-reference)

### DIFF
--- a/templates/components/common/footer.html
+++ b/templates/components/common/footer.html
@@ -83,11 +83,8 @@
         {{/if}}
         {{#if theme_settings.show_copyright_footer}}
             <div class="footer-copyright">
-                <p class="powered-by">&copy; <span id="copyright_year"></span> {{settings.store_name}} </p>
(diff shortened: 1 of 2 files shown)


mcampa:
Thank you 👍

PR #1006, opened 2017-05-05
Revert "Category sidebar display & copyright year in footer"

junedkazi:
Reverts bigcommerce/cornerstone#1001

This has caused a side effect in which "shop by price" filter which was showing without faceted search no longer shows up.

https://jira.bigcommerce.com/browse/STENCIL-3438

@bigcommerce/stencil-team @shanth100 @bc-jason @lord2800


mcampa:
Can we revert only the `or` issue part?

Issue #993, opened 2017-04-10
Blurry text and images in quick view modals


ocvninfo:
As i mentioned on issue heading.
The quick view popup is blurry.
![image](https://cloud.githubusercontent.com/assets/3072481/24854951/4910a0c2-1e0a-11e7-8c49-548d987ccc83.png)

other popup same problem:
![image](https://cloud.githubusercontent.com/assets/3072481/24855023/7349e2b8-1e0a-11e7-9aa5-62959dfa27c0.png)

Is there a way to fix this issue?

bc-AlyssNoland:
Based on [this](https://github.com/bigcommerce/cornerstone/blob/master/templates/components/products/product-view.html#L159), it should be using the default product image in the quick view modal. Have you tried uploading a higher resolution photo to see if that clears up the problem, @ocvninfo?

ocvninfo:
@bc-AlyssNoland Thank for your reply. But what i mean is blurry all text in the popup.
The popup work fine on Firefox but blurry on Chrome.

EdFromBJCS:
@bc-AlyssNoland @ocvninfo I have noticed this as well. The title of this issue is misleading because it is not just pictures, text and other elements are blurry as well. It's subtle, but unmistakable. Looks kind of like the lossy analog picture you get from a VGA-connected monitor. Strange.

ThemeValeBC:
@EdFromBJCS I checked on firefox, The quick view popup working correctly. But it is blurry on chrome

sacr3dc0w:
It would seem that translate(-50%, -50%) is calculating a pixel value on both axes with a decimal place, causing subpixel rendering. One solution could be subtracting 0.5px from the Y-axis. That solved the problem in Chrome and Safari based off my testing.

ocvninfo:
@sacr3dc0w  Thank you for this.
But it still blurry when popup resizing.

Ubersmake:
Closing this issue as I can no longer reproduce it. There are image-related PRs in #1094, #1096, #1104, and #1358. As some of the comments above mentioned, this could have also been a browser-related bug that's been fixed since this issue was first reported.

PR #1010, opened 2017-05-11
Webkit fix: Blurry elements in modal windows

sacr3dc0w:
See issue #993.

translate(-50%, -50%) is calculating a pixel value on both axes with a
decimal place, causing subpixel rendering on elements in quick view
modal windows. Removing half a pixel off the Y-axis seems to solve this.


junedkazi:
@sacr3dc0w can you share before & after screenshots ?

Issue #966, opened 2017-03-16
Product Images can obscure product details


denmch:
Product images on some narrow viewports (e.g., a small browser window on a desktop) can cause `.productView-image` to obscure the left side of `.productView-details` above the medium viewport (by default `min-width: 801px`).

![screenshot 2017-03-15 23 28 32](https://cloud.githubusercontent.com/assets/2019665/23984528/27ae9ae8-09d7-11e7-8505-57835cca1a0c.png)

This is because `.productView-image` is set to both `display: flex` and `min-width: 500px`. The image then fills the available space, which will be no less than 500px, even though the container section (`.productView-images`, which houses the main image and the thumbnails) is set to a width of 50%. The image needs to be able to display smaller than 500px wide, since 50% of the smallest viewport width under this rule is just above 400px.

This could be corrected at line 27 and following of `_productView.scss` by one of the following:

1. resetting the figure's display to its initial setting via `display: initial`
2. removing the `min-height` and `min-width` values
3. changing those values to something like 400px (though preferably by reference to the value of the medium breakpoint so that it adapts if that changes)

```
  @include breakpoint("medium") {
      display: initial; // option 1
      min-height: 400px; // option 2
      min-width: 400px;
    }
```

Reducing the minimum height and width is a safe route, and that's what I've done on the site above

I'm not sure I see the practical benefit of setting minimum image sizes, so I think removing them makes sense, but I'd like to hear thoughts on the best solution or arguments in support of minimum widths if there are any to be made. Let me know if there's something I haven't considered, and if it's agreed that one of these solutions fixes the issue without breaking something or causing side effects, I'll be happy to issue a pull request.

sacr3dc0w:
Setting the display to initial breaks the container, exposing the zoomed product image to the entire viewport. #1019 sets the minimum height/width of `.productView-image` to 366px. This is the minimum before there's no overlap of product details.

#### Screenshots

The red border is just for demonstration purposes.

**display: initial (breaking out of container)**
![Initial](http://i.imgur.com/wKW1jDm.jpg)

**Before**
![Before](http://i.imgur.com/rROzVUs.png)

**After**
![After](http://i.imgur.com/D59P4si.png)

PreciselyAlyss:
Closing due to fix

PR #1019, opened 2017-06-14
Product Images obscure product details on smaller viewports

sacr3dc0w:
Product image obscures product details when scaling down the viewport.

#### What?

A minimum width and height of 366px solves the overlap issue. Setting
display to initial as an alternative solution breaks product zoom.

#### Tickets / Documentation

See [issue #966](https://github.com/bigcommerce/cornerstone/issues/966)

#### Screenshots

The red border is just for demonstration purposes.

**`display: initial` breaking out of container**
Initial

**Before**
Before

**After**
After

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Draft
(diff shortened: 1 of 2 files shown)


junedkazi:
@mcampa

PR #1029, opened 2017-06-20
STENCIL-2599: Adds on/off toggle for Shop by Price.

cristycarpenter:
#### What?

Adds an on/off toggle to the theme editor for the "Shop by Price" panel. One important item to note is when the setting is toggle off, the products on the page show 3 per row which looks a bit funny. This is because the aside element (.page-sidebar) is still present. See video below. I am not sure why. This expression tells it to show if category has subcategories or faceted search or shop by price:

https://github.com/bigcommerce/cornerstone/blob/master/templates/pages/category.html#L27

The faceted_search_enabled property is returning false:

debug

Maybe it's still showing because we're not telling it there are no subcategories? Let me know if you have ideas. Maybe @bookernath, @PascalZajac, @junedkazi, @mcampa 

#### Tickets / Documentation

- https://jira.bigcommerce.com/browse/STENCIL-2599

https://www.screencast.com/t/fEuoOvpn

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Draft
 
+## 1.8.1 (2017-05-05)
+
 ## 1.8.0 (2017-05-04)
 - Add storefront admin bar to replace store not launched preview panel [#997](https://github.com/bigcommerce/cornerstone/pull/997)
 
--- a/config.json
+++ b/config.json
@@ -1,6 +1,6 @@
 {
   "name": "Cornerstone",
-  "version": "1.8.0",
+  "version": "1.8.1",
   "meta": {
     "price": 0,
     "documentation_url": "https://support.bigcommerce.com/articles/Public/Cornerstone-Theme-Manual",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "bigcommerce-cornerstone",
   "description": "The BigCommerce reference theme for the Stencil platform",
-  "version": "1.8.0",
+  "version": "1.8.1",
   "private": true,
   "author": "BigCommerce",
   "license": "MIT",


templates/components/category/shop-by-price.html
```diff
@@ -1,4 +1,4 @@
-{{#if shop_by_price}}
+{{#if theme_settings.shop_by_price}}
```
mcampa: you need to keep the original `if`

```
{{#if shop_by_price}}
```

cristycarpenter: Derp. Completely overlooked that. I'll get it added as well as update the changelog. Thanks, Mario!

templates/components/category/shop-by-price.html
```diff
@@ -1,4 +1,4 @@
-{{#if shop_by_price}}
+{{#if shop_by_price }}
```
junedkazi: I don't think so we should be adding that space in the end.

templates/pages/category.html
```diff
@@ -24,12 +24,18 @@ <h1 class="page-heading">{{category.name}}</h1>
...
         </aside>
+    {{else}}
+        {{#if theme_settings.shop_by_price_visible}}
+            <aside class="page-sidebar" id="faceted-search-container">
+                {{> components/category/sidebar}}
+            </aside>
+        {{/if}}
     {{/or}}
-
+
```
junedkazi: Is this whitespace ?

templates/pages/category.html
```diff
@@ -40,4 +46,4 @@ <h1 class="page-heading">{{category.name}}</h1>
 </div>
 
 {{/partial}}
-{{> layout/base}}
+{{> layout/base}}
```
junedkazi: End of line is missing.

PR #1035, opened 2017-07-02
Replace JavaScript alert/confirmations with sweetalert2

sacr3dc0w:
#### What?

I’ve used this for a few projects and wanted to make it accessible to the community. I’ve cleaned it up and done my best to make it conform to Cornerstone stylings but maybe some additional eyes can fill in any missing blanks. SweetAlert2 is described by their webpage as:

> A beautiful, responsive, customizable, accessible (WAI-ARIA) replacement for JavaScript's popup boxes. Zero dependencies.

#### Documentation

The project’s webpage goes into deeper detail on it’s usage.

- [https://limonte.github.io/sweetalert2/](https://limonte.github.io/sweetalert2/)

##### Example

**_Remove item from cart, desktop (all theme variations)_**
desktop

**_Remove item from cart, tablet (all theme variations)_**
tablet

**_Remove item from cart, mobile (all theme variations)_**
mobile 

**_Compare page error_**
mobile 

**_Compare product error on category page_**
mobile


mjschock:
@sacr3dc0w overall it looks good to me.  my concern is the file size:

screen shot 2017-07-05 at 3 47 45 pm

i see nearly 50k for this.  however, it turns out sweetalert2's package main is pointing to `dist/sweetalert2.js`.  modifying that to point to `dist/sweetalert2.min.js` gets a more reasonable profile:

screen shot 2017-07-05 at 3 51 46 pm

can you please add another alias in `webpack.conf.js` to map to the min file, like so:
```
sweetalert2:  path.resolve(__dirname, 'node_modules/sweetalert2/dist/sweetalert2.min.js'),
```

assets/js/theme/global/compare-products.js
```diff
@@ -57,7 +58,10 @@ export default function (urlContext) {
         const productsToCompare = $this.find('input[name="products\[\]"]:checked');
 
         if (productsToCompare.length <= 1) {
-            alert('You must select at least two products to compare');
+            swal({
+                text: 'You must select at least two products to compare',
```
davidchin: This is an existing problem. So you probably don't need to fix it in this PR. Just pointing it out that the warning message needs to come from the language file, not hard-coded in here.

assets/js/theme/cart.js
```diff
@@ -163,17 +179,23 @@ export default class Cart extends PageManager {
...
+            const result = swal({
+                text: 'Are you sure you want to remove this item from your cart?',
+                type: 'warning',
+                showCancelButton: true,
+                confirmButtonText: 'Yes, delete it!',
+            }).then(() => {
+                const delta = new Date() - openTime;
+
+                // Delta workaround for Chrome's "prevent popup"
+                if (!result && delta > 10) {
```
davidchin: A question. Do we need this hack anymore? It looks like `sweetalert2` is just a regular modal. Unlike `window.confirm`, the browser might not interfer.

assets/scss/settings/vendor/sweetalert2/_settings.scss
```diff
@@ -0,0 +1,64 @@
...
+ */
+$swal-font-body:                                      stencilFontFamily("body-font"), Arial, Helvetica, sans-serif;
+$swal-font-bodySize:                                  remCalc(stencilNumber("fontSize-root", "px") + 2px);
+$swal-font-bodyColor:                                 stencilColor("color-textBase");
+$swal-font-headings:                                  stencilFontFamily("headings-font"), Arial, Helvetica, sans-serif;
+$swal-font-headingsSize:                              remCalc(stencilNumber("fontSize-h1", "px") + 2px);
+$swal-font-headingsColor:                             stencilColor("color-textHeading");
+
+// Modal stylings
+$swal-modal-bgColor:                                  stencilColor("navUser-dropdown-backgroundColor") !important;
```
davidchin: You should probably introduce a new variable instead of using `navUser-dropdown-backgroundColor`? It looks like `navUser-dropdown-backgroundColor` is intended to used for navigation dropdowns only.

assets/scss/settings/vendor/sweetalert2/_settings.scss
```diff
@@ -0,0 +1,64 @@
...
+ * !important usage isn't ideal but this prevents needless
+ * declaration of the button color for each swal() instance.
+ * -----------------------------------------------------------------------------
+ */
+
+// Wrapper
+$swal-button-wrapperMargin:                           25px 0 0;
+
+// Confirm button
+$swal-button-bgColor:                                 stencilColor("button--primary-backgroundColor") !important;
```
davidchin: Do you really need to use `!important` here? If the styles are cascaded properly, you shouldn't need to increase their specificity.

assets/js/theme/common/state-country.js
```diff
@@ -129,7 +130,10 @@ export default function (stateElement, context = {}, options, callback) {
 
         utils.api.country.getByName(countryName, (err, response) => {
             if (err) {
-                alert(context.state_error);
+                swal({
```
davidchin: Do you want to only execute the callback after dismissing the alert? I think `sweetalert2` is non-blocking, unlike `alert`, which stops the execution the script until it is dismissed.

junedkazi:
This is nice but I have been thinking if it makes sense to move this functionality to https://github.com/bigcommerce/stencil-utils so we can build a wrapper around it. It will make it much more easier to change the underlying library in the future & also to upgrade the lib.

mjschock:
i like @junedkazi 's idea of abstracting it as `utils.alert` from stencil.utils.  i'm not sure where exactly it would fit in there though - perhaps in a `components` folder?  it would be nice down the line to have the specifics of the theme 'swappable' so to speak.  maybe we could even ultimately swap in different frameworks.

config.json
```diff
@@ -177,6 +177,7 @@
     "checkRadio-color": "#4f4f4f",
     "checkRadio-backgroundColor": "#ffffff",
     "checkRadio-borderColor": "#dfdfdf",
+    "sweetAlert-backgroundColor": "#ffffff",
```
davidchin: Can we use a more generic config name? This name ties to a specific vendor library. Maybe just `alert-backgroundColor`.

assets/scss/settings/vendor/sweetalert2/_settings.scss
```diff
@@ -0,0 +1,54 @@
...
+// -----------------------------------------------------------------------------
+
+
+//
+// Purpose: Settings for the vendor popup box replacement, SweetAlert2
+//
+// -----------------------------------------------------------------------------
+
+// Font color, size, family
+$swal-font-body:                                      stencilFontFamily("body-font"), Arial, Helvetica, sans-serif;
```
davidchin: All these variables can probably be generalised as well. i.e.: `$alert-font-body`

assets/js/theme/account.js
```diff
@@ -105,7 +107,10 @@ export default class Account extends PageManager {
 
             if (!submitForm) {
                 event.preventDefault();
-                alert('Please select one or more items to reorder.');
+                swal({
```
davidchin: Can we have a intention-revealing function name? `swal` doesn't communicate anything.

assets/js/theme/account.js
```diff
@@ -5,6 +5,7 @@ import Wishlist from './wishlist';
 import validation from './common/form-validation';
 import stateCountry from './common/state-country';
 import { classifyForm, Validators, insertStateHiddenField } from './common/form-utils';
+import swal from 'sweetalert2';
```
davidchin: You may choose to not import the vendor library directly. Instead, create a facade so that it's easier to change its implementation later.

sacr3dc0w:
I'll have to make changes assuming #1040 is approved/merged first.

mjschock:
@sacr3dc0w - could you please update the changelog to include an entry for these changes in the drafts section - i'll merge it in after that

limonte:
> https://limonte.github.io/sweetalert2/

this link is outdated, here's the correct one https://sweetalert2.github.io/

PR #1037, opened 2017-07-06
Reduce theme bundle size by using specific minified libraries

sacr3dc0w:
#### What?

In an effort to reduce the size of the theme bundle, I've pointed two libraries to their minified versions.

**Bundle, before history.min.js**
1_sourcemap-visualizer-history-before

**Bundle, after history.min.js**
2_sourcemap-visualizer-history-after

**Bundle, before slick.min.js**
3_sourcemap-visualizer-slick-before

**Bundle, after slick.min.js**
4_sourcemap-visualizer-slick-after

**Bundle, after both history.min.js and slick.min.js**
5_sourcemap-visualizer-both-after

### DIFF
--- a/assets/js/theme/common/carousel.js
+++ b/assets/js/theme/common/carousel.js
@@ -1,5 +1,5 @@
 import $ from 'jquery';
-import 'slick-carousel/slick/slick';
+import 'slick-carousel/slick/slick.min';
 
 export default function () {
     const $carousel = $('[data-slick]');
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Product Images were obscuring product details on smaller viewports [#1019](https://github.com/bigcommerce/cornerstone/pull/1019)
 - Add region tags to two template files to support payments team banner integration with content service [#1023](https://github.com/bigcommerce/cornerstone/pull/1023)
 - Fix H1-H6 font-sizing [#1034](https://github.com/bigcommerce/cornerstone/pull/1034)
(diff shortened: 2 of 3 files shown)


PreciselyAlyss:
@junedkazi are you comfortable with one of us merging this as it now has a changelog entry?

junedkazi:
@PreciselyAlyss we plan on cutting a new release today. So will merge it after that is done. Since we don't want to add any new scopes to the testing effort which is based on the current master.

assets/js/theme/common/carousel.js
```diff
@@ -1,5 +1,5 @@
 import $ from 'jquery';
-import 'slick-carousel/slick/slick';
+import 'slick-carousel/slick/slick.min';
```
mjschock: @sacr3dc0w - this is great!  can we put them in the `webpack.conf.js` resolve instead:

```
    resolve: {
        alias: {
            html5-history-api: path.resolve(__dirname, 'node_modules/html5-history-api/history.min',
            jquery: path.resolve(__dirname, 'node_modules/jquery/dist/jquery.min.js'),
            jstree: path.resolve(__dirname, 'node_modules/jstree/dist/jstree.min.js'),
            slick-carousel: path.resolve(__dirname, 'node_modules/slick-carousel/slick/slick.min'),
        },
    },
```

that way the imports can simply be:

```
import 'slick-carousel';
...
import 'html5-history-api'
```

seeing as this was already merged, feel free to open another PR

mjschock: @sacr3dc0w i'll just go ahead and create a quick pr

PR #1045, opened 2017-07-18
Add RDFa tags

hennell-git:
**What?**

Add the RDFa BreadcrumbList markup to the generated breadcrumbs to help Google (and other web crawlers) understand the shop categories better.

**Documentation**

Follows the schema.org [breadcrumb markup guidelines](http://schema.org/BreadcrumbList), and passes Google's [structured data testing tool](https://search.google.com/structured-data/testing-tool). Both appear to suggest using OL instead of UL for the list element, but it validates as is and sticking with UL seems preferable to avoid potential css issues.

**Screenshots**
Google's testing tool before the markup - no breadcrumbs only product data detected:
screenshot 2017-07-14 14 01 10
Same page after the markup - breadcrumbs now found:
screenshot 2017-07-14 14 01 24

### DIFF
--- a/templates/components/common/breadcrumbs.html
+++ b/templates/components/common/breadcrumbs.html
@@ -1,11 +1,12 @@
-<ul class="breadcrumbs">
+<ul class="breadcrumbs" vocab="http://schema.org/" typeof="BreadcrumbList">
     {{#each breadcrumbs}}
-        <li class="breadcrumb {{#if @last}}is-active{{/if}}">
+        <li property="itemListElement" typeof="ListItem" class="breadcrumb {{#if @last}}is-active{{/if}}">
             {{#if url}}
-                <a href="{{url}}" class="breadcrumb-label">{{name}}</a>
+                <a property="item" typeof="WebPage" href="{{url}}" class="breadcrumb-label"><span property="name">{{name}}</span></a>
             {{else}}
-                <span class="breadcrumb-label">{{name}}</span>
+                <span property="name" class="breadcrumb-label">{{name}}</span>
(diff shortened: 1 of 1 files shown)


junedkazi:
@hennell-git thank you for your contribution. Can you please give some information about the 4 errors I see in the screenshot ?

hennell-git:
The 4 errors are part of the product markup section - this fix only adds the detected BreadcrumbList section which is error free!

(I was actually investigating our product section errors originally, but that's more complex as the product data might not be supplied by the user which leads to the errors above as data is marked but missing - added in the breadcrumbs code first as it's entirely database generated so can't have those same 'missing elements' issues.)

davidchin:
Nice work! Thanks for making this change. I have a question for you. I think we use Microdata instead of RDFa in this project. Is there a reason why you choose to use RDFa? If there isn't a special reason, I'd like to keep the format consistent.

hennell-git:
I'd like to say there's a complex technical reason, but it's actually just based on the fact I've only ever used RDFa, and didn't really think about it or notice the existing product tagging was in microdata!

No idea if there's any actual significant difference between the two and evidently they can work together, but I'd agree it would make sense for the format to be consistent across the whole project.

davidchin:
Ah ok cool, yeah when you have time, can you please update it? Thanks!

abrambailey:
Hey guys, didn't realize this was a thing, and just submitted a json ld update here https://github.com/bigcommerce/cornerstone/pull/1079

abrambailey:
image

PR #1056, opened 2017-07-27
STENCIL-3380: Adds auto height to alternative nav menu and submenus.

cristycarpenter:
#### What?

This change ensures the alternative navigation menu and its submenus always have the proper background length no matter how many categories a merchant has. 

#### Tickets / Documentation

https://jira.bigcommerce.com/browse/STENCIL-3380

Before/After for Main Menu:
https://www.screencast.com/t/Dueilwqbp

Before/After for Submenus:
https://www.screencast.com/t/qOgJbQnM

### DIFF
--- a/assets/scss/components/stencil/navPages/_navPages.scss
+++ b/assets/scss/components/stencil/navPages/_navPages.scss
@@ -308,6 +308,7 @@
 .has-subMenu.is-root + .navPage-subMenu-horizontal {
     @include breakpoint("medium") {
         border-left: 0;
+        height: $navPage-subMenu-horizontal-height;
         left: inherit;
         margin: auto;
         top: inherit;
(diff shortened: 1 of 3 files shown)


cristycarpenter:
Oops...need to change the position of the height property.

assets/scss/components/stencil/navPages/_navPages.scss
```diff
@@ -320,7 +321,7 @@
         @include breakpoint("medium") {
             border-left: 1px solid stencilColor("navPages-subMenu-separatorColor");
             display: inherit;
-            height: $navPage-subMenu-horizontal-height;
```
junedkazi: Is this variable being used anywhere else ? Or can we remove the definition as well completely ?

cristycarpenter: I didn't see that it was being used anywhere else but was hesitant to remove it myself. If you think it's best, I can go ahead with it.

mcampa: Instead of removing this variable please change the value of the variable to `auto`, in `assets/scss/settings/stencil/navPages/_settings.scss`

mcampa: ... and yes, it is only used here

cristycarpenter: Will do. Thanks, Mario.

cristycarpenter: Got the changes made. Need to squash my commits still. Will do so in the morning :)

PR #1058, opened 2017-08-01
STENCIL-3536 bulk discount rates

mcampa:
Add bulk discount rates to product cards

https://jira.bigcommerce.com/browse/STENCIL-3695

bulk-discount-modal

@bigcommerce/stencil-team

### DIFF
--- a/config.json
+++ b/config.json
@@ -275,6 +275,7 @@
   ],
   "resources": {
     "cart": true,
+    "bulk_discount_rates": false,
     "shop_by_brand": {
       "limit": 10
     }
--- a/templates/components/products/card.html
+++ b/templates/components/products/card.html
@@ -82,5 +82,6 @@
                 {{> components/common/login-for-pricing}}
             {{/or}}
         </div>
+        {{> components/products/bulk-discount-rates}}
     </div>
 </article>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Draft
 - Hide Info in footer if no address is provided in Store Profile. Hide Brands in footer if Merchant has no brands [#1053](https://github.com/bigcommerce/cornerstone/pull/1053)
 - Product illustrations in the storefront when the product catalog is empty [#1054](https://github.com/bigcommerce/cornerstone/pull/1054)
+- Add bulk discount rates to product cards [#1058](https://github.com/bigcommerce/cornerstone/pull/1058)
 
 ## 1.9.1 (2017-07-25)
 - Move some hard-coded validation messages to language file [#1040](https://github.com/bigcommerce/cornerstone/pull/1040)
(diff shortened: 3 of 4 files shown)


templates/components/products/card.html
```diff
@@ -82,5 +82,40 @@ <h4 class="card-title">
                 {{> components/common/login-for-pricing}}
             {{/or}}
         </div>
+        {{#if bulk_discount_rates.length}}
```
junedkazi: I think it would be nice if we moved this into a component.

templates/components/products/card.html
```diff
@@ -82,5 +82,40 @@ <h4 class="card-title">
...
+                    <h2 class="modal-header-title">{{lang 'products.bulk_pricing.modal_title'}}</h2>
+                    <a href="#" class="modal-close" aria-label="Close"><span aria-hidden="true">&#215;</span></a>
+                </div>
+                <div class="modal-body">
+                    <p>{{lang 'products.bulk_pricing.instructions'}}</p>
+                    <ul>
+                    {{#each bulk_discount_rates}}
+                        <li>
+                            {{lang 'products.bulk_pricing.range' min=min max=max}}
+                            {{#if type '===' 'percent'}}
```
junedkazi: from where are we getting this `type` ?

templates/components/products/card.html
mcampa: yes, I will

templates/components/products/card.html
mcampa: from `{{#each bulk_discount_rates}}`
each creates a new context for each block

templates/components/products/card.html
```diff
@@ -82,5 +82,7 @@ <h4 class="card-title">
                 {{> components/common/login-for-pricing}}
             {{/or}}
         </div>
+        {{> components/common/login-for-pricing}}
```
junedkazi: We already display the login for pricing above. Why do we need it here ?

templates/components/products/card.html
```diff
@@ -82,5 +82,7 @@ <h4 class="card-title">
...
+        {{> components/products/bulk-discount-rates}}
```
junedkazi: Do you think this needs to be wrapped in a setting ? 🍹

config.json
```diff
@@ -275,6 +275,7 @@
   ],
   "resources": {
     "cart": true,
+    "bulk_discount_rates": true,
```
junedkazi: Does this mean bulk discounts will be loaded on all pages ?

mcampa: Yes. All pages that have product context. But I'm going to change it to default to `false`

templates/components/products/card.html
mcampa: Is already wrapped, in the component file

templates/components/products/card.html
mcampa: My mistake, this shouldn't be here

nnandani:
Right now not working on 2.0.0

PR #1060, opened 2017-08-04
STENCIL-3658 - Fix carousel next arrows

mjschock:
#### What?

Fixes an issue in Firefox and IE where the svg doesn't allow properties with hashes (#).  Instead of pulling in the fill color directly, it's passed through the new `str-replace` transformation first, replacing '#' with '%23'.

#### Tickets / Documentation

- [STENCIL-3658](https://jira.bigcommerce.com/browse/STENCIL-3658)

#### Screenshots

before:
screen shot 2017-08-04 at 2 59 08 pm

after:
screen shot 2017-08-04 at 2 58 36 pm

cc @bigcommerce/stencil-team @shanth100 @davidchin


mjschock:
here's some further information:

try loading this in Firefox:

data:image/svg+xml;charset=utf8,<svg%20xmlns='http://www.w3.org/2000/svg'%20width='24'%20height='42'%20viewBox='0%200%2024%2042'><path%20d='M1.45679%201.00746147l21%2020.02482143L1.50885%2041.0074615'%20stroke='#989898'%20stroke-width='2'%20stroke-linecap='round'%20stroke-linejoin='round'%20fill='none'%20fill-rule='evenodd'%20/%3E%3C/svg%3E

you'll see this:
screen shot 2017-08-04 at 3 02 49 pm

then replace the # with a %23 and try it again:

data:image/svg+xml;charset=utf8,<svg xmlns='http://www.w3.org/2000/svg' width='24' height='42' viewBox='0 0 24 42'><path d='M1.45679 1.00746147l21 20.02482143L1.50885 41.0074615' stroke='%23989898' stroke-width='2' stroke-linecap='round' stroke-linejoin='round' fill='none' fill-rule='evenodd' /%3E%3C/svg%3E

you'll then see the arrow load:

screen shot 2017-08-04 at 3 03 19 pm

mjschock:
@davidchin - wondering if you might have a better suggestion than this fix?  it's literally the `#` that breaks Firefox.  i found that using `rgba` gets it working but not sure if we want to set the alpha to 0.9...

shanth100:
💚

davidchin:
I suggest to encode the data URI so that it is actually valid. You might choose to write your own SASS function, or use an existing one. i.e.: https://www.npmjs.com/package/sass-svg-uri. So that your fix doesn't change the appearance of the arrow. What do you think?

PR #1069, opened 2017-08-21
STENCIL-3672 - Upgrade dependencies

mjschock:
#### What?

* upgrades many of the dependencies (jquery and foundation major bumps have been left out, for example)

#### Tickets / Documentation

- [STENCIL-3672](https://jira.bigcommerce.com/browse/STENCIL-3672)

cc @bigcommerce/stencil-team


.eslintrc
```diff
@@ -9,6 +9,22 @@
         "no-alert": 0,
         "consistent-return": 0,
         "max-len": 0,
+        "import/first": 0,
```
mjschock: these new additions are here so that we don't need to fix all these lint issues from the new version of airbnb's linter which marks them as errors by default

assets/js/theme/common/faceted-search.js
```diff
@@ -156,7 +156,7 @@ class FacetedSearch {
         const id = $navList.attr('id');
 
         // Toggle depending on `collapsed` flag
-        if (_.contains(this.collapsedFacetItems, id)) {
+        if (_.includes(this.collapsedFacetItems, id)) {
```
mjschock: lodash got rid of `_.contains` in favor of `_.includes`

package.json
```diff
@@ -5,58 +5,60 @@
   "private": true,
   "author": "BigCommerce",
   "license": "MIT",
+  "dependencies": {
+    "@bigcommerce/stencil-utils": "^1.0.7",
```
mjschock: webpack now complains about all of these deps being in devDeps and says to move them here

package.json
```diff
@@ -5,58 +5,60 @@
...
     "es6-shim": "^0.35.3",
-    "eslint-config-airbnb": "^6.0.2",
-    "fastclick": "^1.0.6",
-    "foundation-sites": "^5.5.3",
-    "grunt": "^0.4.5",
-    "grunt-eslint": "^18.0.0",
-    "grunt-karma": "^0.12.2",
+    "eslint": "^4.5.0",
+    "eslint-config-airbnb": "^15.1.0",
+    "eslint-plugin-react": "^7.3.0",
```
mjschock: while we're not using react, the airbnb package requires it as a peer dep

.eslintrc
mcampa: Before adding this, can you try using `eslint --fix` option? to see what can be automatically fixed

mjschock: @mcampa i ran through and tried `fix` after removing each of these one-by-one and was able to auto-fix a few of them.  we could create tickets or issues to go in and correct the others

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Do not scale product thumbnail images [#1094](https://github.com/bigcommerce/cornerstone/pull/1094)
 - Lazy load carousel images [#1090](https://github.com/bigcommerce/cornerstone/pull/1090)
 - Theme Editor menu item updates for ease of use [#1091] (https://github.com/bigcommerce/cornerstone/pull/1091)
+- Upgrades all dependencies except for Foundation and jQuery [#1069](https://github.com/bigcommerce/cornerstone/pull/1069)
```
junedkazi: @mjschock can you fix the line above & remove the space between `[#1091] (https://github.com/bigcommerce/cornerstone/pull/1091)` 🙏

package.json
```diff
@@ -5,58 +5,63 @@
...
+    "foundation-sites": "^5.5.3",
+    "grunt": "^0.4.5",
+    "html5-history-api": "^4.2.7",
+    "jquery": "^2.2.4",
+    "jquery-zoom": "^1.7.15",
+    "jstree": "vakata/jstree",
+    "lazysizes": "3.0.0",
+    "lodash": "^4.17.4",
+    "nod-validate": "^2.0.12",
+    "pace": "hubspot/pace#a03f1f1de62c9cea6c88b2267b8d7a83858b6cb6",
```
junedkazi: Do we know why pace is locked down to a commit & not to an actual release ?

CHANGELOG.md
mjschock: @junedkazi 👍

package.json
mjschock: @junedkazi IIRC it's because they made a breaking change.  i'll do a quick check to see if we can update now

mjschock: yup, there's a breaking change.  let's just create a separate PR with the fix

mjschock:
@junedkazi regarding the icon-defs, yeah i believe it should change this time since grunt-svgstore has been updated.  i just checked out the version from master and ran grunt and it gets updated

PR #1079, opened 2017-08-27
Add JSON LD Schema to breadcrumbs

abrambailey:
This is currently missing from Cornerstone. Please implement.

### DIFF
--- a/templates/components/common/breadcrumbs.html
+++ b/templates/components/common/breadcrumbs.html
@@ -1,3 +1,22 @@
+<script type="application/ld+json">
+    {
+        "@context": "http://schema.org/", 
+        "@type": "BreadcrumbList", 
+        "itemListElement": [
+            {{#each breadcrumbs}}
+                { 
+                    "@type": "ListItem", 
+                    "position": "{{add @index 1}}", 
+                    "item": { 
+                        "@id": "{{url}}", 
+                        "name": "{{name}}" 
+                    } 
+                }{{#if @last}}{{else}},{{/if}}
+            {{/each}}
+        ]
+    }
+</script>
+
 <ul class="breadcrumbs">
     {{#each breadcrumbs}}
         <li class="breadcrumb {{#if @last}}is-active{{/if}}">


PreciselyAlyss:
@abrambailey Thanks for the pull request!

Can you provide more details about the expected behavior and how to test the changes? 

A What/Why/How to test format is great :)

abrambailey:
@PreciselyAlyss Please confirm expected behavior for any page with breadcrumbs by copy pasting the page url or the html source into https://search.google.com/structured-data/testing-tool/u/0/

This is correct behavior:

image

Why? Because it helps Google understand your content, and SEO optimization, etc.

abrambailey:
@davidchin JSON LD is Google's preferred form of schema markup. Not sure why you guys aren't using it?

davidchin:
Hey @abrambailey, thanks for your explanation. Yeah I'm aware of the fact that Google currently prefers JSON-LD for various reasons. I _think_ it wasn't their preference when we first started building this theme. Back then, it wasn't very well supported; search engines such as Bing didn't support it at all. And it seems it's still not supported by Bing and other less sophisticated search engines.

On the other hand, Microdata is supported by most search engines because it has been around for a long time and it's just plain HTML attributes. Google also doesn't penalise the use of Microdata and it can parse it just as well as JSON-LD. Although I agree JSON-LD has many advantages over Microdata, but I think for now, Microdata is a better choice.

In the future, if JSON-LD becomes far superior or widely supported, or Microdata becomes deprecated, we are open to switch to JSON-LD. In that case, I'd like to make an effort to switch all structured data to JSON-LD, instead of just doing it ad-hoc, so that everything is consistent.

I know you put in some effort making this work. Sorry about that. Would it take long for you to update this PR so that it fits our current choice for structured data?

abrambailey:
Thanks @davidchin that makes sense. I suppose I haven't really been too worried about Bing, etc.. so missed that before. Don't worry about the work I put in, I am able to use the JSON-LD for my project, which is what matters. I just thought I'd do the PR to save you all some trouble. Good luck :)

PR #1084, opened 2017-09-04
PAYMENTS-1160: Ensure Apple Pay button is hidden for incompatible browsers.

PascalZajac:
#### What?
Remove an unnecessary `display: block` rule that was causing the Apple Pay button to appear in the Cart Modal even for browsers that do not support Apple Pay.

#### Why?
Shopper confusion. The rule in question is ignoring the `apple-pay-supported` class that gets attached to the `body` element, forcing the button to appear in all browsers. My first solution was to fix this by changing the nesting of the rules to respect this class, but then I realised that the `display: block` rule itself is unnecessary, because it will be inherited from the rule defined on L33. 

Since I can't enable Apple Pay for local development, I have screenshots showing the HTML structure in Chrome and Safari respectively to prove that this should work.
screen shot 2017-09-04 at 3 53 42 pm
screen shot 2017-09-04 at 3 53 21 pm

Ping @bigcommerce/stencil-team @bigcommerce/payments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Fixes image overlapping details on product page and Quick View on small viewports [#1067](https://github.com/bigcommerce/cornerstone/pull/1067)
 - Allow 'none' to be a default selection on product option pick lists [#1068](https://github.com/bigcommerce/cornerstone/pull/1068)
 - Fixes a bug where product options and add to cart do not work when opened in Quick View modals [#1070](https://github.com/bigcommerce/cornerstone/pull/1070)
+- Fixes a bug where the Apple Pay button is displayed in incompatible browsers in the Preview Cart modal [#1084](https://github.com/bigcommerce/cornerstone/pull/1084)
 
 ## 1.9.2 (2017-08-16)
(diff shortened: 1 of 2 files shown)


davidchin:
It seems you have to apply `// scss-lint:disable SelectorDepth, NestingDepth` (unless you want to refactor the CSS, but I guess it's a bit difficult for you because you are not able to turn on Apple Pay locally so I'm happy for you to just fix the bug 😄)

bc-anna:
@philipmuir may be able to help you test it locally if you need (and then you can both teach us!)

PascalZajac:
Thanks to Travis' prompting, I decided to take a different approach to this, and instead remove the `display: block` rule from the `previewCartCheckout` ruleset, which means `display` will just be inherited from the rule defined above. I updated the PR description accordingly.

junedkazi:
@PascalZajac can you add a changelog entry before you merge this one.

PascalZajac:
Sure thing, under the Draft heading right?

junedkazi:
Yes it should be in the draft section.

PR #1089, opened 2017-09-27
Update product UPC when options are selected

junedkazi:
#### What?
Selecting an option does not change the UPC to the UPC set on that option's SKU. The UPCs should change based on the option selected & should display the UPC of the option.

#### Tickets / Documentation
https://jira.bigcommerce.com/browse/STENCIL-3639

@bigcommerce/stencil-team

### DIFF
--- a/assets/js/theme/common/product-details.js
+++ b/assets/js/theme/common/product-details.js
@@ -70,6 +70,7 @@ export default class ProductDetails {
                 $input: $('[data-product-stock]', $scope),
             },
             $sku: $('[data-product-sku]'),
+            $upc: $('[data-product-upc]'),
             quantity: {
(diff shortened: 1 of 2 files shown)


assets/js/theme/common/product-details.js
```diff
@@ -70,6 +70,7 @@ export default class ProductDetails {
                 $input: $('[data-product-stock]', $scope),
             },
             $sku: $('[data-product-sku]'),
+            $upc: $('[data-product-upc]'),
```
junedkazi: The data tag already exists https://github.com/bigcommerce/cornerstone/blob/0d95be52e150449f78aa4ea868a0ab5e8cd3185f/templates/components/amp/products/product-view-details.html#L30

PR #1092, opened 2017-10-03
STENCIL-3527: Adding toggle for weight and dimensions on product pages.

cristycarpenter:
#### What?

Adds toggle settings to the theme editor for displaying (or not displaying) weight and dimensions on product pages.

#### Tickets / Documentation

This is a feature request from customers as documented here:

- https://jira.bigcommerce.com/browse/STENCIL-3527

screenshot

Video of working changes: https://www.screencast.com/t/juQ8DUTeG3F8

### DIFF
--- a/config.json
+++ b/config.json
@@ -69,6 +69,8 @@
     "show_accept_paypal": false,
     "show_accept_visa": false,
     "show_product_details_tabs": true,
+    "show_product_weight": true,
+    "show_product_dimensions": false,
     "product_list_display_mode": "grid",
     "logo-position": "center",
     "logo_size": "250x100",
--- a/lang/en.json
+++ b/lang/en.json
@@ -678,6 +678,13 @@
         "including_tax": "Including Tax",
         "excluding_tax": "Excluding Tax",
         "weight": "Weight:",
+        "width": "Width:",
+        "height": "Height:",
+        "depth": "Depth:",
+        "measurement": {
+            "metric": "cm",
+            "imperial": "in"
+        },
         "you_save": "(You save {amount})",
         "gift_wrapping": "Gift wrapping:",
         "gift_wrapping_available": "Options available",
--- a/templates/components/products/product-view.html
+++ b/templates/components/products/product-view.html
@@ -76,10 +76,39 @@
                     <dt class="productView-info-name">{{lang 'products.availability'}}</dt>
                     <dd class="productView-info-value">{{product.availability}}</dd>
                 {{/if}}
-                {{#if product.weight}}
+                {{#all product.weight theme_settings.show_product_weight}}
                     <dt class="productView-info-name">{{lang 'products.weight'}}</dt>
                     <dd class="productView-info-value" data-product-weight>{{product.weight}}</dd>
-                {{/if}}
+                {{/all}}
+                {{#all product.width product.height product.depth theme_settings.show_product_dimensions}}
+                    <dt class="productView-info-name">{{lang 'products.width'}}</dt>
+                    <dd class="productView-info-value" data-product-width>
+                        {{product.width}}
+                        {{#if settings.measurements.length '==' 'Centimeters'}}
+                        ({{lang 'products.measurement.metric'}})
+                        {{else}}
+                        ({{lang 'products.measurement.imperial'}})
+                        {{/if}}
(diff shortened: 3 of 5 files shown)


config.json
```diff
@@ -382,6 +384,8 @@
         "show_accept_paypal": false,
         "show_accept_visa": false,
         "show_product_details_tabs": true,
+        "show_product_weight": true,
+        "show_product_dimensions": true,
```
mcampa: Not needed. This only needs to be added once to the global settings on line 72 and 73

I noticed that there are several cases like this :(. It only needs to be added to the variations if the values are different.

config.json
```diff
@@ -654,6 +658,8 @@
...
```
mcampa: same as above

lang/en.json
```diff
@@ -678,6 +678,13 @@
         "including_tax": "Including Tax",
         "excluding_tax": "Excluding Tax",
         "weight": "Weight:",
+        "width": "Width:",
+        "height": "Height:",
+        "depth": "Depth:",
+        "measurement": {
+        "metric": "cm",
+        "imperial": "in"
```
mcampa: add indentation

templates/components/products/product-view.html
```diff
@@ -76,10 +76,39 @@ <h2 class="productView-brand"{{#if schema}} itemprop="brand" itemscope itemtype=
                     <dt class="productView-info-name">{{lang 'products.availability'}}</dt>
                     <dd class="productView-info-value">{{product.availability}}</dd>
                 {{/if}}
-                {{#if product.weight}}
+                {{#if theme_settings.show_product_weight}}
```
mcampa: you still need to check for `product.weight`. You can use `all` helper

```
{{#all product.weight theme_settings.show_product_weight}}
```

config.json
cristycarpenter: Removed and GTK!

config.json
cristycarpenter: Removed :)

lang/en.json
cristycarpenter: Indented now.

templates/components/products/product-view.html
cristycarpenter: Added this to weight as well as dimensions since dimensions are optional. Thanks for catching these!

templates/components/products/product-view.html
```diff
@@ -76,10 +76,39 @@ <h2 class="productView-brand"{{#if schema}} itemprop="brand" itemscope itemtype=
                     <dt class="productView-info-name">{{lang 'products.availability'}}</dt>
                     <dd class="productView-info-value">{{product.availability}}</dd>
                 {{/if}}
-                {{#if product.weight}}
+                {{#all product.weight theme_settings.show_product_weight}}
                     <dt class="productView-info-name">{{lang 'products.weight'}}</dt>
                     <dd class="productView-info-value" data-product-weight>{{product.weight}}</dd>
-                {{/if}}
+                {{/all}}
+                {{#all product.width product.height. product.depth theme_settings.show_product_dimensions}}
```
mcampa: there is an extra dit in `product.height.` 😛

config.json
```diff
@@ -69,6 +69,8 @@
     "show_accept_paypal": false,
     "show_accept_visa": false,
     "show_product_details_tabs": true,
+    "show_product_weight": true,
+    "show_product_dimensions": true,
```
junedkazi: If we never displayed this info before we should default it to `false`. We don't want people calling support that new data started showing up on their product details page.

aleachjr:
What is the performance impact here? We have heard from merchants it is very slow using theme editor with a bunch of theme options to change. This seems it would slow down theme editor even more. Do we know either way?

templates/components/products/product-view.html
cristycarpenter: :woman_facepalming: ✔️

config.json
cristycarpenter: Good point. I had thought about that but forgot to implement. Good news is I am better at squashing ;)

mcampa:
This is just adding two more fields, so the impact should be marginal. 
I agree with you that TE has performance issues, and we have to invest on fix it.

PR #1111, opened 2017-10-27
STENCIL-3974 - PLP Price Ranges

mjschock:
#### What?

* shows price ranges instead of prices when they are present in the context on product list pages (PLP)

#### Tickets / Documentation

- [STENCIL-3974](https://jira.bigcommerce.com/browse/STENCIL-3974)

@bigcommerce/stencil-team - this is ready to go

### DIFF
--- a/templates/components/products/card.html
+++ b/templates/components/products/card.html
@@ -81,7 +81,11 @@
 
         <div class="card-text" data-test-info-type="price">
             {{#or customer (if theme_settings.restrict_to_login '!==' true)}}
-                {{> components/products/price price=price customer=customer}}
+                {{#if price_range}}
+                    {{> components/products/price-range}}
+                {{else}}
+                    {{> components/products/price price=price customer=customer}}
+                {{/if}}
             {{else}}
                 {{> components/common/login-for-pricing}}
             {{/or}}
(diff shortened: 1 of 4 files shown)


shanth100:
💚

templates/components/products/price-range.html
```diff
@@ -0,0 +1,3 @@
+{{#with price_range}}
+    {{min.formatted}} - {{max.formatted}}
+{{/with}}
```
junedkazi: EOL is missing

mjschock: 👍

templates/components/products/card.html
```diff
@@ -81,7 +81,11 @@ <h4 class="card-title">
 
         <div class="card-text" data-test-info-type="price">
             {{#or customer (if theme_settings.restrict_to_login '!==' true)}}
-                {{> components/products/price price=price customer=customer}}
+                {{#if price_range}}
```
junedkazi: what about product view ? https://github.com/bigcommerce/cornerstone/blob/master/templates/components/products/product-view.html#L19 are we not planning on showing the range in the quick view & product detail page until an option is selected ?

shanth100: From what I understand, this is only for PLP and nowhere else.

PR #1125, opened 2017-11-28
STRF-3496: Create year option in date picker if earliest year equals …

bdav87:
#### What?
When the date picker product option has a date restriction within one year, the year options do no populate on the product page, preventing a shopper from adding the item to the cart if the option is required.

This fix creates an option for the year if the earliest_year and latest_year are the same.

#### Tickets / Documentation
https://jira.bigcommerce.com/browse/STRF-3496

Before:
screen shot 2017-11-28 at 1 42 27 pm

After:
screen shot 2017-11-28 at 1 43 24 pm


bdav87:
Updated to auto-select the year if only 1 year is available in the date picker.

junedkazi:
@bdav87 can you add an entry into the changelog file. See https://github.com/bigcommerce/cornerstone/pull/1131/files as an example.

PR #1130, opened 2017-12-04
STRF-4158 - Fix Bold Featured Products Clickability

mjschock:
#### What?

* `config.json` overode `homepage_featured_products_column_count` for Bold themes, causing the clickabilty to break.

#### Tickets / Documentation

- [STRF-4158](https://jira.bigcommerce.com/browse/STRF-4158)

ping @bigcommerce/stencil-team

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Draft
 - Fixes functionality of carousel links in IE and Edge. [#1093](https://github.com/bigcommerce/cornerstone/pull/1093)
 - Add image width & height for carousel images. [#1126](https://github.com/bigcommerce/cornerstone/pull/1126)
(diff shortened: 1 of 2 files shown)


config.json
```diff
@@ -357,7 +357,6 @@
       },
       "settings": {
         "homepage_featured_products_count": 6,
-        "homepage_featured_products_column_count": 3,
```
junedkazi: I am not sure I understand how does this break click ability. As a developer I should be able to override the setting in each variation over the default. It should not break the theme.

mjschock: @junedkazi you're right, the underlying issue was introduced with #948.  i've reverted that change and this one fixes both issues.

mjschock:
@junedkazi ♻️  - i'm just fixing the bug here now.  i created a separate ticket to address why those properties exist only for Bold themes.

PR #1137, opened 2017-12-20
STRF-4220 - Fix image-overlap on Orders page

mjschock:
#### What?

* Fixes issue where the order image would overlap order information (for Firefox)

#### Tickets / Documentation

- [STRF-4220](https://jira.bigcommerce.com/browse/STRF-4220)

### before
before

### after
after

ping @bigcommerce/storefront-team

cc @BC-EChristensen - wondering if you have any concerns with this change, as it relates to lazy loading of the images?

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -21,7 +21,9 @@
         "global-require": 0,
         "newline-per-chained-call": 0,
         "arrow-parens": 0,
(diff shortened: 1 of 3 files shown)


mjschock:
@junedkazi ♻️ - added changelog

PR #1143, opened 2018-01-03
CATALOG-1447 - Update product cards to use price ranges

bookernath:
#### What?

-Remove "as low as" feature (no longer needed with price ranges)
-Add new "price range" boolean theme editor setting, defaulting to true, to display ranges
-Augment price component to support ranges if available and enabled (graceful fallback for stores not enabled)
-Update schema.org to use PriceSpecification so that we can correctly support ranges and also inc tax/exc tax

#### Tickets / Documentation

- [CATALOG-1447](https://jira.bigcommerce.com/browse/CATALOG-1447)
- [CATALOG-1450](https://jira.bigcommerce.com/browse/CATALOG-1450)

image

#### Other important notes
- This effectively removes support for display of Retail Price when there's a price range. This is intentional, because it might not make too much sense to display a singular Retail Price in conjunction with a price range. As retail prices get range support, we can add it back in.
- Similarly, I haven't yet done anything for the behavior of Sale Price

### DIFF
--- a/templates/components/products/price-range.html
+++ b/templates/components/products/price-range.html
@@ -1,3 +1,37 @@
-{{#with price_range}}
-    {{min.formatted}} - {{max.formatted}}
-{{/with}}
+{{#and price_range.min.with_tax price_range.max.with_tax}}
+    <div class="price-section price-section--withTax" {{#if schema_org}}itemprop="offers" itemscope itemtype="http://schema.org/Offer"{{/if}}>
+        <span data-product-price-with-tax class="price">{{price_range.min.with_tax.formatted}} - {{price_range.max.with_tax.formatted}}</span>
+        {{#and price_range.min.without_tax price_range.max.without_tax}}
+            <abbr title="{{lang 'products.excluding_tax'}}">{{lang 'products.price_with_tax' tax_label=price_range.min.tax_label}}</abbr>
+        {{else if schema_org}}
+                <meta itemprop="availability" content="{{product.availability}}">
+                <meta itemprop="itemCondition" itemtype="http://schema.org/OfferItemCondition" content="http://schema.org/{{product.condition}}Condition">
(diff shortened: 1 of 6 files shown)


templates/components/products/price-range.html
```diff
@@ -1,3 +1,35 @@
 {{#with price_range}}
-    {{min.formatted}} - {{max.formatted}}
-{{/with}}
+    {{#and price_range.min.with_tax price_range.max.with_tax}}
```
mjschock: since we're using a `{{#with price_range}}` above we can do `{{#and min.with_tax max.with_tax}}` instead and `{{#and}}` seems fine to me

bookernath: Thanks, will update

bookernath:
Screenshot from integration:

image

templates/components/products/price-range.html
```diff
@@ -1,3 +1,37 @@
-{{#with price_range}}
-    {{min.formatted}} - {{max.formatted}}
-{{/with}}
+{{#and price_range.min.with_tax price_range.max.with_tax}}
+    <div class="price-section price-section--withTax" {{#if schema_org}}itemprop="offers" itemscope itemtype="http://schema.org/Offer"{{/if}}>
+        <span data-product-price-with-tax class="price">{{price_range.min.with_tax.formatted}} - {{price_range.max.with_tax.formatted}}</span>
+        {{#and price_range.min.without_tax price_range.max.without_tax}}
+            <abbr title="{{lang 'products.excluding_tax'}}">{{lang 'products.price_with_tax' tax_label=price_range.min.tax_label}}</abbr>
+        {{else if schema_org}}
+                <meta itemprop="availability" content="{{product.availability}}">
```
junedkazi: I am wondering if we can move the meta into a price schema component.

PR #1159, opened 2018-01-31
STRF-4373: Removing lazyload for logo in order-confirmation.html

jfasanobc:
#### What?

The image for the order-confirmation.html was not loading, and as with the issue in [STRF-4138](https://jira.bigcommerce.com/browse/STRF-4138), I've removed the lazyload for the logo for this template page.

#### Tickets / Documentation

- [STRF-4138](https://jira.bigcommerce.com/browse/STRF-4138)
- [STRF-4373](https://jira.bigcommerce.com/browse/STRF-4373)

[Before](https://www.screencast.com/t/VRojxYj1y8)

[After](https://www.screencast.com/t/cyl6e8sv1)

### DIFF
--- a/templates/pages/order-confirmation.html
+++ b/templates/pages/order-confirmation.html
@@ -16,7 +16,7 @@
         <h2 class="checkoutHeader-heading">
             <a class="checkoutHeader-link" href="{{urls.home}}">
                 {{#if checkout.header_image}}
-                    <img alt="{{settings.store_logo.title}}" class="checkoutHeader-logo lazyload" id="logoImage" data-sizes="auto" src="{{cdn 'img/loading.svg'}}" data-src="{{ checkout.header_image }}"/>
+                    <img alt="{{settings.store_logo.title}}" class="checkoutHeader-logo" id="logoImage" src="{{ checkout.header_image }}"/>
                 {{ else }}
                     <span class="header-logo-text">{{settings.store_logo.title}}</span>
                 {{/if}}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Draft
+- Fix logo not loading on order confirmation page [#1159](https://github.com/bigcommerce/cornerstone/pull/1159)
 
 ## 1.12.1 (2018-01-23)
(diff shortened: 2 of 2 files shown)


junedkazi:
@mjschock just making sure you saw this PR bcoz there is a ticket assigned to you for this sprint.

templates/pages/order-confirmation.html
```diff
@@ -16,7 +16,7 @@ <h1 class="is-srOnly">{{lang 'checkout.title'}}</h1>
         <h2 class="checkoutHeader-heading">
             <a class="checkoutHeader-link" href="{{urls.home}}">
                 {{#if checkout.header_image}}
-                    <img alt="{{settings.store_logo.title}}" class="checkoutHeader-logo lazyload" id="logoImage" data-sizes="auto" src="{{cdn 'img/loading.svg'}}" data-src="{{ checkout.header_image }}"/>
+                    <img alt="{{settings.store_logo.title}}" class="checkoutHeader-logo" id="logoImage" src="{{ checkout.header_image }}"/>
```
junedkazi: I am fine with this change but instead of removing the lazy load we should be investigating what is causing the failure to load in the first place.

jfasanobc: True, I was just looking at a quick solution for now, as it's only this and another page from a previous PR. I tried researching it a bit, but it was bit over my head as to why. Thanks for reviewing, I wasn't sure as to why it wouldn't let me add reviewers.

BC-EChristensen: Do we support running theme scripts on the order confirmation page? I believe the lazysizes script just isn't being inserted there, despite being part of global.js. Assuming that's the case, this is quite possibly the correct solution, short of manually inserting a `<script>` tag in the order confirmation template.

junedkazi: I am guessing that is what is happening as well.

mjschock:
i'm going to go ahead a merge this - we'll create a separate ticket to pull out the shared code between the checkout and order-confirmation pages.  i was thinking of something like a common checkout partial where we pass the down `{{{ checkout.checkout_content }}}` or `{{{ checkout.order_confirmation_content }}}` respectively as `{{> @partial-block}}`.  @jfasanobc - if you want to create another PR to do that you're more than welcome but this is fine for now.

PR #1167, opened 2018-02-21
CATALOG-2408 - Bulk Pricing updates per-variant

bookernath:
#### What?

Price Lists support different bulk pricing tiers for each variant, so we need to update the viewModel whenever these tiers change.

#### Tickets / Documentation

- [CATALOG-2408](https://jira.bigcommerce.com/browse/CATALOG-2408)

GIF

### DIFF
--- a/assets/js/theme/common/product-details.js
+++ b/assets/js/theme/common/product-details.js
@@ -31,16 +31,16 @@ export default class ProductDetails {
             this.addProductToCart(event, $form[0]);
         });
 
-        // Update product attributes. If we're in quick view and the product has options,
-        // then also update the initial view in case items are oos
-        if (_.isEmpty(productAttributesData) && hasOptions) {
+        // Update product attributes. Also update the initial view in case items are oos
+        // or have default variant properties that change the view
+        if (hasOptions) {
             const $productId = $('[name="product_id"]', $form).val();
 
-            utils.api.productAttributes.optionChange($productId, $form.serialize(), (err, response) => {
+            utils.api.productAttributes.optionChange($productId, $form.serialize(), 'products/bulk-discount-rates', (err, response) => {
                 const attributesData = response.data || {};
-
+                const attributesContent = response.content || {};
                 this.updateProductAttributes(attributesData);
-                this.updateView(attributesData);
+                this.updateView(attributesData, attributesContent);
             });
         } else {
             this.updateProductAttributes(productAttributesData);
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -6,7 +6,7 @@
   "author": "BigCommerce",
   "license": "MIT",
   "dependencies": {
-    "@bigcommerce/stencil-utils": "^1.0.10",
+    "@bigcommerce/stencil-utils": "bookernath/stencil-utils#d263f2ef03aa88e1361c8b4331204e4e4e8bd5e4",
```
bookernath: this gets replaced with a new version once the utils PR is merged.

assets/js/theme/cart.js
```diff
@@ -98,7 +98,7 @@ export default class Cart extends PageManager {
             const $messageBox = $('.alertMessageBox');
             const item = $('[name="item_id"]', $form).attr('value');
 
-            utils.api.productAttributes.optionChange(item, $form.serialize(), (err, result) => {
```
mjschock: looking at the new `optionChange` api, it seems that we don't need to pass `null` here

assets/js/theme/common/product-details.js
```diff
@@ -107,11 +108,11 @@ export default class ProductDetails {
             return;
         }
 
-        utils.api.productAttributes.optionChange(productId, $form.serialize(), (err, response) => {
+        utils.api.productAttributes.optionChange(productId, $form.serialize(), 'products/bulk-discount-rates', (err, response) => {
             const productAttributesData = response.data || {};
-
+            const bulkPricingHtml = response.content || {};
```
mjschock: could we match the productAttributesData here, like so:

```
const content = response.content || {};
```
...
```
updateView(data, content)
```
etc.

assets/js/theme/common/product-details.js
```diff
@@ -393,6 +393,13 @@ export default class ProductDetails {
             viewModel.$addToCart.prop('disabled', false);
             viewModel.$increments.prop('disabled', false);
         }
+
+        // If Bulk Pricing rendered HTML is available
+        if (content) {
```
mjschock: does this work?  i thought it was grabbing a field off of content... like this instead:

```
updateView(data, content) { // content will be {} if response.content is empty above
```
...
```
if (content.bulkPricingContent) { // or whatever it's called
  viewModel.$bulkPricing.html(content.bulkPricingContent)
```

bookernath: Content returns a blank string which evaluates to no if there are no tiers.

bookernath:
@mjschock I think I've addressed most of your feedback if you wanna take a peek. Also rebased to pull in the JQ update.

PR #1178, opened 2018-03-07
MERC-3389: Move Google Font CSS loading to end of body.

pedelman:
#### What?

Currently we are loading the Google WebFont CSS in the HTML head of the page. Since loading CSS is a blocking operation, the page will be blocked from rendering the document while we are fetching the script.

Also since this blocks, if the web font css goes down, the entire page will be blocked from loading. In the following screenshot, you can see 99% of the page will be failed to render if the source goes down. This is referred to as a "Single Point of Failure" (SPOF).

screen shot 2018-03-06 at 11 02 29 pm

#### Tickets / Documentation

- Tool Used: [Spof-O-Matic](https://chrome.google.com/webstore/detail/spof-o-matic/plikhggfbplemddobondkeogomgoodeg?hl=en-US)
- [Frontend SPOF](https://www.stevesouders.com/blog/2010/06/01/frontend-spof/)

### DIFF
--- a/templates/layout/amp-iframe.html
+++ b/templates/layout/amp-iframe.html
@@ -8,7 +8,6 @@
         <link href="{{ head.favicon }}" rel="shortcut icon">
         <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1">
         {{{stylesheet '/assets/css/theme.css'}}}
-        {{ getFontsCollection }}
         <script src="{{cdn 'assets/modernizr-custom.js'}}"></script>
 
         {{{ head.scripts }}}
@@ -46,5 +45,6 @@
             }, false);
         </script>
 
+        {{ getFontsCollection }}
     </body>
 </html>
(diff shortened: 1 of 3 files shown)


mattolson:
We have to decide if we are ok with a flash of unstyled text. After this is merged, the page may need to re-render once the fonts are loaded. Some browsers (like Chrome) handle this gracefully, others (like Firefox and IE) do not.

mattolson:
https://developers.google.com/fonts/docs/technical_considerations

bookernath:
Hey @pedelman, I'd prefer to solve this same issue via use of `font-display`

Google Fonts needs to be updated to support it, but apparently it's planned.

Sources:
https://developers.google.com/web/updates/2016/02/font-display
https://css-tricks.com/font-display-masses/
https://github.com/google/fonts/issues/358
https://medium.com/@pierluc/enable-font-display-on-google-fonts-with-cloudflare-workers-d7604cb30eab

PR #1189, opened 2018-03-11
Webpack 4 + Javascript optimizations

mattolson:
## What?

* Upgrade to webpack 4
* Clean up stencil bootstrap
* Simplify interface for PageManager and get rid of async library dependency
* Get rid of html5-history-api library dependency (no longer needed)
* Get rid of fastclick library dependency (no longer needed)
* Upgrade several babel dependencies
* Separate webpack config into common, dev, prod, and add npm scripts to build
* Get rid of minifications in babel loader, instead rely on webpack
* Get rid of LoaderOptionsPlugin
* Get rid of CommonsChunkPlugin (webpack 4 will automaticaly do this for prod)

## Performance Impact
* Initial load of javascript is down from 450kb to 350kb. This PR lowers time to first paint by 50ms and time to first interactive by 210ms.

#### Before
screen shot 2018-03-11 at 2 47 33 pm

#### After
screen shot 2018-03-11 at 2 47 40 pm

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,13 @@
 language: node_js
 node_js:
-- 4
+  - 6
 sudo: required
 dist: trusty
 install:
-- bundle install
-- npm install -g grunt-cli
-- npm install -g bigcommerce/stencil-cli
-- npm install
+  - bundle install
+  - npm install -g grunt-cli
+  - npm install -g bigcommerce/stencil-cli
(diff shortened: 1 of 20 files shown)


mjschock:
@mattolson we should update the readme to remove references to the before/loaded/after handlers before we merge this to master.  might need to update the dev docs too

PR #1192, opened 2018-03-13
STRF-4527: Fix amp image sizes

mattolson:
## What?

* Instead of hardcoding 300x300, use the actual image dimensions
  for product images on the product list

#### Before
screen shot 2018-03-12 at 5 29 09 pm

#### After
screen shot 2018-03-12 at 5 28 14 pm

### DIFF
--- a/templates/components/amp/products/card.html
+++ b/templates/components/amp/products/card.html
@@ -8,7 +8,7 @@
             {{/if}}
         {{/or}}
         <a data-vars-product-link="{{url}}" data-vars-product-id="{{id}}" data-vars-product-name="{{name}}" href="{{url}}">
-            <amp-img class="card-image" src="{{getImage image 'productgallery_size' (cdn theme_settings.default_image_product)}}" width="300" height="300" layout="responsive" alt="{{image.alt}}"></amp-img>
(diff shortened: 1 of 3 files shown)


mattolson:
@bigcommerce/storefront-team

bc-jcruiz:
interesting, I wonder why was that hardcoded? Oh well, glad it was a quick fix

templates/components/amp/products/card.html
```diff
@@ -8,7 +8,7 @@
             {{/if}}
         {{/or}}
         <a data-vars-product-link="{{url}}" data-vars-product-id="{{id}}" data-vars-product-name="{{name}}" href="{{url}}">
-            <amp-img class="card-image" src="{{getImage image 'productgallery_size' (cdn theme_settings.default_image_product)}}" width="300" height="300" layout="responsive" alt="{{image.alt}}"></amp-img>
+          <amp-img class="card-image" src="{{getImage image 'productgallery_size' (cdn theme_settings.default_image_product)}}" width="{{first (split theme_settings.productgallery_size 'x')}}" height="{{last (split theme_settings.productgallery_size 'x')}}" layout="responsive" alt="{{image.alt}}"></amp-img>
```
tpietsch: not familiar with this template language. 
width="{{first (split theme_settings.productgallery_size 'x')}}" height="{{last (split theme_settings.productgallery_size 'x')}}"

not sure what these are doing but is it guaranteed to have a 'first'/'last'? otherwise would you end up with a 0 default height and image wouldn't display?

mattolson: Yeah, we are. This code is pulling values out of the theme configuration, and they are formatted as `${width}x${height}`. For example: https://github.com/bigcommerce/cornerstone/blob/master/config.json#L80

PR #1195, opened 2018-03-26
Catalog 2406

bc-vandana:
#### What?

Adding stencil logic for the rrp strikethrough on PDP and PLP

#### Tickets / Documentation

https://jira.bigcommerce.com/browse/CATALOG-2406 has the mockups for all the scenarios for products and variants. 

https://jira.bigcommerce.com/browse/CATALOG-2406

### Tests
STORE_DOMAIN="https://store-hkyr4dowa1.my-integration.zone" STORE_EMAIL="vandana.premkumar+1522082908@bigcommerce.com" STORE_PASSWORD="db421f1fddd534d3800a" ./functional_phpunit.sh --filter PriceListsStoreFrontPLPUITest --group V3

Starting Functional Tests against specified store:
Cleaning up screenshots...

Domain: https://store-hkyr4dowa1.my-integration.zone
Email: vandana.premkumar+1522082908@bigcommerce.com
Password: db421f1fddd534d3800a

PHP Warning:  Module 'pcntl' already loaded in Unknown on line 0

Warning: Module 'pcntl' already loaded in Unknown on line 0
STORE Hash URL: https://store-hkyr4dowa1.my-integration.zone

Build commit id: fea802701841281c261c4c1edcc7547454d0b455
PHPUnit 6.0.8 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.0.27 with Xdebug 2.6.0
Configuration: /Users/vandana.premkumar/app/bigcommerce-app-vm/codebases/bigcommerce/tests/Functional/functional.xml

....                                                                4 / 4 (100%)

Time: 2.69 minutes, Memory: 76.00MB

OK (4 tests, 89 assertions)

STORE_DOMAIN="https://store-hkyr4dowa1.my-integration.zone" STORE_EMAIL="vandana.premkumar+1522082908@bigcommerce.com" STORE_PASSWORD="db421f1fddd534d3800a" ./functional_phpunit.sh --filter PriceListsStoreFrontPDPUITest  --group V3
selenium-server-standalone-3.5.0.jar already exists, skipping download
Gecko driver exists, skipping download
Starting Selenium Server with options '-Dwebdriver.gecko.driver=./geckodriver_v0.17.0 -jar ./selenium-server-standalone-3.5.0.jar -enablePassThrough false' in the background
Selenium Server started with PID 87827
To reuse this store for future requests, run:
STORE_DOMAIN="https://store-hkyr4dowa1.my-integration.zone" STORE_EMAIL="vandana.premkumar+1522082908@bigcommerce.com" STORE_PASSWORD="db421f1fddd534d3800a" ./functional_phpunit.sh --filter PriceListsStoreFrontPDPUITest --group V3

Starting Functional Tests against specified store:
Cleaning up screenshots...

Domain: https://store-hkyr4dowa1.my-integration.zone
Email: vandana.premkumar+1522082908@bigcommerce.com
Password: db421f1fddd534d3800a

PHP Warning:  Module 'pcntl' already loaded in Unknown on line 0

Warning: Module 'pcntl' already loaded in Unknown on line 0
STORE Hash URL: https://store-hkyr4dowa1.my-integration.zone

Build commit id: fea802701841281c261c4c1edcc7547454d0b455
PHPUnit 6.0.8 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.0.27 with Xdebug 2.6.0
Configuration: /Users/vandana.premkumar/app/bigcommerce-app-vm/codebases/bigcommerce/tests/Functional/functional.xml

..........	.....                                                   15 / 15 (100%)

Time: 14.81 minutes, Memory: 78.00MB

OK (15 tests, 268 assertions)
Functional Tests complete

@bookernath @junedkazi @mjschock


lang/en.json
```diff
@@ -674,7 +674,9 @@
         "choose_an_option": "Please choose an option",
         "select_one": "Please select one",
         "description": "Description",
-        "retail_price": "MSRP",
+        "retail_price": "MSRP:",
+        "was": "Was:",
```
junedkazi: Can we prefix value with `price` like `price_was`. Same for `now`.

lang/en.json
```diff
@@ -687,7 +689,8 @@
             "metric": "cm",
             "imperial": "in"
         },
-        "you_save": "(You save {amount})",
+        "you_save_1": "(You save",
+        "you_save_2": ")",
```
junedkazi: You should change it to a single variable `you_save` like `(You save {price})`. And then you can call it as `{{lang 'products.you_save' price=price.saved.formatted}}`.

templates/components/products/price.html
```diff
@@ -44,10 +68,12 @@
...
         {{#if price.saved}}
-            <div class="price-section price-section--saving">
-                <span class="price">
-                    {{lang 'products.you_save' amount=price.saved.formatted}}
+            <div class="price-section price-section--saving price">
+                <span class="price">{{lang 'products.you_save_1'}}</span>
+                <span data-product-price-saved class="price">
+                    {{price.saved.formatted}}
                 </span>
+                <span class="price">{{lang 'products.you_save_2'}}</span>
```
junedkazi: Please update this based on the suggestions above

lang/en.json
bc-vandana: That unfortunately does not work, and is causing the bug CATALOG-3121. Which is the main reason for the split

templates/components/products/price.html
bc-vandana: That unfortunately does not work, and is causing the bug CATALOG-3121. Which is the main reason for the split

lang/en.json
junedkazi: Can you make that change & show it to me or Michael. I am curious why it does not work.

bc-vandana: You can see the first "you save" has the accurate value, but the second "you save" one that uses `{{lang 'products.you_save' price=price.saved.formatted}}` does not update the value.
screen shot 2018-03-26 at 4 43 47 pm

lang/en.json
```diff
@@ -674,7 +674,9 @@
         "choose_an_option": "Please choose an option",
         "select_one": "Please select one",
         "description": "Description",
-        "retail_price": "MSRP",
+        "retail_price": "MSRP:",
+        "price_was": "Was:",
+        "price_now": "Now:",
```
bookernath: 🍹 in my opinion we would leave the `:` out of the lang file and in the template directly, @junedkazi do you have an opinion?

mjschock: i think you could argue the colon is part of the language.  we also have it in MSRP: above so i think think it's fine to stay

templates/components/products/price.html
mjschock: yeah, this seems fine because we are dealing with updating the already-rendered view here and so can't pass it to the helper at this point.  but let's name these something more descriptive

mjschock: i'm wondering - maybe we could simply have something like a you_save = "You save" and then include the parens as part of the html or css so we could have:

```
<span class="price">{{lang 'products.you_save'}}
<span data-product-price-saved class="price">
  {{price.saved.formatted}}
</span>
```

::before {
  content: "(";
}

::after {
  content: ")";
}

thetechie92: Easiest would be in html just in case css messes up. You would always be 100% sure it works with html

bc-vandana: splitting it in en.json seems the best for now. But we will need to fix the underlining problem of having a dynamic value to be added as a parameter.

lang/en.json
bc-vandana: thank you

lang/en.json
bc-vandana: done

bc-vandana:
I added more code yesterday to fix some UI fixes where ajax response values are not showing up in UI when variant is modified. JZ is testing in an integration store that has the updated theme.
http://vandanapremkumar1522343934-testington.my-integration.zone/

bc-jz:
This PR is looking good. The only lingering issue I am seeing is that if retail or sale price are not showing at the time of page load then they will not show when choosing variants that have retail and/or sale price values. This is a problem with our current template as well though and I have created a JIRA to document:

https://jira.bigcommerce.com/browse/STRF-4701

💚

PR #1197, opened 2018-03-29
fix(catalog): CATALOG-3085 default selection in quickview

Souldjinn:
#### What?
   refrain from updating the price shown on
    quickview by default when the modal is rendered unless theres an
    actual default selection to be made.

#### Tickets / Documentation
- [Ticket](https://jira.bigcommerce.com/browse/CATALOG-3085)

Before: The quickview page would instantly make a request for stock/price based on selection even if the product doesnt have a default selection, which would result in getting the base price.
screen shot 2018-03-29 at 4 38 01 pm

After: The quickview page still makes the default request so that it can mark things as OOS when necessary, but it doesnt try to change the price in the view if its not applicable. 
screen shot 2018-03-29 at 4 36 42 pm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Draft
 - Add representation for products and variants with retail price that has sale price. [#1195](https://github.com/bigcommerce/cornerstone/pull/1195) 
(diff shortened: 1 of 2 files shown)


bc-jz:
💚 - this update looks to be working as expected

PR #1205, opened 2018-04-20
STRF-4679 - fixing product layout when sidebar off

tpietsch:
#### What?

Added the theme setting check for the shop by price sidebar

#### Tickets / Documentation

https://jira.bigcommerce.com/browse/STRF-4679

screen shot 2018-04-20 at 9 34 48 am
screen shot 2018-04-20 at 9 34 55 am

@bigcommerce/platform-engineering @bigcommerce/storefront-team

### DIFF
--- a/templates/components/category/sidebar.html
+++ b/templates/components/category/sidebar.html
@@ -1,20 +1,22 @@
-<nav>
-    {{#if category.subcategories}}
-    <div class="sidebarBlock">
-        <h5 class="sidebarBlock-heading">{{category.name}}</h5>
-        <ul class="navList">
-            {{#each category.subcategories}}
-            <li class="navList-item">
-                <a class="navList-action" href="{{url}}" alt="{{name}}" title="{{name}}">{{name}}</a>
-            </li>
-            {{/each}}
-        </ul>
-    </div>
-    {{/if}}
+<aside class="page-sidebar" id="faceted-search-container">
+    <nav>
+        {{#if category.subcategories}}
+        <div class="sidebarBlock">
+            <h5 class="sidebarBlock-heading">{{category.name}}</h5>
+            <ul class="navList">
+                {{#each category.subcategories}}
+                <li class="navList-item">
+                    <a class="navList-action" href="{{url}}" alt="{{name}}" title="{{name}}">{{name}}</a>
+                </li>
+                {{/each}}
+            </ul>
+        </div>
+        {{/if}}
 
-    {{#if category.faceted_search_enabled}}
-        {{> components/faceted-search/index category}}
-    {{else}}
-        {{> components/category/shop-by-price shop_by_price=category.shop_by_price category_url=category.url}}
-    {{/if}}
-</nav>
+        {{#if category.faceted_search_enabled}}
+            {{> components/faceted-search/index category}}
+        {{else}}
+            {{> components/category/shop-by-price shop_by_price=category.shop_by_price category_url=category.url}}
+        {{/if}}
+    </nav>
+</aside>
--- a/templates/pages/category.html
+++ b/templates/pages/category.html
@@ -24,11 +24,15 @@ category:
 {{{category.description}}}
 {{{snippet 'categories'}}}
 <div class="page">
-    {{#or category.subcategories category.faceted_search_enabled category.shop_by_price}}
-        <aside class="page-sidebar" id="faceted-search-container">
+    {{#if category.subcategories}}
+        {{> components/category/sidebar}}
+    {{else if category.faceted_search_enabled}}
+        {{> components/category/sidebar}}
+    {{else if category.shop_by_price}}
+        {{#if theme_settings.shop_by_price_visibility}}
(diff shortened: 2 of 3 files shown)


templates/pages/category.html
```diff
@@ -24,11 +24,13 @@ <h1 class="page-heading">{{category.name}}</h1>
 {{{category.description}}}
 {{{snippet 'categories'}}}
 <div class="page">
-    {{#or category.subcategories category.faceted_search_enabled category.shop_by_price}}
-        <aside class="page-sidebar" id="faceted-search-container">
-            {{> components/category/sidebar}}
-        </aside>
-    {{/or}}
+    {{#if theme_settings.shop_by_price_visibility}}
+        {{#or category.subcategories category.faceted_search_enabled category.shop_by_price}}
```
tpietsch: pretty sure this is not the correct solution because shop by price could be off while these other features are on correct?

templates/pages/category.html
```diff
@@ -24,11 +24,13 @@ <h1 class="page-heading">{{category.name}}</h1>
 {{{category.description}}}
 {{{snippet 'categories'}}}
 <div class="page">
-    {{#or category.subcategories category.faceted_search_enabled category.shop_by_price}}
-        <aside class="page-sidebar" id="faceted-search-container">
-            {{> components/category/sidebar}}
-        </aside>
-    {{/or}}
+    {{#if theme_settings.shop_by_price_visibility}}
```
junedkazi: This won't work. Currently it is just shop by price but if you have faceted search enabled there are other elements in the side bar https://github.com/bigcommerce/cornerstone/blob/67305e81473f01a93a351ad4b2e80b60fc9c66ba/templates/components/category/sidebar.html. 

This change will hide those blocks as well.

templates/components/category/sidebar.html
```diff
@@ -18,3 +20,4 @@ <h5 class="sidebarBlock-heading">{{category.name}}</h5>
         {{> components/category/shop-by-price shop_by_price=category.shop_by_price category_url=category.url}}
     {{/if}}
 </nav>
+</aside>
```
tpietsch: verified sidebar is only used by category.html

templates/pages/category.html
```diff
@@ -25,9 +25,15 @@ <h1 class="page-heading">{{category.name}}</h1>
...
+        {{#if category.subcategories}}
             {{> components/category/sidebar}}
-        </aside>
+        {{else if category.faceted_search_enabled}}
+            {{> components/category/sidebar}}
+        {{else if category.shop_by_price}}
+            {{#if theme_settings.shop_by_price_visibility}}
+                {{> components/category/sidebar}}
+            {{/if}}
+        {{/if}}
```
tpietsch: @junedkazi fixed to handle the cases separately. if you have suggestions or if you can point me in the direction of whatever templating lang this is so i can read up.

templates/pages/category.html
tpietsch: figured as such - fixed

templates/components/category/sidebar.html
```diff
@@ -1,20 +1,22 @@
...
-        {{> components/category/shop-by-price shop_by_price=category.shop_by_price category_url=category.url}}
-    {{/if}}
-</nav>
+        {{#if category.faceted_search_enabled}}
+            {{> components/faceted-search/index category}}
+        {{else}}
+            {{> components/category/shop-by-price shop_by_price=category.shop_by_price category_url=category.url}}
+        {{/if}}
+    </nav>
+</aside>
```
tpietsch: verified this is only used by the file im changing - also not sure why the diff is showing like this all i did was move the aside tag in here

templates/pages/category.html
junedkazi: This is https://handlebarsjs.com

templates/pages/category.html
```diff
@@ -25,9 +25,15 @@ <h1 class="page-heading">{{category.name}}</h1>
 {{{snippet 'categories'}}}
 <div class="page">
     {{#or category.subcategories category.faceted_search_enabled category.shop_by_price}}
```
mattolson: I think the if/elseif below can be moved into this #or block helper. We import most of https://github.com/helpers/handlebars-helpers#helpers so you can use #or and #and together to achieve the same thing. Try this:
`#or category.subcategories category.faceted_search_enabled (and category.shop_by_price theme_settings.shop_by_price_visibility)`

mattolson: Try this instead:
```
{{#any category.subcategories category.faceted_search_enabled (all category.shop_by_price theme_settings.shop_by_price_visibility)}}
```

tpietsch: both have the "TypeError: Uncaught error: opts.inverse is not a function" error

mattolson: Ok, forget that, I guess you need an else clause for that to work. I would just go back to your current solution, but can you get rid of the redundant `#or` that wraps all this?

PR #1210, opened 2018-04-24
STRF-4803 - adding helper function to remove empty files from jquery …

tpietsch:
https://stackoverflow.com/questions/49672992/ajax-request-fails-when-sending-formdata-including-empty-file-input-in-safari

our version of jquery has a problem making ajax calls with empty files in safari. Appears the solution is to remove empty files before sending unless anyone has other suggestions.

@bigcommerce/storefront-team @bigcommerce/platform-engineering

### DIFF
--- a/assets/js/theme/common/product-details.js
+++ b/assets/js/theme/common/product-details.js
@@ -57,6 +57,25 @@ export default class ProductDetails {
     }
 
     /**
+     * https://stackoverflow.com/questions/49672992/ajax-request-fails-when-sending-formdata-including-empty-file-input-in-safari
+     * Safari browser with jquery 3.3.1 has an issue uploading empty file parameters. This function removes any empty files from the form params
+     * @param formData: FormData object
+     * @returns FormData object
+     */
+    filterEmptyFilesFromForm(formData) {
+        try {
+            for (const [key, val] of formData) {
+                if (val instanceof File && !val.name && !val.size) {
+                    formData.delete(key);
+                }
+            }
+        } catch (e) {
+            console.error(e);
+        }
+        return formData;
+    }
+
+    /**
      * Since $productView can be dynamically inserted using render_with,
      * We have to retrieve the respective elements
      *
@@ -234,7 +253,7 @@ export default class ProductDetails {
         this.$overlay.show();
 
         // Add item to cart
-        utils.api.cart.itemAdd(new FormData(form), (err, response) => {
+        utils.api.cart.itemAdd(this.filterEmptyFilesFromForm(new FormData(form)), (err, response) => {
             const errorMessage = err || response.data.error;
 
             $addToCartBtn


assets/js/theme/common/product-details.js
```diff
@@ -56,6 +56,25 @@ export default class ProductDetails {
...
+     * https://stackoverflow.com/questions/49672992/ajax-request-fails-when-sending-formdata-including-empty-file-input-in-safari
+     * Safari browser has an issue uploading empty file parameters. This function removes any empty files from the form params
+     * @param form: FormData object
+     * @returns FormData object
+     */
+    filterEmptyFilesFromForm(form) {
+        for (const pair of form.entries()) {
+            const key = pair[0];
+            const val = pair[1];
+            if (typeof val === 'object' && val instanceof File) {
```
lord2800: 🍹 Are there any instances where `val instanceof File` passes but `typeof val === 'object'` does not? You can simplify this to just an `instanceof` check if not.

tpietsch: guess not - was trying to be safe for the undef case but realized that "undefined instanceof File" is safe

tpietsch: would it be better actually to leave the object check and just check if the size/name fields are propertyOf -> and then check the values are not empty/undefined?

assets/js/theme/common/product-details.js
```diff
@@ -56,6 +56,29 @@ export default class ProductDetails {
...
 
+    /**
+     * https://stackoverflow.com/questions/49672992/ajax-request-fails-when-sending-formdata-including-empty-file-input-in-safari
+     * Safari browser with jquery 3.3.1 has an issue uploading empty file parameters. This function removes any empty files from the form params
+     * @param form: FormData object
+     * @returns FormData object
+     */
+    filterEmptyFilesFromForm(form) {
+        try {
+            for (const pair of form.entries()) {
```
mjschock: 🍹 could use destructuring (i.e. `const [key, val] of form.entries()`)

tpietsch: do we have browser support requirements?

assets/js/theme/common/product-details.js
```diff
@@ -56,6 +56,27 @@ export default class ProductDetails {
...
 
+    /**
+     * https://stackoverflow.com/questions/49672992/ajax-request-fails-when-sending-formdata-including-empty-file-input-in-safari
+     * Safari browser with jquery 3.3.1 has an issue uploading empty file parameters. This function removes any empty files from the form params
+     * @param form: FormData object
+     * @returns FormData object
+     */
+    filterEmptyFilesFromForm(form) {
+        try {
+            for (const [key, val] of form.entries()) {
```
mjschock: looks like you should be able to use `const [key, val] of form`

`
An object implementing FormData can directly be used in a for...of structure, instead of entries(): for (var p of myFormData) is equivalent to for (var p of myFormData.entries()).
` (https://developer.mozilla.org/en-US/docs/Web/API/FormData).  maybe rename the param to formData to be explicit?

assets/js/theme/common/product-details.js
```diff
@@ -56,6 +56,27 @@ export default class ProductDetails {
...
+                if (val instanceof File) {
```
mjschock: might as well pull the second conditional into the first:
`if (val instanceof File && val.name === '' && val.size === 0) {`

could also simplify to `if (val instanceof File and !val.name && !val.size) {`

assets/js/theme/common/product-details.js
mjschock: we do but don't need to worry about them in this case.  we've got webpack and babel to transpile down to the envs we specify in the browserlist file.  :)

assets/js/theme/common/product-details.js
tpietsch: i dont think safaris supports for of on formdata

Issue #1136, opened 2017-12-14
Linking to IDs in tabbed sections do not open those tabs


jkelly-cas:
### Expected behavior
Visiting http://cornerstone-light-demo.mybigcommerce.com/all/orbit-terrarium-large/#tab-similar should open the #tab-similar section automatically. 

### Actual behavior
Site scrolls to the reviews section, but does not open the tab.

### Steps to reproduce behavior
Visit http://cornerstone-light-demo.mybigcommerce.com/all/orbit-terrarium-large/#tab-similar
page will not open #tab-similar section

See http://arcade-natural-demo.mybigcommerce.com/trekking-axe-classic/#product-tab-information for a working demo.

See http://partswarehouse-classic-demo.mybigcommerce.com/heater-bypass-hose-tube/#tab-similarProducts for another non-working demo.

### Notes
This would be really handy for bookmarking purposes and linking from other pages to specific sections in tabs.

funvill:
Any update on this issue ? I am also experiencing it

funvill:
Re-posted to bigcommerce forms 
https://forum.bigcommerce.com/s/question/0D51B00004IKmeBSAT/

junedkazi:
@funvill @jkelly-cas can you give us any additional information like OS, browser on which it is happening. I am currently on Chrome 65.0.3325.181 on a Mac OS High Sierra. See the video below https://www.useloom.com/share/ce09629a6965449e9a9c43a99a30f59d

flyingL123:
@junedkazi you're video shows the bug. The link goes to `#tab-similar`, but it isn't open. The open tab is the default `#tab-related`.

funvill:
Chrome 65 
Windows 10
http://whatsmybrowser.org/b/K64CMXW

Two tabs at the bottom of the screen. 

- Customers Also Viewed (#tab-similar) 
- Related Products (#tab-related) 

When I got to the page with the following url https://cornerstone-light-demo.mybigcommerce.com/all/orbit-terrarium-large/#tab-similar I was expecting to see the "Customers Also Viewed" tab instead of the "Related Products" tab. 

Same problem exists in your video.

funvill:
![click](https://user-images.githubusercontent.com/314120/38333831-8eccf998-380e-11e8-8da5-9d95c2fc7f8a.gif)

junedkazi:
@funvill @flyingL123 thanks for correcting me. I see the issue now.

[...]

PR #1215, opened 2018-05-01
Fix issue with duplicate IDs on product options, Fix issue #1136 (open correct product tabs)

flair-duncan:
#### What?

In certain circumstances, product option elements end up with the same ID causing a clash, stopping some options being selected.

Each option component has been given a descriptive ID to prevent the clash. As the add to cart action relies on the name attribute, this does not interfere with the cart system.

#### Screenshots

[Video of issue before changes](https://www.useloom.com/share/1e740a49826f44cdbd56f81bd01bd7d3)

[Video after changes](https://www.useloom.com/share/9fe6d969da9445da91532da6f301de23)

#### What?

When the product page URL contains a fragment identifier that relates to tabbed content, i.e. `#tab-similar` or `#tab-warranty`, the correct tab should appear when the page loads.

#### Related Issue

#1136

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Fix Product Options hiding Add to Cart on a Google AMP page [#1214](https://github.com/bigcommerce/cornerstone/pull/1214)
 - Hide blank review stars when there are no reviews on a product [#1209](https://github.com/bigcommerce/cornerstone/pull/1209)
 - Fix styling of subpage links in Contact Us page [#1216](https://github.com/bigcommerce/cornerstone/pull/1216)
+- Add lazyloading to main product video and fix video thumbnail bug [#1217](https://github.com/bigcommerce/cornerstone/pull/1217)
 
 ## 1.17.0 (2018-04-26)
 - Fix empty object issue in app.js [#1204](https://github.com/bigcommerce/cornerstone/pull/1204)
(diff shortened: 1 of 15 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 - Add the +/- icons for the category filtering [#1211](https://github.com/bigcommerce/cornerstone/pull/1211)
 - Fix overlapping logo when using "original" sizing and large logos [#1213](https://github.com/bigcommerce/cornerstone/pull/1213)
 - Fix Product Options hiding Add to Cart on a Google AMP page [#1214](https://github.com/bigcommerce/cornerstone/pull/1214)
+- Fix syntactically incorrect code in carousel and bulk-discount-rates components
+- Fix the possibility of duplicate IDs occurring in product options in certain situations
```
junedkazi: Instead of 3 entries. Please move url change into a separate PR. Also just add one changelog entry per PR with link to the PR. So it would look something like this ```
Fix duplicate IDs occurrence in product options in certain situations & syntax fix in carousel and bulk-discount-rates components [#1215](https://github.com/bigcommerce/cornerstone/pull/1215)
```

File: templates/components/products/options/input-checkbox.html [hunk 2]
```diff
@@ -7,17 +7,17 @@
         {{/if}}
     </label>
 
-    <input 
+    <input
         class="form-checkbox"
         type="checkbox"
         name="attribute[{{id}}]"
-        id="attribute-{{id}}"
+        id="attribute_check_{{id}}"
```
junedkazi: Can we just stick with using `-` instead on `_`  🍹

flair-duncan:
Hi @junedkazi - changes made as requested, and separated the other issue into PR #1220

Issue #1217, opened 2018-05-02
Lazyload product videos - videos.html


PeteyRev:
Any reason why this is not being utilized?

```
            <iframe
                id="player"
                type="text/html"
                width="640"
                height="390"
                frameborder="0"
                webkitAllowFullScreen
                mozallowfullscreen
                allowFullScreen
                class="lazyload"
                data-src="//www.youtube.com/embed/{{this.featured.id}}"
                data-video-player>
            </iframe>

```

Id be happy to make a pull request, just wondering since the video thumbnails are being lazy loaded

bookernath:
PR welcome!

PeteyRev:
@bookernath Thanks, [done](https://github.com/bigcommerce/cornerstone/pull/1218)

I also noticed a bigger issue that is fixed in the same PR, the video thumbnails when clicked aren't changing the main video

junedkazi:
Fixed in https://github.com/bigcommerce/cornerstone/pull/1221

PR #1218, opened 2018-05-02
add lazyload to main video and fixed thumbnail bug

PeteyRev:
#### What?

I originally made this just to lazyload the main video on product page load, but then I noticed that there is an existing bug that will not update the main video when the thumbnails are selected

#### Tickets / Documentation

[#1217](https://github.com/bigcommerce/cornerstone/issues/1217)

[video](https://screencast.com/t/f8jWdRIGmf4)

### DIFF
--- a/templates/components/products/card.html
+++ b/templates/components/products/card.html
@@ -61,13 +61,13 @@
         </figcaption>
     </figure>
     <div class="card-body">
-        {{#if show_rating}}
+        {{#and rating show_rating}}
             <p class="card-text" data-test-info-type="productRating">
                 <span class="rating--small">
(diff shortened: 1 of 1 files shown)


Ubersmake:
@PeteyRev Thanks for this! I've tested this fix against Chrome/FF/Safari/IE11.

Could you also add an entry into CHANGELOG.md under "Draft" describing your changes? See https://github.com/bigcommerce/cornerstone/commit/e10049fe8cd5da0b3a57c0b57c4de5759f3253ec#diff-4ac32a78649ca5bdd8e0ba38b7006a1e as an example.

PeteyRev:
@junedkazi should be gtg now

PR #1229, opened 2018-05-08
Cornerstone 2.0: Performance improvements

mattolson:
# What?
The goal of this release is to improve frontend theme performance. This is the work of @mattolson, @junedkazi and @BC-EChristensen during the Stranger Hackings hackathon.

## Remove Modernizr
* Modernizr is a blocking script that has very little benefit for us. Removing this script improves page responsiveness.
* We used Modernizr via the css classes `csscolumns`, `flexbox`, `objectfit`, and `js`.
* The `csscolumns` class was used for productMasonry for older browsers, but since then all of our supported browsers support css columns. There is one small problem with Firefox: it does not support break-inside, but it does support page-break-inside, which does the same thing. See https://caniuse.com/#feat=multicolumn
* The `flexbox` class is unused
* The `objectfit` class is used by carousel. Reimplmented without using Modernizer. The previous implmentation put a background image on the wrapper div and then hid it if object-fit is supported. The new implementation does not put the background image on the wrapper div by default, but instead assumes that your browser supports object-fit (all ours do except for IE). For IE, we use javascript to copy the image source from the image tag to the background image of the wrapper div and then hide the image.
* The `js` class is used for a few things in css, so mimic this behavior with a simple inline script.

## Save space for carousel
* Add wrapper div for hero carousel that has the appropriate height for the image to be displayed after lazy loading.
* This improves frontend performance because the browser will be less likely to load images below the fold immediately.

## Add dns prefetch hints
* Add dns prefetch & preconnect to core resources that we know we will need (configured cdn and font providers). This uses a new helper `resourceHints` that was introduced in paper 2.0.8.

## Webpack 4 & Javascript optimizations
* Upgrade to webpack 4
* Clean up stencil bootstrap
* Simplify interface for PageManager and get rid of async library dependency
* Get rid of html5-history-api library dependency (no longer needed)
* Get rid of fastclick library dependency (no longer needed)
* Upgrade several babel dependencies
* Separate webpack config into common, dev, prod, and add npm scripts to build
* Get rid of minifications in babel loader, instead rely on webpack
* Get rid of LoaderOptionsPlugin
* Get rid of CommonsChunkPlugin (webpack 4 will automaticaly do this for prod)

## Inject svg via ajax
* Use svg-injector to fetch the svg icon sprite rather than embed it directly on the page. This improves frontend performance because the icons will be cached after the first request and we reduce the page weight by about 28kb.
* Move location of generated svg so it is retrievable by the frontend through the CDN.
* Update svgstore grunt task

## Reduce defaults for product counts
* Reduce the default number of featured, new, and bestselling products on the home page.
* These can always be updated in Theme Editor, but the defaults don't need to be so high.

## Lazier image loading
* Use laziest possible setting for images

# Screenshots
Using lighthouse for benchmarking against a production store, I found the following:

#### Desktop
screen shot 2018-05-07 at 5 29 36 pm

#### Mobile
screen shot 2018-05-07 at 5 30 19 pm

#### Desktop
screen shot 2018-05-11 at 4 23 04 pm

#### Mobile
screen shot 2018-05-11 at 4 20 21 pm


bookernath:
I'm not qualified to review this but I am totally stoked about it!

templates/layout/base.html
```diff
@@ -2,9 +2,17 @@
 <html class="no-js" lang="{{ locale_name }}">
     <head>
         <title>{{ head.title }}</title>
+
+        <link rel="dns-prefetch preconnect" href="//cdn8.bigcommerce.com" crossorigin>
```
bookernath: should these be hardcoded, or can/should we populate them via a variable somehow?

CC @chrisboulton

chrisboulton: They should be fed in via a variable somehow. We have various endpoints, which depend on store assignment or store environment.

junedkazi: I am pretty sure we are seeding it for stapler to generate asset urls.

config.json
```diff
@@ -41,9 +41,9 @@
     "Firefox ESR"
   ],
   "settings": {
-    "homepage_new_products_count": 12,
-    "homepage_featured_products_count": 8,
-    "homepage_top_products_count": 8,
+    "homepage_new_products_count": 5,
```
junedkazi: Let’s keep this consistent with the other and make it 4 :tropical_drink:

junedkazi:
Also just want to confirm with hthat screenshots are correct. The time stamp for the before is later than the after if it makes sense 😜

webpack.prod.js
```diff
@@ -0,0 +1,12 @@
+var webpack = require('webpack');
```
tpietsch: is having two files how webpack deals with env?

webpack.common.js
```diff
@@ -1,13 +1,12 @@
-var CleanWebpackPlugin = require('clean-webpack-plugin'),
-    config = require('./config.json'),
-    LodashModuleReplacementPlugin = require('lodash-webpack-plugin'),
+var CleanPlugin = require('clean-webpack-plugin'),
```
tpietsch: var?

tpietsch:
noticed lots of VARS for the require dependecys that prob should be const right?

templates/layout/amp-iframe.html
```diff
@@ -9,7 +9,11 @@
         <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1">
         {{{stylesheet '/assets/css/theme.css'}}}
         {{ getFontsCollection }}
-        <script src="{{cdn 'assets/modernizr-custom.js'}}"></script>
+
+        <script>
+            // Change document class from no-js to js so we can detect this in css
+            document.documentElement.className = 'js';
```
mjschock: what is this about?  seems a little hacky...  can we get the same result another way?

templates/layout/base.html
```diff
@@ -2,14 +2,26 @@
...
+        <link rel="dns-prefetch preconnect" href="//cdn7.bigcommerce.com" crossorigin>
+        <link rel="dns-prefetch preconnect" href="//fonts.googleapis.com" crossorigin>
```
junedkazi: Also all of this meta needs to be seeded from bcapp so we are in full control if we need to change it in the future without making any updates to the theme.

config.json
junedkazi: We should confirm this change does not break any tests which are looking for X products

templates/layout/base.html
mattolson: I'll change this so the cdn paths come from the site settings, which is available in the theme context.

config.json
mattolson: I made it five because we use a carousel for this element and I wanted to show off its capabilities.

mattolson:
@junedkazi yes the screenshots are correct. I was switching back and forth between Cornerstone 1.17 and a custom theme based on this branch. I did Cornerstone 1.17 after my branch.

webpack.prod.js
mattolson: It's not the only way, but it's the way I chose. It let's us extract common config and then extend it per environment. https://simonsmith.io/organising-webpack-config-environments/

templates/layout/base.html
mattolson: Yep, I will update the cdn preconnect to come from the context. However, the use of Google Fonts is a theme choice, so it seems appropriate to put it in the theme.

templates/layout/amp-iframe.html
mattolson: This is the last remnant of removing Modernizr. Modernizr does essentially this same thing, and this `js` class name is used by css to present alternatives when Javascript is enabled or not.

webpack.common.js
mattolson: Admittedly, the codebase has not been cleaned up fully to work with ES6 syntax. I can do it for the webpack configuration, but I don't want to do a full code cleanup as part of this PR.

danieldelcore:
Glorious 👏 PR  👏 Description  👏

assets/scss/components/stencil/heroCarousel/_heroCarousel.scss
```diff
@@ -4,42 +4,36 @@
...
 //    However, on the homepage, we want the carousel to be flush with the header.
 //
-// 4. Allows image to scale on large screens while preventing the top and bottom
+// 3. Allows image to scale on large screens while preventing the top and bottom
 //    from becoming cut off.
 // -----------------------------------------------------------------------------
 
 .heroCarousel {
-    margin-bottom: (spacing("double") + spacing("single"));
+    margin-bottom: 30px; // from slick-theme.scss
```
icatalina: 🍹 no `remCalc`?

templates/layout/amp-iframe.html
icatalina: This is a quite common pattern. Not too sure about the way it is done, this is not `changing the document class from no-js to js` it is replacing whatever it is in the body with `js`.

It might be better if we are more granular with the change:

```
document.documentElement.classList.remove('no-js');
document.documentElement.classList.remove('js');
```

assets/js/app.js
```diff
@@ -63,32 +63,16 @@ window.stencilBootstrap = function stencilBootstrap(pageType, contextJSON = null
...
-                if (globalClass) {
-                    globalClass.load();
-                }
-
-                if (pageClass) {
-                    pageClass.load();
+                // Find the appropriate page loader based on pageType
+                const pageClassImporter = pageClasses[pageType];
+                if (typeof pageClassImporter === 'function') {
+                    const PageClass = (await pageClassImporter()).default;
```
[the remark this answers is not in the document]
zackangelo: Yeah, `await` is non-blocking in JS, so not like the `Await` object in Scala. Fun fact, there's an async/await macro library for Scala that does the same thing: https://github.com/scala/scala-async 

It basically rewrites the code to use `.map` and `.flatMap` in the compiler.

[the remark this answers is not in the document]
mattolson: @icatalina Yes, we are running everything through babel in stencil CLI.

templates/layout/amp-iframe.html
BC-EChristensen: Yeah that bugged me too when I was working in this code, @icatalina, might be outside the scope of this PR to worry about changing this but I do think that approach would be both more clear and safer.

templates/layout/base.html
```diff
@@ -2,14 +2,21 @@
...
         {{{ head.meta_tags }}}
         {{{ head.config }}}
         {{#block "head"}} {{/block}}
+
         <link href="{{ head.favicon }}" rel="shortcut icon">
         <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1">
+
+        <script>
+            // Change document class from no-js to js so we can detect this in css
+            document.documentElement.className = document.documentElement.className.replace('no-js', 'js');
```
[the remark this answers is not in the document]
mattolson: This is just using the Javascript String class replace method on the `className` attribute, which is available on all browsers. I'm not using `classList` here.

templates/layout/amp-iframe.html
mattolson: I'm not using `classList` here, but `className`, which is a string, and then using Javascript's String replace on it. Should be safe to use in all browsers. This is a pretty standard thing, and exactly what Modernizr is doing under the hood. 

https://github.com/Modernizr/Modernizr/blob/5eea7e2a213edc9e83a47b6414d0250468d83471/src/setClasses.js#L22

templates/layout/base.html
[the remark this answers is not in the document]
mattolson: FWIW, this is what Modernizr does as well: https://github.com/Modernizr/Modernizr/blob/5eea7e2a213edc9e83a47b6414d0250468d83471/src/setClasses.js#L22

assets/scss/components/stencil/heroCarousel/_heroCarousel.scss
mattolson: Reverting this change, it is unnecessary.

mattolson:
@icatalina ♻️

templates/layout/amp-iframe.html
mattolson: Oh wait, I see -- I made a change to this in `base.html` but forgot to bring it into `amp-iframe.html`. I'll do that now.

BC-EChristensen: Yeah sorry to be clear I was talking about `className = 'js'`, not `className.replace`

templates/layout/base.html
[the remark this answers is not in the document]
icatalina: Oh, sorry misread that, hehe

icatalina:
hi @mattolson !

what about the await? https://github.com/bigcommerce/cornerstone/pull/1229#discussion_r187503667

mattolson:
@icatalina I already replied inline there -- we run everything through Babel as part of Stencil CLI.

assets/js/theme/account.js
```diff
@@ -27,7 +27,7 @@ export default class Account extends PageManager {
         this.passwordRequirements = this.context.passwordRequirements;
```
tpietsch: screen shot 2018-05-17 at 11 54 15 am
think this is causing a problem for the CustomerAddressBookUITest

config.json
tpietsch: $this->assertCount(12, $productsFromUI, 'Expected to see 12 new products');
NewProductsUITest.php

assets/js/theme/account.js
tpietsch: screen shot 2018-05-17 at 12 24 54 pm
Functional\Tests\Customers\Stencil\CustomerSignInUITest

assets/js/app.js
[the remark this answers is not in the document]
tpietsch: await is diff than a promise because it blocks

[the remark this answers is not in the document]
tpietsch: const pageClass = await xx  <-- how can this be async what if somone uses the pageClass val before the await is resolved?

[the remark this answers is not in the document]
tpietsch: actually....i think this might be causing the test failure RegisteredCustomerMessagesUITest::testRequiredFieldWhileCreatingMessage

config.json
```diff
@@ -41,9 +41,9 @@
...
+    "homepage_featured_products_count": 4,
```
tpietsch: FeaturedAndPopularProductsUITest failure here - number of expected needs to be updated in the test

assets/js/theme/account.js
tpietsch: tests/Functional/Tests/Customers/Stencil/CreateCustomerUITest.php

tpietsch: tests/Functional/Tests/Customers/Stencil/RegisteredCustomerAccountSettingsUITest.php
tests/Functional/Tests/Customers/Stencil/RegisteredCustomerWishListUITest.php

assets/js/theme/page-manager.js
```diff
@@ -1,35 +1,22 @@
...
-    after(next) {
-        next();
+    onReady() {
     }
 
-    type() {
-        return this.constructor.name;
-    }
+    static load(context) {
+        const page = new this(context);
```
tpietsch: const page = new PageManager(context); ?

new this(context) ? what would that even refer to?

assets/js/app.js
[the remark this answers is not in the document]
mjschock: @tpietsch for the `const pageClass = await xx` example, it's simply sugar for using a promise.then, so it will not set the pageClass val until after the await is resolved.  but it's non-blocking in that the thread is released when invoking await to handle other things.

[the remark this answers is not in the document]
tpietsch: it just kinda not my face cause now you have a variable in scope that you can use where the value  might change right? first its undef then its def and between that time it can be read from incorrectly
In a normal Promise(resolve -> resolve.data) <-- you are more explicit about when the data is being used here it seems like it could cause intermittent weirdness. And also the variable is declared as a const which i would expect as non-changing value

config.json
```diff
@@ -41,9 +41,9 @@
...
+    "homepage_top_products_count": 4,
```
Ubersmake: Updating the new/featured/popular products counts in tests. There'll be another PR for that.

assets/js/theme/account.js
mattolson: Fixed. Thanks for looking into this.

config.json
mattolson: Yes, we'll need to update the tests when this is published.

assets/js/theme/page-manager.js
mattolson: This will call the subclass constructor.

Ubersmake:
💚 Stencil regression suite has passed with the exception of those tests that hit the changed featured/new/popular product counts.

config.json
Ubersmake: Tracking test updates in https://github.com/bigcommerce/bigcommerce/pull/24622.

config.json
mattolson: Yep, we'll update the tests once this is published.

mattolson:
@davidchin I'm sorry, I didn't see your review prior to merging. I will address these comments in a future PR. We intend to continue to invest in Cornerstone performance.

davidchin:
@mattolson no worries.

.travis.yml
```diff
@@ -1,13 +1,13 @@
...
 sudo: required
 dist: trusty
 install:
-- bundle install
-- npm install -g grunt-cli
-- npm install -g bigcommerce/stencil-cli
-- npm install
+  - bundle install
+  - npm install -g grunt-cli
+  - npm install -g bigcommerce/stencil-cli
```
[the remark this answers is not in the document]
mattolson: https://github.com/bigcommerce/cornerstone/pull/1390

assets/js/theme/common/carousel.js
```diff
@@ -7,4 +7,16 @@ export default function () {
     if ($carousel.length) {
         $carousel.slick();
     }
+
+    // Alternative image styling for IE, which doesn't support objectfit
+    if (typeof document.documentElement.style.objectFit === 'undefined') {
+        $('.heroCarousel-slide').each(() => {
+            const $container = $(this);
```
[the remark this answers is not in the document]
mattolson: https://github.com/bigcommerce/cornerstone/pull/1390

webpack.common.js
```diff
@@ -1,13 +1,12 @@
...
-    LodashModuleReplacementPlugin = require('lodash-webpack-plugin'),
-    path = require('path'),
-    webpack = require('webpack');
+const CleanPlugin = require('clean-webpack-plugin'),
+      LodashPlugin = require('lodash-webpack-plugin'),
+      path = require('path'),
+      webpack = require('webpack');
 
+// Common configuration, with extensions in webpack.dev.js and webpack.prod.js.
 module.exports = {
```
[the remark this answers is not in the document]
mattolson: https://github.com/bigcommerce/cornerstone/pull/1390

webpack.common.js
```diff
@@ -49,36 +45,25 @@ module.exports = {
...
-            'async': path.resolve(__dirname, 'node_modules/async/dist/async.min.js'),
-            'html5-history-api': path.resolve(__dirname, 'node_modules/html5-history-api/history.min.js'),
-            jquery: path.resolve(__dirname, 'node_modules/jquery/dist/jquery.min.js'),
             'jquery-migrate': path.resolve(__dirname, 'node_modules/jquery-migrate/dist/jquery-migrate.min.js'),
-            'jquery-zoom': path.resolve(__dirname, 'node_modules/jquery-zoom/jquery.zoom.min.js'),
             jstree: path.resolve(__dirname, 'node_modules/jstree/dist/jstree.min.js'),
             'pace': path.resolve(__dirname, 'node_modules/pace/pace.min.js'),
             'slick-carousel': path.resolve(__dirname, 'node_modules/slick-carousel/slick/slick.min.js'),
+            'svg-injector': path.resolve(__dirname, 'node_modules/svg-injector/dist/svg-injector.min.js'),
             sweetalert2: path.resolve(__dirname, 'node_modules/sweetalert2/dist/sweetalert2.min.js'),
```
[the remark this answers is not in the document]
mattolson: I tried removing this and it didn't have any impact on the bundle sizes.

package.json
```diff
@@ -21,20 +18,21 @@
...
+    "babel-eslint": "^8.2.2",
+    "babel-loader": "^7.1.4",
+    "babel-plugin-dynamic-import-webpack": "^1.0.2",
+    "babel-plugin-lodash": "^3.3.2",
     "babel-plugin-syntax-dynamic-import": "^6.18.0",
     "babel-plugin-transform-regenerator": "^6.26.0",
     "babel-plugin-transform-runtime": "^6.23.0",
-    "babel-preset-env": "^1.2.2",
-    "clean-webpack-plugin": "^0.1.17",
+    "babel-preset-env": "^1.6.1",
```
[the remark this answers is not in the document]
mattolson: Because we are using [useBuiltIns](https://github.com/bigcommerce/cornerstone/blob/master/webpack.common.js#L30), this will tree-shake babel-polyfill to just what we are using. However, I will also add a .browserslistrc file to set targets just in case.

PR #1232, opened 2018-05-10
fix(storefront): STRF-4875 Fix for sort query parameter with faceted search.

Ubersmake:
#### What?

The sort search parameter was being cleared in faceted search (product filtering) after a range facet was set or updated.

This issue appears if a sort is done *before* applying a range. It does not appear if a sort is applied *after* applying a range, although updating a range will cause the issue to appear.

The root cause is a collision in the template on "sort", which might be looking at a function with the same name rather than the url/search parameter.

This also gets rid of brackets appearing in some search parameters when a range is applied.

#### Tickets / Documentation

- [STRF-4875](https://jira.bigcommerce.com/browse/STRF-4875)

#### Screenshots

Before:

before

After:

after


templates/components/faceted-search/facets/range.html
```diff
@@ -23,11 +23,11 @@ <h5 class="accordion-title">
     <div id="facetedSearch-content--{{hyphenate facet }}" class="accordion-content {{#unless ../start_collapsed }} is-open {{/unless}}">
         <form id="facet-range-form" class="form" method="get" data-faceted-search-range novalidate>
             {{#each current_selected_items}}
-                <input type="hidden" name="{{ param_name }}[]" value="{{ param_value }}"/>
```
lord2800: This could actually be problematic. What `param_name` values actually come out here? I can verify that none are treated as an array, but if any are actually supposed to be arrays, this change will break that.

bc-vandana:
Cool. Thank you for fixing this @Ubersmake

templates/components/faceted-search/facets/range.html
Ubersmake: I'll undo that change because other things are using []. But for the record,

*Without* touching range I can get a url parameter string like:
`?_bc_fsnf=1&brand[]=36&brand[]=35&Color[]=Black&Color[]=Blue`

Which has identical results to:
`?_bc_fsnf=1&brand=36&brand=35&Color=Black&Color=Blue`

Values aren't being added to the parameters in any sort of array-like structure. Their keys are simply repeated.

Specific to changes in `range.html`, with the brackets `?brand=37&brand=36&search_query=&min_price=1&max_price=99` returns the same results as pre-change `?brand[]=37&brand[]=36&search_query=&min_price=1&max_price=99`

anilgautamm:
@mattolson @Ubersmake  There is another problem with product filtering when you keep the price filter in the top, and color or other filters in the bottom, when we select a color filter and then select the price range filter, it automatically clears the other filter other than price filter.

junedkazi:
@anilgautamm can you please create an issue in the issue queue so it is looked into.

PR #1236, opened 2018-05-12
Remove duplicate import of normalize at theme.scss

jbruni:
#### What?

Normalize is already being imported at line 35.

It does not seem necessary to import it a second time, again, at line 71.

#### Tickets / Documentation

N/A

N/A


jbruni:
Forget it. The first import is only importing the settings. :man_facepalming:

PR #1238, opened 2018-05-14
fix(storefront): STRF-4786 Do not escape blog summary HTML.

Ubersmake:
#### What?

Blog summary content was being escaped, resulting in space characters to show up as `&nbsp;`. Now that content is being un-escaped like content in blog posts.

Note that this does not apply to all whitespace characters. Tabs are not treated as "tabs" in the summary and will not show up. Newlines are not part of the blog summary.

#### Tickets / Documentation

- [STRF-4786](https://jira.bigcommerce.com/browse/STRF-4786)
- [Handlebars HTML Escaping](http://handlebarsjs.com/#html-escaping)

Before:

before

After:

after

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Draft
 - Fix for sort disappearing on range update with product filtering [#1232](https://github.com/bigcommerce/cornerstone/pull/1232)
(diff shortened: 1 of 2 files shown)


mattolson:
@Ubersmake Please rebase

Issue #1206, opened 2018-04-20
Is "browserlist" file used?


jbruni:
@mjschock - I see this `browserlist` file has been introduced at this commit: https://github.com/bigcommerce/cornerstone/commit/b2b9007fa094608c9afa3dc364f69b28a080d8aa#diff-942e0874b8917b9f3d914cb757c8556a and never changed since then.

I could not find anything else in the repository and in that commit which actually makes usage of this file. As far as I've looked, I could not find any Webpack plugin or configuration actually making usage of it.

In fact, I found at Browserlists' documentation that one of the expected configuration file names is `browserslist` (note the PLURAL "browsers" instead of singular "browser").

- https://evilmartians.com/chronicles/autoprefixer-7-browserslist-2-released
- https://github.com/browserslist/browserslist

So... it looks like to me this `browserlist` has no effect on anything. But I'm possibly wrong. Can you elucidate?

Thanks!

mjschock:
@jbruni - it appears you're right, that file is not in use.   IIRC, that was added for the babel env preset.  but, the name is wrong as you said (it should be browser*s*list) and the babel env preset will not make use of a browserslist file until babel 7.

anyway, thanks for bringing this to our attention.  please feel free to create a PR that either removes it or renames it (in prep for babel 7).  regarding the actual behavior, babel preset will be using `> 0.5%, last 2 versions, Firefox ESR, not dead` as the default instead.

in the meantime, if you would like to adjust the browsers that will be targeted, you can configure the env preset to do so.  here are the docs:  https://babeljs.io/docs/plugins/preset-env/#targetsbrowsers

Ubersmake:
Thanks @jbruni . I'm going to close because:
1. Browserlist wasn't used.
2. Browserlist definitely won't be used anymore after #1241.

PR #1241, opened 2018-05-18
Delete browserlist

jbruni:
#### What?

As discussed at #1206, this file is currently useless.

When the time comes (Babel 7 release), a `browserslist` (plural at browser**s**) can be introduced.

#### Tickets / Documentation

- [Is "browserlist" file used?](https://github.com/bigcommerce/cornerstone/issues/1206)

### DIFF
--- a/browserlist
+++ b/browserlist
@@ -1,3 +1,0 @@
-> 1%
-last 2 versions
-Firefox ESR


bookernath:
@jbruni if all goes well it could be as early as next week

PR #1245, opened 2018-05-22
fix(content): STRF-2456 Fix product rating reviews

davidwrpayne:
#### What?
We should still show the ratings section when there are no reviews. We just don't want to show the schema meta tags with it.
@mattolson @mjschock @bigcommerce/storefront-team

### DIFF
--- a/templates/components/products/product-view.html
+++ b/templates/components/products/product-view.html
@@ -29,38 +29,36 @@
                 {{/or}}
             </div>
             {{{region name="product_below_price"}}}
-            {{#if product.rating}}
-                <div class="productView-rating"{{#if schema}} itemprop="aggregateRating" itemscope itemtype="http://schema.org/AggregateRating"{{/if}}>
-                    {{#if settings.show_product_rating}}
-                        {{#if product.num_reviews '>' 0}}
-                            {{#if schema}}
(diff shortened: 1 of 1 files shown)


templates/components/products/product-view.html
```diff
@@ -29,38 +29,36 @@ <h2 class="productView-brand"{{#if schema}} itemprop="brand" itemscope itemtype=
...
-            {{#if product.rating}}
-                <div class="productView-rating"{{#if schema}} itemprop="aggregateRating" itemscope itemtype="http://schema.org/AggregateRating"{{/if}}>
-                    {{#if settings.show_product_rating}}
-                        {{#if product.num_reviews '>' 0}}
-                            {{#if schema}}
-                                <meta itemprop="ratingValue" content="{{product.rating}}">
-                                <meta itemprop="ratingCount" content="{{product.num_reviews}}">
-                                <meta itemprop="reviewCount" content="{{product.num_reviews}}">
-                            {{/if}}
+            <div class="productView-rating"{{#if product.num_reviews '>' 0}}{{#if schema}} itemprop="aggregateRating" itemscope itemtype="http://schema.org/AggregateRating"{{/if}}{{/if}}>
```
davidwrpayne: Couldn't use {{#and }} here. I think its because of the conditional.

PR #1247, opened 2018-05-23
Avoid undefined context in WishList instance

jbruni:
#### What?

The value of `this.context` is `undefined` within WishList class instance, in the product and account pages. The class `constructor` expects a `context` parameter, but no parameter is provided when the instance is created from "product-details.js" and "account.js".

This has no side-effect when no customization is made to the theme. But any customization will expect `this.context` to have the `context` data, rather than `undefined`.

#### Tickets / Documentation

I have reported it here, but it seems it has been ignored or not understood:

- https://github.com/bigcommerce/cornerstone/commit/31bdddbf64293d85192cae00e966d531f527b150#r29038225

Thank you.

### DIFF
--- a/assets/js/theme/account.js
+++ b/assets/js/theme/account.js
@@ -27,7 +27,7 @@ export default class Account extends PageManager {
         this.passwordRequirements = this.context.passwordRequirements;
 
         // Instantiates wish list JS
-        Wishlist.load();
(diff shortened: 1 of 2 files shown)


jbruni:
If the issue is difficult to understand, please let me know. I can record a short video showing what is expected, and what happens instead.

PR #1248, opened 2018-05-25
feat(content): STRF-4925 add newsletter summary to the newsletter subscribe button

davidwrpayne:
#### What?
We need to display a summary in case of GDPR regulations. 

#### Tickets / Documentation
- [STRF-4925](https://jira.bigcommerce.com/browse/STRF-4925)


junedkazi:
Can we get a screenshot of before & after please.

PR #1262, opened 2018-06-08
fix(storefront): STRF-4701 - Fixed a use case that prevented retail/sale prices from displaying on PDP

bc-jz:
#### What?
Updated PDP price and price-range templates to always include containing elements for rrp, non-sale, and sale prices even if they are not present in the context on initial page load. This ensures the proper elements are available to be updated by Javascript when option seleections are made that do have any of the prices available.

#### Tickets / Documentation
https://jira.bigcommerce.com/browse/STRF-4701

Pricing behavior examples:
https://docs.google.com/document/d/1rXL4t6fQh70bHZ6mnoPvvFcVIh9ZOjcLtfAdlPSqFEc/edit?usp=sharing


templates/components/products/price-range.html
```diff
@@ -1,21 +1,44 @@
...
                 <meta itemprop="valueAddedTaxIncluded" content="true">
             </div>
         {{/if}}
     </div>
+    {{else}}
+        {{#if price.with_tax}}
+            <div class="price-section price-section--withTax rrp-price--withTax" style="display: none;">
+            {{lang 'products.retail_price'}}
+            <span data-product-rrp-with-tax class="price price--rrp">
+                {{#if price.rrp_with_tax}}
```
Souldjinn: might want to have this if check rest above `{{lang 'products.retail_price'}}` that way it doesnt show the retail price text without there being a price to display.

templates/components/products/price-range.html
```diff
@@ -33,8 +56,47 @@
         {{/if}}
     </div>
 {{/and}}
+{{#and price.retail_price_range.min.without_tax price.retail_price_range.max.without_tax}}
+        <div class="price-section price-section--withoutTax rrp-price--withoutTax" {{#if schema_org}}itemprop="offers" itemscope itemtype="http://schema.org/Offer"{{/if}}>
+            {{lang 'products.retail_price'}}
+            <span data-product-rrp-price-without-tax class="price price--rrp">{{price.retail_price_range.min.without_tax.formatted}} - {{price.retail_price_range.max.without_tax.formatted}}</span>
+            {{#if schema_org}}
```
bc-jz: I am curious about this meta content......I included it because it was already in the code that I was sampling but does it make sense for us to have these meta pricing properties for a retail price range?

I noticed that regarding an individual variant price, we do not include meta properties for a retail price - only the actual price.  So maybe these metas are safe to leave off of retail pricing or they should be added to the retail price for non-range pricing.

templates/components/products/price.html
```diff
@@ -2,19 +2,23 @@
...
     {{#if price.with_tax}}
-        <div class="price-section price-section--withTax rrp-price--withTax">
-            {{#if price.rrp_with_tax}}
-                {{lang 'products.retail_price'}}
-                <span data-product-rrp-with-tax class="price price--rrp"> {{price.rrp_with_tax.formatted}}</span>
-            {{/if}}
+        <div class="price-section price-section--withTax rrp-price--withTax" {{#unless price.rrp_with_tax}}style="display: none;"{{/unless}}>
+            {{lang 'products.retail_price'}}
+            <span data-product-rrp-with-tax class="price price--rrp">
+                {{#if price.rrp_with_tax}}
```
Souldjinn: same here with moving the if statement up a bit? maybe you had a reason for showing the text though.

Souldjinn: Ah, actually, if I understand this right, you can probably take these if statements out entirely-- since you're basically changing the way you chose wether or not to display with the `style="display: none;"` That should probably do it.

bc-jz: I think I need the IF statement because I am not sure if `price.rrp_with_tax` exists....not sure how stencil acts if I try to print a value that is not there.  If if does not break anything to try and print a value that does not exist (ex:`{{price.rrp_with_tax.formatted}}`) then i can leave out those IFs.

The style display none ensures that the "Retail price" label will not be seen until toggled on by the JS.

templates/components/products/price-range.html
bc-jz: The retail price text will not show until JS updates this element to no longer be hidden.

templates/components/products/price.html
Souldjinn: I just mean that you're already checking if theres a value for price with tax above:
```
<div class="price-section price-section--withTax rrp-price--withTax" {{#unless price.rrp_with_tax}}style="display: none;"{{/unless}}>
```

you dont probably need it again 
```
  {{#if price.rrp_with_tax}}
```
right?

bc-jz: I'll test what happens if I exclude it....the `{{#unless..}}` conditional will make the element hide when there is not a retail price value to enter....im just not sure how stencil will react if i try to access a non-existent property

templates/components/products/price-range.html
```diff
@@ -1,21 +1,28 @@
...
-                <meta itemprop="price" content="{{price.retail_price_range.min.without_tax.value}}">
-                <meta itemprop="maxPrice" content="{{price.retail_price_range.max.without_tax.value}}"  />
-                <meta itemprop="priceCurrency" content="{{currency_selector.active_currency_code}}">
-                <meta itemprop="valueAddedTaxIncluded" content="true">
+{{#and price.retail_price_range.min.with_tax price.retail_price_range.max.with_tax}}
+    <div class="price-section price-section--withTax rrp-price--withTax" {{#if schema_org}}itemprop="offers" itemscope itemtype="http://schema.org/Offer"{{/if}}>
+        {{theme_settings.pdp-retail-price-label}}
+        <span data-product-rrp-price-with-tax class="price price--rrp">{{price.retail_price_range.min.with_tax.formatted}} - {{price.retail_price_range.max.with_tax.formatted}}</span>
+    </div>
+    {{else}}
```
Souldjinn: does this `else` block need to close somewhere?  not sure which `if` its connected to.

templates/components/products/price-range.html
```diff
@@ -1,21 +1,28 @@
...
+                {{theme_settings.pdp-retail-price-label}}
+                <span data-product-rrp-with-tax class="price price--rrp">
+                    {{price.rrp_with_tax.formatted}}
+                </span>
             </div>
         {{/if}}
-    </div>
 {{/and}}
 {{#and price.price_range.min.with_tax price.price_range.max.with_tax}}
+    <div class="price-section price-section--withTax non-sale-price--withTax" style="display: none;">
```
Souldjinn: Interesting that you chose to default with `display: none` here, should this also have an `unless` block around it?

templates/components/products/price-range.html
```diff
@@ -1,21 +1,28 @@
...
+        {{theme_settings.pdp-non-sale-price-label}}
+        <span data-product-non-sale-price-with-tax class="price price--non-sale">
+            {{price.non_sale_price_with_tax.formatted}}
```
Souldjinn: Price range, I think, is associated with calculated_price which I think would be like "sale price" in this context? worth double checking, but I'm assuming if you dont have a range then you want to display the more closely associated price value.

templates/components/products/price-range.html
```diff
@@ -1,21 +1,28 @@
...
+        </span>
+    </div>
     <div class="price-section price-section--withTax" {{#if schema_org}}itemprop="offers" itemscope itemtype="http://schema.org/Offer"{{/if}}>
+        <span class="price-label">{{theme_settings.pdp-price-label}}</span>
+        <span class="price-now-label" style="display: none;">{{theme_settings.pdp-sale-price-label}}</span>
```
Souldjinn: same concern here with the `unless` block. when the page loads and there is a price range, do you see it at first not display and then suddenly display after the js loads?

templates/components/products/price-range.html
```diff
@@ -33,8 +40,31 @@
         {{/if}}
     </div>
 {{/and}}
+{{#and price.retail_price_range.min.without_tax price.retail_price_range.max.without_tax}}
```
Souldjinn: I think you might want this above the `price_range` html, that way it would appear where the other retail price range appears.

templates/components/products/price.html
```diff
@@ -31,30 +32,31 @@
                 </div>
             {{/if}}
             {{#if price.without_tax}}
-                <abbr title="{{lang 'products.excluding_tax'}}">{{lang 'products.price_with_tax' tax_label=price.tax_label}}</abbr>
+                <abbr title="{{lang 'products.including_tax'}}">{{lang 'products.price_with_tax' tax_label=price.tax_label}}</abbr>
```
Souldjinn: 🙃 im not sure whats going on here: the if check block is `{{#if price.without_tax}}` but then we're showing including tax?

templates/components/products/price-range.html
bc-jz: This else block is connected to the `{{#and}}` on line 1 - it is the alternate path when the `{{#and}}` fails.  It does not need it's own closing line as it closes with the `{{/and}}`

templates/components/products/price-range.html
bc-jz: At least for right now a sale price should _never_ display if a price range is being displayed. At least that is my thinking. I want the element to be on the page but want to avoid having a `sale_price` from the lowest priced variant display next to a price range.

templates/components/products/price-range.html
bc-jz: The "price_range" we generate will take any sale price that is affecting a variant into account.  As a result the "price range" which we are displaying in the `price-section--withTax` block covers both sale and non-sale prices. The price range does not have a concept of letting the user know whether or not the prices included in it's range are on sale.

This "sale price" HTML block we have here though is not intended to display a price range.  It is specifically to for use for displaying the "non-sale price" of a variant after making option selections on the page.

templates/components/products/price-range.html
bc-jz: You will see the page load with a price range which then changes to an individual price _if_ you have default option selections enabled on the product and it has a price range. This is the same issue that exists right now, this PR unfortunately does not address that issue. I don't think that I can correct this JS loading issue by just adding an "unless" statement here. I'd either need to add some logic that tries to identify if there are any default options selected (which can be parsed from the option value content in the page context) or I would need to add a new property to the context that gives me this information (a `has_default_options` boolean property). I am aiming for the latter but it did not fit into the scope of this PR.

bc-jz: If you have a suggestion on how to use the "unless" though that maybe I am missing then lets discuss.

templates/components/products/price-range.html
bc-jz: I moved the `without_tax` range (and other without tax elements) below the with tax elements to make sure the order of these elements does not differ between the use case of having a price range on initial page load vs having a single price. On the single price display (seen in `prices.html`) we also have all of the with tax elements above the without_tax elements in a descending order of retail price, non-sale price, and sale price (or price).  I believe the way I have this page ordered will ensure that both codepaths produce the same order of displaying prices.

templates/components/products/price.html
bc-jz: Yep, this is specifically for the use case of displaying both including and excluding tax prices on the page. When the condition passes we have both price including and excluding tax in our context. In such a case we want to display identifying text after the price `(including tax)` and `(excluding tax)`, respectively.

Using this specific line as an example, we have immediately before this point printed the price including tax. So _if_ the context also has a price excluding tax, then we need to print `(Including tax)` to the page to sit right after the including tax price we printed a few lines above.

config.json
```diff
@@ -288,7 +288,11 @@
     "geotrust_ssl_common_name": "",
     "geotrust_ssl_seal_size": "M",
     "navigation_design": "simple",
-    "price_ranges": true
+    "price_ranges": true,
+    "pdp-price-label": "",
+    "pdp-sale-price-label": "Now:",
```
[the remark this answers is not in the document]
Souldjinn: very nice, dont want to remove the old strings from whatever lang files where in it before?

templates/components/products/price-range.html
Souldjinn: ah, then it should share indentation with the and on line 1

config.json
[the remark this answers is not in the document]
bc-jz: Thanks for the reminder!  I will do that.  I had meant to run a search on those language values and make sure they are not used elsewhere, then remove them if not.

templates/components/products/price-range.html
bc-jz: that makes sense!  It threw me off because it does not have it's own ending line, but you are right the indentation would be more clear that way

templates/components/products/price-range.html
Souldjinn: I must be confused about something, it seems like if options are selected, then what we want to be showing is `price.without_tax` or `price.with_tax` as the calculated price for the selection-- and `non_sale_price`  is a display price to reflect savings. I'm probably missing something here.

bc-jz: Once options are selected we do want to show the `price.without_tax` or `price.with_tax`.  We also want to show the `price.non_sale_price_without_tax` or `price.non_sale_price_with_tax`. The HTML you are referencing with this comment facilitates the latter.

Issue #1265, opened 2018-06-09
Duplicate input ID in product review form


xpothemes:
Hello,
I see one error on the theme where an ID is used multiple times.
http://cornerstone-light-demo.mybigcommerce.com/all/able-brewing-system/

![image](https://user-images.githubusercontent.com/36004695/41187071-b9cd54e8-6bcc-11e8-9697-068bb630c69f.png)

PR #1276, opened 2018-06-18
Fix duplicate input ID's in product review form

sacr3dc0w:
### What?

Duplicate ID's were in the DOM on the write review form. This was happening because ID's weren't 
being passed to the form components.

**Before**
`{{> components/common/forms/text name="revfromname" label=(lang 'products.reviews.form_write.name') value=product.reviews.author}}`

**After**
`{{> components/common/forms/text id="writeReview-name" name="revfromname" label=(lang 'products.reviews.form_write.name') value=product.reviews.author}}`

### Documentation
- #1265 

### Sreenshots

**Before**
01 - before

**After**
02 - after

### DIFF
--- a/templates/components/products/modals/writeReview.html
+++ b/templates/components/products/modals/writeReview.html
@@ -39,25 +39,25 @@
 
                 <!-- Name -->
                 {{#if product.reviews.author}}
-                    {{> components/common/forms/text name="revfromname" label=(lang 'products.reviews.form_write.name') value=product.reviews.author}}
+                    {{> components/common/forms/text id="revfromname" name="revfromname" label=(lang 'products.reviews.form_write.name') value=product.reviews.author}}
                 {{else}}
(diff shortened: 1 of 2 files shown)


templates/components/products/modals/writeReview.html
```diff
@@ -39,25 +39,25 @@ <h5 class="product-title">{{ product.title }}</h5>
 
                 <!-- Name -->
                 {{#if product.reviews.author}}
-                    {{> components/common/forms/text name="revfromname" label=(lang 'products.reviews.form_write.name') value=product.reviews.author}}
+                    {{> components/common/forms/text id="writeReview-name" name="revfromname" label=(lang 'products.reviews.form_write.name') value=product.reviews.author}}
```
junedkazi: Can we just use the `name` attribute as the `id` value so it is consistent with what we have on other forms in cornerstone.

PR #1278, opened 2018-06-19
Disable zoom and link for default "No Image" image.

Ubersmake:
#### What?

When a product has no images Cornerstone will display a default "No Image" image.

This image is zoomable, like other product images. And clicking on the image (if you can manage to click the 0x0 <a/> element) will open the full-size version of the image in the current view.

This PR removes that behavior. If a product has no images and thus displays the "No Image" image, the "No Image" image is treated like placeholder text. It does not zoom. It has no link.

#### Tickets / Documentation

- [STRF-5007](https://jira.bigcommerce.com/browse/STRF-5007)

##### Before

Product Details Page:

productbefore

Quick View:

quickbefore

##### After

Product Details Page:

productafter

Quick View:

quickafter

### DIFF
--- a/templates/components/products/product-view.html
+++ b/templates/components/products/product-view.html
@@ -151,17 +151,23 @@
 
     <section class="productView-images" data-image-gallery>
         <figure class="productView-image"
-                data-image-gallery-main
-                data-zoom-image="{{getImage product.main_image 'zoom_size' (cdn theme_settings.default_image_product)}}"
+                {{#if product.main_image}}
+                    data-image-gallery-main
+                    data-zoom-image="{{getImage product.main_image 'zoom_size' (cdn theme_settings.default_image_product)}}"
+                {{/if}}
                 >
             <div class="productView-img-container">
-                <a href="{{getImage product.main_image 'zoom_size' (cdn theme_settings.default_image_product)}}">
-                    <img class="productView-image--default lazyload" 
-                         data-sizes="auto" 
-                         src="{{cdn 'img/loading.svg'}}" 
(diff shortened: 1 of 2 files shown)


templates/components/products/product-view.html
```diff
@@ -151,17 +151,23 @@ <h2 class="productView-brand"{{#if schema}} itemprop="brand" itemscope itemtype=
...
-                         data-sizes="auto" 
-                         src="{{cdn 'img/loading.svg'}}" 
+                    <img class="productView-image--default lazyload"
+                         data-sizes="auto"
+                         src="{{cdn 'img/loading.svg'}}"
                          data-src="{{getImage product.main_image 'product_size' (cdn theme_settings.default_image_product)}}"
                          alt="{{product.main_image.alt}}" title="{{product.main_image.alt}}" data-main-image>
                 </a>
+                {{else}}
+                <img class="productView-image--default" src="{{cdn theme_settings.default_image_product}}">
```
junedkazi: spacing is off .. it should be 1 level deep.

templates/components/products/product-view.html
```diff
@@ -151,17 +151,23 @@ <h2 class="productView-brand"{{#if schema}} itemprop="brand" itemscope itemtype=
 
     <section class="productView-images" data-image-gallery>
         <figure class="productView-image"
+                {{#if product.main_image}}
                 data-image-gallery-main
                 data-zoom-image="{{getImage product.main_image 'zoom_size' (cdn theme_settings.default_image_product)}}"
+                {{/if}}
                 >
             <div class="productView-img-container">
+                {{#if product.main_image}}
```
junedkazi: Same spacing here .. move the content 1 level inside.

PR #1284, opened 2018-06-22
STRF-5012: Change file location of product default image in orders-li…

bwher:
…st.html

#### What?

Currently products with no image will spin indefinitely in the account order history section. Changing the file location of theme_settings.default_image_product to go up two directories rather than just one fixes the issue


junedkazi:
@bwher changes look good. Can you also add an entry in the changelog draft section please. See some of the open or closed PR on how to add that entry.

PR #1290, opened 2018-06-28
Fix initial aria attributes for 'Customers Also Viewed' products tab

jbruni:
#### What?

Just like the `is-active` class is being set for the "Customers Also Viewed" product tab initial rendering, we should also take care of the initial accessibility (`aria`) attributes for them.

#### Tickets / Documentation

- No links. Just look how the `is-active` class is handled, in the very same template file.


junedkazi:
@jbruni can you add a changelog entry & also some screenshots for before & after.

Issue #1065, opened 2017-08-11
Specs Tab


thetechie92:
Just` a crazy idea I did if people are interested for their stuff as well. You can take the product.custom_fields and turn it into a specs tab for your site.

...
`{{#if product.custom_fields}}`
`<li class="tab">`
    `<a class="tab-title" href="#tab-spec">Specifications</a>`
    `</li>`
`{{/if}}`
...
`{{#if product.custom_fields}}`
`<div class="tab-content" id="tab-spec">`
    `<dl class="productView-info">`
           ` {{#each product.custom_fields}}`
                    `<p><b>{{name}}</b> {{{value}}}</p>`
               `{{/each}}`
    `</dl>   `         
`</div>{{/if}}`

Ubersmake:
Closing this issue. This feature request is something we've talked about internally, but the delay is in figuring out both design and UX we're happy with for the majority of BC merchants. #1295 took it one step further, and that PR provides a great custom implementation of this feature.

sacr3dc0w:
@thetechie92 #1480

PR #1295, opened 2018-06-29
Add option to display custom fields under "Specifications" tab on PDP

sacr3dc0w:
#### What?

Adds an option to display custom fields under a "Specifications" tab.

#### Tickets / Documentation

- #1065 

**False, default**
01 false default

**False, default - tab region**
02 false tab region

**True**
03 true

**True, tab region - closed**
04 true tab closed

**True, tab region - open**
05 true tab open

**True, tab region - open, mobile**
true mobile

**True, tab region - closed**
06 theme editor


sacr3dc0w:
I've updated the PR and added the ability to change the tab label in the theme editor.

screen shot 2018-06-29 at 7 45 32 pm

sacr3dc0w:
Upon rebasing and enabling `show_product_reviews_tabs` from #1302, I discovered this will create display issues if a product contains warranty information.

Required adding `breakpoint("large")` for `.tab-title` and slightly reducing the side padding of the tabs for this viewport.

**Before**
1 before

**After**
2 after

schema.json
```diff
@@ -1173,6 +1173,18 @@
...
+      {
+         "type": "checkbox",
+         "label": "Product custom fields in tabs",
+         "force_reload": true,
+         "id": "show_custom_fields_tabs"
+      },
+      {
+        "type": "text",
+        "label": "Product custom fields tab label",
+        "force_reload": true,
```
BC-EChristensen: Since this setting isn't relevant if the above setting isn't checked, we should add

```
"reference": "show_custom_fields_tabs",
"reference_default": false,
```

This will cause this setting to be invisible if the above checkbox is not checked

bc-zoharmuzafi:
@sacr3dc0w  A new tab is a great idea and we would like to merge this pr, however, we would like to share the naming with our design team to get their approval

PR #1301, opened 2018-07-04
Fix blog_post import statement in app.js

sacr3dc0w:
app.js currently includes `blog.js` code on the `blog_post` page class, rather than `blog-post.js` code. This fixes that.


junedkazi:
@sacr3dc0w can we rebase this PR please ?

Issue #1114, opened 2017-11-02
It's hard to track submission of the contact form


Tiggerito:
### Expected behavior
When someone submits the contact form it should be possible to track that submission in systems like Google Analytics. Even better if we can track the funnel flow of a submission attempt to analyse what issues people have.

### Actual behavior
The contact form is a single page powered by JavaScript. There is no simple way to detect if the form has been submitted.

### Steps to reproduce behavior
Submit a contact form.

### Possible solution 1
Provide some ways to hook into events that happen while submitting a form. e.g. submission errors, captcha failures and then final success. 

### Possible solution 2
I've developed some JavaScript to detect changes in the page and trigger Google Analytics events on submissions and displaying errors. 

To make it work I need to edit some of the theme files by adding ids to some elements. It would be a lot nicer if I did not need to edit the themes. My notes:

The contact-us-page and contact-us-success ids need adding to templates/pages/contact-us.html. e.g.

<div **id="contact-us-page"** class="page-content page-content--centered">
    {{#if forms.contact.success}}
            <div **id="contact-us-success"**>{{{lang 'forms.contact_us.successful' shopPath=urls.home}}}</div>
    {{ else}}
    <p>{{{page.content}}}</p>
    {{> components / page / contact - us - form}}
        {{/if}}
</div>

contact-us-error needs adding to templates/components/contact-us-form.html. Dynamic validation avoids this, but the captcha triggers it

{{ #if forms.contact.error}}
<div **id="contact-us-error"**>{{> components / common / alert - error forms.contact.error}}</div>
{{/if}}

Ubersmake:
Thank you, @Tiggerito ! We've merged #1317.

PR #1317, opened 2018-07-30
Add div and id attributes so that contact form steps can be tracked

Tiggerito:
#### What?

This is my first ever submission like this on GitHub. Please tell me if I've got it all wrong.

The contact form is single page, which means it is very hard for people to track form submissions on systems like Google Analytics. There is no success page to track it as a goal.

I've developed a way to hook into the page and detect user activity via JavaScript, this JavaScript can directly send a signal to Analytics so that it can track submissions. 

But my solution, and probably anyone elses requires some minor edits to the theme.

The code needs to detect if certain elements are present. These elements indicate if the page has been submitted, or if the submission had errors (also useful to track). However the theme does not include an easy way to find those elements, and there are not always clean elements to detect.

The edits add id attributes to some div elements and add some div elements with id attributes. This makes it possible to detect if the page is a contact page and if it currently contains a success or error messages. 

With that, people can write JavaScript to detect what is going on in the page. And most importantly, detect if the form was successfully submitted. With Google Analytics, events or fake page views can be sent which can be used to trigger a contact form sent goal.

#### Tickets / Documentation

I raised the issue a while ago. 

https://github.com/bigcommerce/cornerstone/issues/1114

And decided to have a go at making the fix myself.

I could not work out how to raise this change under the issue.

Nothing is visible.

### DIFF
--- a/templates/components/page/contact-us-form.html
+++ b/templates/components/page/contact-us-form.html
@@ -2,7 +2,7 @@
     {{inject 'contactEmail' (lang 'forms.validate.contact.email_address')}}
     {{inject 'contactQuestion' (lang 'forms.validate.contact.question')}}
     {{#if forms.contact.error}}
-        {{> components/common/alert-error forms.contact.error}}
+    <div id="contact-us-error">{{> components/common/alert-error forms.contact.error}}</div>
     {{/if}}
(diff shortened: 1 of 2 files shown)


bookernath:
Makes sense to me! CC @mjschock

PR #1319, opened 2018-07-31
show dots only if carousel has more than one slide

bc-zoharmuzafi:
#### What?
Currently the carousel on the homepage having "dots" even if the carousel has only one slide. This PR is fixing this issue and dots will appear only if there are multiple slides.

#### Tickets / Documentation
- [Slick Docs](http://kenwheeler.github.io/slick/)

##### before
screen shot 2018-07-31 at 11 27 29 am

##### after
screen shot 2018-07-31 at 11 29 01 am

screen shot 2018-07-31 at 12 02 10 pm


jasonherbst8:
Do the dots still render if there are multiple slides?
Please verify that case, and if so, thumbs up from me.

PR #1326, opened 2018-08-08
change src for invoice logo image

bc-zoharmuzafi:
#### What?
Logo img in the invoice is missing. This change will add logo img to the invoice

##### Before
screen shot 2018-08-08 at 10 23 24 am

###### Original logo size
screen shot 2018-08-10 at 5 08 16 pm

###### Optimized
screen shot 2018-08-10 at 5 09 02 pm

###### Specified Dimensions   
screen shot 2018-08-10 at 5 10 02 pm

### DIFF
--- a/templates/pages/account/orders/invoice.html
+++ b/templates/pages/account/orders/invoice.html
@@ -6,7 +6,13 @@
     <div class="Invoice">
         <h1>
             {{#if settings.store_logo.image}}
-                <img class="header-logo-image lazyload" data-sizes="auto" src="{{cdn 'img/loading.svg'}}" data-src="{{getImage settings.store_logo.image 'logo_size'}}" alt="{{settings.store_logo.title}}" title="{{settings.store_logo.title}}">
+                {{#if theme_settings.logo_size '===' 'original'}}
+                    <img class="header-logo-image-unknown-size" src="{{getImage settings.store_logo.image 'logo_size'}}" alt="{{settings.store_logo.title}}" title="{{settings.store_logo.title}}">
+                {{else}}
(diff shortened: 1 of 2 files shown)


templates/pages/account/orders/invoice.html
```diff
@@ -6,7 +6,7 @@
     <div class="Invoice">
         <h1>
             {{#if settings.store_logo.image}}
-                <img class="header-logo-image lazyload" data-sizes="auto" src="{{cdn 'img/loading.svg'}}" data-src="{{getImage settings.store_logo.image 'logo_size'}}" alt="{{settings.store_logo.title}}" title="{{settings.store_logo.title}}">
+                <img class="header-logo-image lazyload" data-sizes="auto" src="{{getImage settings.store_logo.image}}" data-src="{{getImage settings.store_logo.image 'logo_size'}}" alt="{{settings.store_logo.title}}" title="{{settings.store_logo.title}}">
```
junedkazi: So this is a lazy load image which will load the loading image first & then replace it will original image. Do we see anything in the console as to why the image is not loading ? Or else we completely remove lazy loading. See https://github.com/bigcommerce/cornerstone/blob/81d24b58633365ea1974e9d3e633111d66cc6d57/templates/components/common/store-logo.html

bc-zoharmuzafi: I looked into that and find out that this issue happens because of the printing (the image is not ready by the time the printing windows opened.
I delete the lazy load and used the same method as referenced, however, the code in the reference use lazy load as well so I should modify it according to the requirements.

Ubersmake:
💚

PR #1334, opened 2018-08-22
Require Webpack config only when used

jbruni:
#### What?

Detailed description has been written and discussed at issue https://github.com/bigcommerce/stencil-cli/issues/379

This change helps to prevent that issue to happen, because the time to be [ready for receiving messages from `stencil-cli`](https://github.com/bigcommerce/cornerstone/blob/2.3.2/stencil.conf.js#L54-L67) is dramatically reduced.

#### Tickets / Documentation

- [bigcommerce/stencil-cli GitHub issue 379](https://github.com/bigcommerce/stencil-cli/issues/379)

### DIFF
--- a/stencil.conf.js
+++ b/stencil.conf.js
@@ -1,6 +1,4 @@
 var webpack = require('webpack');
-var devConfig = require('./webpack.dev.js');
-var prodConfig = require('./webpack.prod.js');
 
 /**
  * Watch options for the core watcher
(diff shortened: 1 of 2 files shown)


stencil.conf.js
```diff
@@ -26,6 +22,9 @@ var watchOptions = {
  * Watch any custom files and trigger a rebuild
  */
 function development() {
+    var webpack = require('webpack');
```
junedkazi: @jbruni can we just move the `var webpack = require('webpack');` outside so we don't have to specify it for each env.

junedkazi:
@jbruni can you also add a changelog entry please.

jbruni:
@junedkazi - Requested changes have been done. :+1:

PR #1337, opened 2018-08-27
fix(storefront): STRF-5282 Unable to add products to cart with 2 or more non-required File Upload options using Safari

bc-NikitaPuzanenko:
#### What?
Safari only does one iteration in the current`for of` loop. This change will create a new FormData object and append valid entries using a standard for loop instead.

Before:
before

After:
after

### DIFF
--- a/assets/js/theme/common/product-details.js
+++ b/assets/js/theme/common/product-details.js
@@ -60,20 +60,25 @@ export default class ProductDetails {
     /**
      * https://stackoverflow.com/questions/49672992/ajax-request-fails-when-sending-formdata-including-empty-file-input-in-safari
      * Safari browser with jquery 3.3.1 has an issue uploading empty file parameters. This function removes any empty files from the form params
-     * @param formData: FormData object
+     * @param form: form NodeList
      * @returns FormData object
      */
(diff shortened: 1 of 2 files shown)


assets/js/theme/common/product-details.js
```diff
@@ -60,20 +60,24 @@ export default class ProductDetails {
...
      */
-    filterEmptyFilesFromForm(formData) {
+    filterEmptyFilesFromForm(form) {
         try {
-            for (const [key, val] of formData) {
-                if (val instanceof File && !val.name && !val.size) {
-                    formData.delete(key);
+            const filteredFormData = new FormData();
+            for (let i = 0; i < form.elements.length; i++) {
+                if (form.elements[i].type !== 'file' && form.elements[i].value && form.elements[i].name) {
```
lord2800: This can be cleaned up a bit:
```javascript
let element = form.elements[i];
if (element.type === 'file' && element.size > 0) {
    filteredFormData.append(element.name, element.files[0], element.files[0].name);
} else if (element.name && element.value) {
    filteredFormData.append(element.name, element.value);
}
```

PR #1338, opened 2018-08-28
Update qty

bc-zoharmuzafi:
#### What?
Currently if user change manually (write a number) the qty of item, the cart is not updated and the only way to update the qty it is using the inc and dec arrows. This change added "focusout" event on the qty field, which allow the user to change manually the number of the qty., and the cart will be updated once focus leaves the field. 

#### Test
screen shot 2018-09-12 at 4 51 21 pm

### DIFF
--- a/assets/js/theme/cart.js
+++ b/assets/js/theme/cart.js
@@ -18,29 +18,15 @@ export default class Cart extends PageManager {
         this.bindEvents();
     }
 
-    cartUpdate($target, preVal = null) {
+    cartUpdate($target) {
         const itemId = $target.data('cartItemid');
         const $el = $(`#qty-${itemId}`);
-        const oldQty = preVal !== null ? preVal : parseInt($el.val(), 10);
+        const oldQty = parseInt($el.val(), 10);
         const maxQty = parseInt($el.data('quantityMax'), 10);
         const minQty = parseInt($el.data('quantityMin'), 10);
         const minError = $el.data('quantityMinError');
         const maxError = $el.data('quantityMaxError');
-        let qty = null;
-
-        switch ($target.data('action')) {
-        case 'inc':
-            qty = oldQty + 1;
-            break;
-        case 'dec':
-            qty = oldQty - 1;
-            break;
-        default:
-            qty = $el.attr('value');
-        }
-
-        const newQty = qty;
-
+        const newQty = $target.data('action') === 'inc' ? oldQty + 1 : oldQty - 1;
         // Does not quality for min/max quantity
         if (newQty < minQty) {
             return swal({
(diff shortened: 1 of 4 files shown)


templates/components/cart/content.html
```diff
@@ -110,6 +110,9 @@ <h4 class="cart-item-name"><a href="{{url}}">{{name}}</a></h4>
                                data-quantity-max-error="{{lang 'products.quantity_max' quantity=max_purchase_quantity}}"
                                min="1"
                                pattern="[0-9]*"
+                               data-cart-itemid="{{id}}"
```
junedkazi: why do we have this twice ?

assets/js/theme/cart.js
```diff
@@ -18,15 +18,28 @@ export default class Cart extends PageManager {
         this.bindEvents();
     }
 
-    cartUpdate($target) {
+    cartUpdate($target, preVal = null) {
         const itemId = $target.data('cartItemid');
         const $el = $(`#qty-${itemId}`);
-        const oldQty = parseInt($el.val(), 10);
+        const oldQty = preVal !== null ? preVal : parseInt($el.val(), 10);
```
junedkazi: shouldn't we be running `parseInt ` on `preVal` as well ?

assets/js/theme/cart.js
```diff
@@ -18,15 +18,28 @@ export default class Cart extends PageManager {
...
         const maxQty = parseInt($el.data('quantityMax'), 10);
         const minQty = parseInt($el.data('quantityMin'), 10);
         const minError = $el.data('quantityMinError');
         const maxError = $el.data('quantityMaxError');
-        const newQty = $target.data('action') === 'inc' ? oldQty + 1 : oldQty - 1;
+        let qty = null;
```
junedkazi: why not initialize the `qty` value with the default value.

assets/js/theme/cart.js
```diff
@@ -18,18 +18,34 @@ export default class Cart extends PageManager {
...
+        const newQty = qty;
 
         // Does not quality for min/max quantity
-        if (newQty < minQty) {
+        if (isNaN(newQty)) {
+            return swal({
+                text: "Not a valid entry",
+                type: 'error',
+            });
+        } else if(newQty < minQty) {
```
bc-zoharmuzafi: @junedkazi I changed that to display error if there is an invalid input, however that will change the current logic that fallback in case of invalid input to the min min_purchase_quantity.  If we choose to keep it, I would like to write a test for this case

junedkazi:
@Ubersmake can we test this PR before we merge it.

Ubersmake:
@junedkazi @bc-zoharmuzafi Taking a look.

Ubersmake:
There are a lot of corner cases here.

Using the current arrow-only behavior for comparison:

- Reducing a line item's quantity to 0 will remove that line item from the cart.
- If that line item was the only line item in the cart, the cart will be refreshed with a "Your cart is empty" message.

With the addition of the item quantity feature:

- Reducing the only/last/remaining line item's quantity to 0 will remove that line item from the cart, but will show a cart page without line items, as opposed to a "Your cart is empty" message.
- Using a negative quantity will cause an error modal to appear. The form value of the line item remains as it was before the error, but the display value is the negative quantity. This display should be updated, similar to what happens with non-integer inputs.

assets/js/theme/cart.js
Ubersmake: I like this logic because I (and users) can now enter trash into the quantity field. Maybe the text should be "Not a valid number?" Because other errors will show up in the case of out of stock.

Ubersmake:
💚

Issue #1292, opened 2018-06-28
Home page carousel data is pre-HTML escaped


carsonreinke:
### Expected behavior
Stencil object [Home Slide Show](https://stencil.bigcommerce.com/docs/home-slide-show-object) should return `heading`, `text`, and `button_text` as unescaped raw data.  This allows actual template to decide if data should be HTML entity escaped or not.

### Actual behavior
Stencil object is returned with data already HTML escaped.  Use of Handlebars `{{ }}` produces double escaped HTML and `{{{ }}}` produces single HTML escaped output.

### Steps to reproduce behavior
Adding any HTML entities to Home Page Carousel content within the BigCommerce Admin under `Storefront > Home Page Carousel > Slide 1 > Button Text`.  Current template file uses a mix of Handlebars escaping: https://github.com/bigcommerce/cornerstone/blob/master/templates/components/carousel-content.html

junedkazi:
@carsonreinke can you open a PR to address the issue. I will be happy to review it.

carsonreinke:
@junedkazi Unfortunately, this is external to what can be resolved within Cornerstone.  The actual Stencil object has this problem.  Support suggested I post it here.

NjTheCoder:
@carsonreinke I am also seeing this issue

Ubersmake:
Thanks all for raising this. I've written up an engineering ticket to follow up.

satyasaadhak:
Was this ever resolved? I am facing this same issue right now. I want to insert some spans in the carousel slide content, but it just shows up as it is isn't of being rendered as a DOM element.

PR #1341, opened 2018-09-05
Countries with symbols & punctuation in the name are not encoding

jake-bc:
This is fixing the country name options entities on the account signup form.
This fix is for Cornerstone / Other themes will need to be updated by 3rd parties such as Pixel Union
https://jira.bigcommerce.com/browse/STRF-5396

Countries with symbols & punctuation in the name are not encoding, for example:
`"Cote d&#039;Ivoire"`

**updated theme file template/components/common/forms/select.html**

**-Line 11 Was:** <option {{#if selected}} selected {{/if}} value="{{value}}">{{label}}</option>
screen shot 2018-09-05 at 12 49 06 pm

**-Line 11 Now Is:** <option {{#if selected}} selected {{/if}} value="{{value}}">{{{label}}}</option>
screen shot 2018-09-05 at 12 47 02 pm

@bigcommerce/storefront-team

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
-## Draft
+## Draft 
+- Fix encoding issues on Account Signup Form ("&#039;" characters showing in country name)[#1341] (https://github.com/bigcommerce/cornerstone/pull/1341)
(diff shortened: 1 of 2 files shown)


carsonreinke:
Related to #1292, data is already HTML encoded.

PR #1345, opened 2018-09-14
Fix eslint warning in cart.js focus event from #1338

sacr3dc0w:
Fix missing function expression name eslint warning from #1338

### DIFF
--- a/assets/js/theme/cart.js
+++ b/assets/js/theme/cart.js
@@ -227,7 +227,7 @@ export default class Cart extends PageManager {
         });
 
         // cart qty manually updates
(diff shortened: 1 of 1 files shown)


assets/js/theme/cart.js
```diff
@@ -227,7 +227,7 @@ export default class Cart extends PageManager {
         });
 
         // cart qty manually updates
-        $('.cart-item-qty-input', this.$cartContent).on('focus', function () {
+        $('.cart-item-qty-input', this.$cartContent).on('focus', () => {
```
jbruni: #1418

Issue #1339, opened 2018-09-03
Mapping Custom Templates to JavaScript Modules


sacr3dc0w:
https://stencil.bigcommerce.com/docs/js-101#map-custom

Not really an issue, I wanted to provide an alternative solution to the one in the documentation. After `pageClasses` is declared in `app.js`, add something along the lines of the code below.

```
const customClasses = {
    'pages/custom/category/custom-category': () => import('./theme/custom/custom-category'),
    'pages/custom/page/custom-page': () => import('./theme/custom/custom-page'),
    'pages/custom/product/custom-product': () => import('./theme/custom/custom-product'),
};
```

After `const context` is declared in `window.stencilBootstrap`, paste the follow:

```
const template = context.template;
const templateCheck = Object.keys(customClasses).indexOf(template);
```

Finally, we need to actually load the module if `templateCheck` is greater than `-1`. Paste the following code after `if (typeof pageClassImporter === 'function')`

```
if (templateCheck > -1) {
    // Find the appropriate page loader based on template
    const customClassImporter = customClasses[template];
    if (typeof customClassImporter === 'function') {
        const CustomClass = (await customClassImporter()).default;
        CustomClass.load(context);
    }
}
```
----------------------------

What `app.js` should look like when done.

```
__webpack_public_path__ = window.__webpack_public_path__; // eslint-disable-line

import 'babel-polyfill';
import $ from 'jquery';
import 'jquery-migrate';
import Global from './theme/global';

const getAccount = () => import('./theme/account');
const getLogin = () => import('./theme/auth');
const pageClasses = {
    account_orderstatus: getAccount,
    account_order: getAccount,
    account_addressbook: getAccount,
    shippingaddressform: getAccount,
    account_new_return: getAccount,
    'add-wishlist': () => import('./theme/wishlist'),
    account_recentitems: getAccount,
    account_downloaditem: getAccount,
    editaccount: getAccount,
    account_inbox: getAccount,
    account_saved_return: getAccount,
    account_returns: getAccount,
    login: getLogin,
    createaccount_thanks: getLogin,
    createaccount: getLogin,
    getnewpassword: getLogin,
    forgotpassword: getLogin,
    blog: () => import('./theme/blog'),
    blog_post: () => import('./theme/blog-post'),
    brand: () => import('./theme/brand'),
    brands: () => import('./theme/brands'),
    cart: () => import('./theme/cart'),
    category: () => import('./theme/category'),
[...]

PR #1346, opened 2018-09-14
Blueprint for Mapping Custom Templates to JavaScript Modules

sacr3dc0w:
Here's the blueprint for more easily mapping custom templates to javascript modules (#1339) in `app.js`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Draft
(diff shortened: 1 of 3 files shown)


junedkazi:
@sacr3dc0w I was looking at this PR in more detail. So this is definitely an issue for older themes which were forked off of cornerstone before Jan 2017 because then the mapping was between `template file => js file` but then we changed the mapping to `page type => js file` so it shouldn't matter if you load custom template it should load the right JS file. So this should not be an issue on latest cornerstone release which make me wonder if we should merge this or add this information to BigCommerce documentation.

PR #1349, opened 2018-09-18
(WIP) card edit

jonathandelefortrie:
#### What?

Billing address form attached to the payment method.

#### Tickets

- [Ticket](https://jira.bigcommerce.com/browse/PAYMENTS-3478)
- [Design](https://www.figma.com/file/XUicUp9urJ7JGMdjDAi8ySgG/Card-vaulting?node-id=0%3A4144)

Coming soon...


config.json
```diff
@@ -302,7 +302,1261 @@
       "discover",
       "mastercard",
       "visa"
-    ]
+    ],
```
jonathandelefortrie: Hey @junedkazi , as we decided to create the input fields manually for the card management feature, I created three new attributes on the config.json to add the default countries, states and form validations. I would like to know if this the right place for storing those values? Thanks

junedkazi: why do we want to hard code the counties and states list ? Can't we make a call to bcapp and fetch the information ?

aleachjr: I agree, I don't think we should be hardcoding this list. //cc @precariouspanther @zackangelo @chrisboulton another possibly opportunity to use the new service that we have been discussing.

precariouspanther: Yes please, we've had several occasions where we have had to update these lists - it'd be preferable to avoid fragmenting this. The other concern is these values need to be matched to names within our lists which change (rather than the ISO codes) so there is a risk that changes to the official list would be a backwards breaking change for this that we wouldn't realise.
Please consider driving this from either @aleachjr 's geo service or BCApp's current definition (Which will be migrated)

jonathandelefortrie:
Hey @junedkazi @bc-zoharmuzafi , that PR is now ready for review. if you can have a look that would be awesome. Thanks!

lang/en.json
```diff
@@ -357,12 +357,14 @@
         },
         "payment_methods": {
             "heading": "Payment Methods",
+            "payment_method": "Payment method",
+            "billing_address": "Billing address",
             "card_ending_in": "ending in {last_four}",
             "card_expiry": "{month}/{year}",
             "new_payment_method": "Add new payment method",
             "no_methods": "You currently have no payment methods added to your account",
-            "name_on_card": "Name on card",
```
junedkazi: So we no longer be showing name on the card ?

jonathandelefortrie: Hey @junedkazi , yes we realized the ```name on card``` was not something we keep when vaulting the card. Hey @jackosaurus , please correct me if I'm wrong.

jackosaurus: That's correct, we do not store "Name on card" (we were mistaken)

PR #1335, opened 2018-08-25
 Added "activePage" as a active class in navigation menu 

taruntds:
activePage is a active class which will be applied when any page is active. For example. If you visit any category page then this "activePage" class will be activated and that particular category name will be highlighted in different color.

PR #1354, opened 2018-09-21
taruntds's "Added "activePage" as a active class in navigation menu and web pages"

Ubersmake:
#### What?

A new PR for #1335 

> activePage is a active class which will be applied when any page is active. For example. If you visit any category page then this "activePage" class will be activated and that particular category name will be highlighted in different color.

The active page has the same styling as a navigation link on hover.

#### Tickets / Documentation

#1335 

##### Category Page (Bath):

Before:
screen shot 2018-09-21 at 1 15 08 pm

After:
screen shot 2018-09-21 at 1 17 05 pm

##### Narrow Category Page (Publications):

Before:
screen shot 2018-09-21 at 1 36 45 pm

After:
screen shot 2018-09-21 at 1 33 46 pm

##### Nested Category Page (Shop All > Bath):

Before:
screen shot 2018-09-21 at 1 21 35 pm

After:
screen shot 2018-09-21 at 1 22 07 pm

##### Narrow Nested Category Page (Shop All > Bath):

Before:
screen shot 2018-09-21 at 1 36 37 pm

After:
screen shot 2018-09-21 at 1 33 39 pm

### DIFF
--- a/templates/components/common/navigation-menu.html
+++ b/templates/components/common/navigation-menu.html
@@ -16,7 +16,7 @@
         {{#unless theme_settings.hide_content_navigation}}
             {{#each pages}}
                  <li class="navPages-item navPages-item-page">
-                     <a class="navPages-action" href="{{url}}">{{name}}</a>
+                     <a class="navPages-action{{#if name '==' ../page.title}} activePage{{/if}}" href="{{url}}">{{name}}</a>
                  </li>
              {{/each}}
         {{/unless}}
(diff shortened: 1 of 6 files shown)


templates/components/common/navigation-menu.html
```diff
@@ -16,7 +16,7 @@
         {{#unless theme_settings.hide_content_navigation}}
             {{#each pages}}
                  <li class="navPages-item navPages-item-page">
-                     <a class="navPages-action" href="{{url}}">{{name}}</a>
+                     <a class="navPages-action{{#if name '==' ../page.title}} activePage{{/if}}" href="{{url}}">{{name}}</a>
```
junedkazi: why this check ?

Ubersmake: This is for the parent-page scenario. If a page is nested in a tree this will highlight both the current page and its parent, like in: https://user-images.githubusercontent.com/1546172/45905025-c491ab80-bda3-11e8-8302-fab27866a954.png

PR #1362, opened 2018-09-26
Fix errors in breadcrumb schema markup

sacr3dc0w:
#### What?

In late August, Google updated how they processed breadcrumbs, causing schema markup to begin throwing errors when you run a site through their structured data testing tool.

### _**Notice:**_
**While this PR resolves those errors, it will require `{{url}}` being made available for the last [Breadcrumbs Object](https://stencil.bigcommerce.com/docs/breadcrumbs-object)** I'm not sure what that would take on your guys end, but here's the code that will help cross the finish line once that's made available.

#### Screenshots

**Before**
1 before

**With the fix, not quite fully resolved.**
2 almost

**With the fix and a URL plugged into the meta tag, fully resolving the errors thrown.**
3 done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Draft
+- Fix broken breadcrumb schema markup [#1362](https://github.com/bigcommerce/cornerstone/pull/1362)
 
 ## 2.5.0 (2018-09-26)
 - Blueprint for Mapping Custom Templates to JavaScript Modules [#1346](https://github.com/bigcommerce/cornerstone/pull/1346)
--- a/templates/components/common/breadcrumbs.html
+++ b/templates/components/common/breadcrumbs.html
@@ -2,8 +2,9 @@
     {{#each breadcrumbs}}
         <li class="breadcrumb {{#if @last}}is-active{{/if}}" itemprop="itemListElement" itemscope itemtype="http://schema.org/ListItem">
             {{#if url}}
-                <a href="{{url}}" class="breadcrumb-label" itemprop="item">{{name}}</a>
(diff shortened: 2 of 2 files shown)


bc-zoharmuzafi:
LGTM.

Tiggerito:
The first position should be 1 (however things still work with a zero):

https://developers.google.com/search/docs/data-types/breadcrumb
"The position of the breadcrumb in the breadcrumb trail. Position 1 signifies the beginning of the trail."

https://schema.org/BreadcrumbList
"The specific values of 'position' are not assigned meaning for a BreadcrumbList, but they should be integers, e.g. beginning with '1' for the first item in the list."

The inclusion of the current page at the end of the breadcrumb is optional with regard to Google and getting breadcrumb rich snippets. However, the schema spec indicates it is "typically" included:

https://schema.org/BreadcrumbList
"A BreadcrumbList is an ItemList consisting of a chain of linked Web pages, typically described using at least their URL and their name, and typically ending with the current page."

I personally don't bother to include it as it is a bit of a fiddle to do. The url parameter would be great for this and other issues I have.

In the above final example the last name should be the product name. I presume that was just a copy paste thing just for the example.

To be honest, I remove all structured data added by the BC themes and add it using our own json-ld code. It's easier to do that than hunt through the microdata making fixes. It would be nice to have an option to switch off all structured data added by a theme.

carsonreinke:
@Tiggerito I second the JSON-LD use.  Maybe an option to switch between Microdata, JSON-LD, or none.

Tiggerito:
On issue I encounter at times is that third party tools like review systems may use json-ld or microdata for their markup. If that is in a different syntax to what the theme uses, it is not possible to link the entities being defined (In Google at the moment). e.g. the product entity does not link to its review entities. Breaking a major reason for structured data, which is to define how entities relate to each other.

PR #1371, opened 2018-10-12
MERC-4196: Fixes issue with Slick slider for mobile safari

pedelman:
#### What?

Loading Cornerstone inside an iframe on mobile safari will cause your web browser to become unresponsive. 

There look to be multiple issues at play, the first being some css changes as suggested in the linked ticket. I have also tracked down the setting which is causing another issue, it looks like something in this [block of code](https://github.com/kenwheeler/slick/blob/master/slick/slick.js#L595) is breaking on mobile safari in an iframe. 

You can reproduce this issue in isolation using the following codepen example.

Using the iPad emulator in xcode, open safari and navigate to the following page.

- https://codepen.io/pingoo/pen/BRVPxQ

Compare results against desktop web browser.

#### Tickets / Documentation

There is an open issue on Slick, the open source slider we are using.

- [kenwheeler/slick#2834](https://github.com/kenwheeler/slick/issues/2834)

#### Screenshots

Before:
screen shot 2018-10-11 at 8 30 06 pm

After:
screen shot 2018-10-11 at 8 30 44 pm

### DIFF
--- a/templates/components/products/carousel.html
+++ b/templates/components/products/carousel.html
@@ -3,31 +3,8 @@
         "dots": true,
         "infinite": false,
         "mobileFirst": true,
-        "slidesToShow": 2,
-        "slidesToScroll": 2,
-        "responsive": [
-            {
-                "breakpoint": 1260,
-                "settings": {
-                    "slidesToScroll": 3,
-                    "slidesToShow": {{ columns }}
-                }
-            },
-            {
-                "breakpoint": 800,
-                "settings": {
-                    "slidesToScroll": 3,
-                    "slidesToShow": 5
-                }
-            },
-            {
-                "breakpoint": 550,
-                "settings": {
-                    "slidesToScroll": 3,
-                    "slidesToShow": 3
-                }
-            }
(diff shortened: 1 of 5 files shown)


templates/components/products/carousel.html
```diff
@@ -3,31 +3,8 @@
...
-            },
-            {
-                "breakpoint": 550,
-                "settings": {
-                    "slidesToScroll": 3,
-                    "slidesToShow": 3
-                }
-            }
-        ]
+        "slidesToShow": {{ columns }},
```
pedelman: Maybe we should consider hard-coding this to 2 or 3 to give better backward compatibility? @bigcommerce/storefront-team @bigcommerce/merc-team

BC-EChristensen: That was my thought as well, just go with 3 to keep thing as consistent as possible (though I don't really see why full page scrolls would be undesirable)

pedelman:
@Ubersmake updated to include the change log ♻️

PR #1373, opened 2018-10-24
card add

jonathandelefortrie:
#### What?

Add payment method.
- Adding new icon ``` lock.svg ```.
- Unify the CSS classes for the add and edit form.
- New template for edit form with an injection of global variables into context.
- Extended the account.js with a new method to initiate credit card and billing address fields validations, mapping form fields name and value to body request (bigpay).
- Added a payment-method.js file for credit card validations and helpers with unit test.
- Various fixes: blacklist key in payment methods list, remove the statement in edit form for bigpay token.

#### Tickets

- [Ticket](https://jira.bigcommerce.com/browse/PAYMENTS-2580)

Desktop
screen shot 2018-10-26 at 2 21 57 pm

Mobile
screen shot 2018-10-26 at 2 22 28 pm


jonathandelefortrie:
Hey @junedkazi , this is code the for ```add card``` form. The form uses ajax to store the credit card, so it requires some extra javascript validation which I have added in a standalone file. Can you have a look at the PR and let know if the implementation is okay? Thank you

assets/js/theme/account.js
```diff
@@ -201,6 +208,128 @@ export default class Account extends PageManager {
...
+                        const refObj = obj;
+                        // Map dynamic fields to api payload
+                        switch (item.name) {
+                        case 'FormField[2][4]':
+                            refObj.first_name = item.value;
+                            break;
+                        case 'FormField[2][5]':
+                            refObj.last_name = item.value;
+                            break;
+                        case 'FormField[2][6]':
```
danieldelcore: I would really like to avoid doing this if we can help it. 
Is there a way that we can either: 
- Build the form manually 
- Get the backend to send proper `names` over the wire so we can avoid or simplify this mapping?

🍹

package.json
```diff
@@ -8,6 +8,7 @@
   "dependencies": {
     "@bigcommerce/stencil-utils": "^1.1.2",
     "babel-polyfill": "^6.26.0",
+    "creditcards": "^3.0.1",
```
junedkazi: No package lock file please.

assets/js/theme/account.js
junedkazi: I agree we need a better way of building these switch cases instead of relying on numeric keys.

jonathandelefortrie: Hey @junedkazi @danieldelcore , the best alternative will be to create those input fields manually so I have a better control of the form names.

assets/js/theme/account.js
```diff
@@ -201,6 +208,103 @@ export default class Account extends PageManager {
         });
     }
 
+    initPaymentMethodFormValidation($paymentMethodForm) {
+        // Inject validations into form fields before validation runs
+        $paymentMethodForm.find('#first_name.form-field').attr('data-validation', `{ "type": "singleline", "label": "${this.context.firstNameLabel}", "required": true, "maxlength": 0 }`);
```
jonathandelefortrie: Hey @junedkazi , we have removed the dynamic input fields for the static components. however, I had to inject the validation through javascript because it was breaking the template engine.

PR #1379, opened 2018-11-07
STRF-5640 - Load cart quantity in FE

bookernath:
#### What?

This is the first step in improving storefront performance by moving most/all Cart interactions to the frontend.

This change moves all functionality related to calculating the cart quantity in the header to FE API calls, after the initial HTML loads.

In future PRs, we'll be able to actually turn off the loading of Cart details in the HTML in Cornerstone, which will improve performance, especially for large carts. However, this PR does *not* disable Cart data, it simply introduces the FE behavior.

The "last known" cart quantity is cached in `localStorage` so it loads quickly when the page loads and then it's quietly updated if new information is received from the API. There is a graceful fallback if `localStorage` is unavailable in the environment or otherwise fails (e.g. if the 5MB storage limit is already reached for the domain).

#### Tickets / Documentation

- [STRF-5640](https://jira.bigcommerce.com/browse/STRF-5640)

### DIFF
--- a/assets/js/theme/global.js
+++ b/assets/js/theme/global.js
@@ -24,11 +24,11 @@ export default class Global extends PageManager {
         window.lazySizesConfig = window.lazySizesConfig || {};
         window.lazySizesConfig.loadMode = 1;
 
+        cartPreview(this.context.secureBaseUrl);
         quickSearch();
         currencySelector();
         foundation($(document));
         quickView(this.context);
-        cartPreview();
         compareProducts(this.context.urls);
         carousel();
         menu();
--- a/assets/js/theme/global/cart-preview.js
+++ b/assets/js/theme/global/cart-preview.js
@@ -8,13 +8,15 @@ export const CartPreviewEvents = {
     open: 'opened.fndtn.dropdown',
 };
 
-export default function () {
+export default function (secureBaseUrl) {
     const loadingClass = 'is-loading';
     const $cart = $('[data-cart-preview]');
     const $cartDropdown = $('#cart-preview-dropdown');
     const $cartLoading = $('<div class="loadingOverlay"></div>');
 
-    $('body').on('cart-quantity-update', (event, quantity) => {
+    const $body = $('body');
+
+    $body.on('cart-quantity-update', (event, quantity) => {
         $('.cart-quantity')
             .text(quantity)
             .toggleClass('countPill--positive', quantity > 0);
@@ -48,5 +50,33 @@ export default function () {
             $cartLoading
                 .hide();
         });
+    });
+
+    let quantity = 0;
+
+    // Get existing quantity from localStorage if found
+    if (utils.tools.storage.localStorageAvailable()) {
+        if (localStorage.getItem('cart-quantity')) {
+            quantity = Number(localStorage.getItem('cart-quantity'));
+            $body.trigger('cart-quantity-update', quantity);
+        }
+    }
+
+    // Get updated cart quantity from the Cart API
+    const cartQtyPromise = new Promise((resolve, reject) => {
+        utils.api.cart.getCartQuantity({ baseUrl: secureBaseUrl }, (err, qty) => {
+            if (err) {
+                reject(err);
+            }
+            resolve(qty);
+        });
+    });
+
+    // If the Cart API gives us a different quantity number, update it
+    cartQtyPromise.then(qty => {
+        $body.trigger('cart-quantity-update', qty);
+        if (utils.tools.storage.localStorageAvailable()) {
+            localStorage.setItem('cart-quantity', qty);
+        }
     });
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - shotaK's Add context to the menu collapsible factory target elements [#1382](https://github.com/bigcommerce/cornerstone/pull/1382)
 - Added default rule for product carousel card title to break words on overflow. [#1389](https://github.com/bigcommerce/cornerstone/pull/1389)
 - Only show cookie privacy notice for EU IP addresses [#1381](https://github.com/bigcommerce/cornerstone/pull/1381)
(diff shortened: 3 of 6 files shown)


assets/js/theme/global/cart-preview.js
```diff
@@ -49,4 +51,53 @@ export default function () {
                 .hide();
         });
     });
+
+    // Check if we can use localStorage
+    function storageAvailable(type) {
```
mattolson: Would prefer to move this out to the top level of this file, or even to `assets/js/theme/common` since we will probably want this for other purposes later.

assets/js/theme/global/cart-preview.js
```diff
@@ -49,4 +51,53 @@ export default function () {
...
+                e.name === 'NS_ERROR_DOM_QUOTA_REACHED') &&
+                // acknowledge QuotaExceededError only if there's something already stored
+                storage.length !== 0;
+        }
+    }
+
+    let quantity = 0;
+
+    // Get existing quantity from localStorage if found
+    if (storageAvailable('localStorage')) {
```
mattolson: Would be useful to have a helper function

```
function localStorageAvailable() {
    return storageAvailable('localStorage');
}
```

assets/js/theme/global/cart-preview.js
```diff
@@ -49,4 +51,53 @@ export default function () {
...
+        if (localStorage.getItem('cart-quantity')) {
+            quantity = Number(localStorage.getItem('cart-quantity'));
+            $body.trigger('cart-quantity-update', quantity);
+        }
+    }
+
+    // Get updated cart quantity from the Cart API
+    const cartQtyPromise = new Promise(resolve => {
```
mattolson: I just noticed that the `getCartQuantity` function in stencil-utils does not handle api errors properly.

https://github.com/bigcommerce/stencil-utils/blob/master/src/api/cart.js#L26 should be modified to check for an error and then call `callback(err)`. When successful, it should call `callback(null, quantity)`. The standard for JS callback functions is that the first param the error and the second param is the result.

Once that is done, this PR should be modified to handle the error and call the reject function in the Promise.

assets/js/theme/global.js
```diff
@@ -24,11 +24,11 @@ export default class Global extends PageManager {
         window.lazySizesConfig = window.lazySizesConfig || {};
         window.lazySizesConfig.loadMode = 1;
 
+        cartPreview();
```
bookernath: If we can load the cart qty from "cache" I want to do that ASAP, so making this first.

The network activity this kicks off is async.

assets/js/theme/global/cart-preview.js
bookernath: https://github.com/bigcommerce/stencil-utils/pull/88

assets/js/theme/common/local-storage-check.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * Check if localStorage is available for use
+ * https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API#Testing_for_availability
+ * @param type
+ * @returns boolean
+ */
+export function storageAvailable(type) {
```
junedkazi: I think this belongs in stencil utils so it makes it easier for us to make changes in the future.

bookernath:
@mattolson ♻️ 

Please note that this requires stencil-utils 3.0.0 (containing https://github.com/bigcommerce/stencil-utils/pull/88) to be released to NPM before we can merge this.

assets/js/theme/common/local-storage-check.js
mattolson: I think since we are just starting to dabble in local storage and we don't yet know exactly what kind of interface we'd want into it, it's totally fine to start out just in cornerstone and then if we have a need for it elsewhere, extract it

mattolson:
Don't we also want `cart: false` here? https://github.com/bigcommerce/cornerstone/blob/master/config.json#L326

assets/js/theme/common/local-storage-check.js
mattolson: On the other hand, @junedkazi you make a good point that putting things in stencil-utils makes it easier for those that have forked Cornerstone in private themes to keep things up to date. @bookernath how hard would it be to add this to stencil-utils instead?

bookernath:
Now depends on https://github.com/bigcommerce/stencil-utils/pull/90 and I've targeted this PR at 3.1.0

bookernath:
@mattolson ♻️ once more :)

bookernath:
This can't be merged yet; found some weird behavior on HTTP/shared SSL storefronts.

bookernath:
Now depends on https://github.com/bigcommerce/stencil-utils/pull/92 and another stencil-utils bump; I've assumed this will be 3.1.1 for purposes of package.json.

bookernath:
Depends on https://github.com/bigcommerce/stencil-utils/pull/93

carsonreinke:
For your consideration https://github.com/bigcommerce/cornerstone/issues/1405#issuecomment-446663721

carsonreinke:
@bookernath 

> @carsonreinke can you give an example of what you mean?

GTM is our biggest use case.  Delaying pushing data for the data layer changes what trigger executes the tag.  The tag cannot fire on page load, DOM load, or window loaded.  A custom event would need to be added to trigger once the API call came back.

> Just to make sure I'm articulating this clearly, we don't plan to do anything with the existing {{cart}} resource - we intend to keep supporting it and all the existing themes and stores that might be using it.
> 
> However, towards the goal of making the storefront faster, we plan to move the interactions to the frontend and disable the cart object by default in Cornerstone which will result in a faster response time for the page, especially for stores with certain configurations like a lot of discount rules, or shoppers who have a lot of items in their cart.
> 
> Anyone who still wants to use the Cart resource can keep it enabled.

Oh I see, so it is opt-in.  Well, that kind of solves my problem, so maybe this is just a non-issue.

PR #1380, opened 2018-11-08
STRF-4612 - Remove unnecessary API call to get cookie notification st…

bookernath:
…atus

#### What?

You can already infer the status of the cookie setting from the presence of cookie notification text; this property returns false if the setting is disabled. So this API call was unnecessary. It should probably also be deprecated and removed from stencil-utils.

This handlebars check is enough: https://github.com/bigcommerce/cornerstone/blob/master/templates/layout/base.html#L35

#### Tickets / Documentation

- [STRF-4612](https://jira.bigcommerce.com/browse/STRF-4612)

#### Dependencies

Depends on https://github.com/bigcommerce/stencil-utils/pull/87 and a stencil-utils bump

#### Video

http://recordit.co/U3qUIDCHl4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Draft
+- Remove unnecessary API call to get cookie notification status [#1380](https://github.com/bigcommerce/cornerstone/pull/1380)
 
 ## 2.6.0 (2018-11-05)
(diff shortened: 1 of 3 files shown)


bookernath:
I need to figure out how to stop the alert from firing, which is driven by a constructor in stencil-utils.

bookernath:
Depends on https://github.com/bigcommerce/stencil-utils/pull/87 and a stencil-utils bump

mattolson:
I have published stencil-utils 2.0, which you should bring into this PR.

mattolson:
wow

PR #1381, opened 2018-11-09
STRF-4612 only show cookie privacy notice for EU IP addresses

bookernath:
#### What?

Depends on https://github.com/bigcommerce/bigcommerce/pull/27154 in which we surface a new key in the context indicating if the shopper has an EU IP address.

This will only show the cookie notification for those shoppers, so that you don't bug shoppers when you aren't legally required to.

Leaving this in the theme code allows the behavior to be further customized by a theme developer if desired.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Remove extra font only used for textual store logo. [#1375](https://github.com/bigcommerce/cornerstone/pull/1375)
 - shotaK's Add context to the menu collapsible factory target elements [#1382](https://github.com/bigcommerce/cornerstone/pull/1382)
 - Added default rule for product carousel card title to break words on overflow. [#1389](https://github.com/bigcommerce/cornerstone/pull/1389)
+- Only show cookie privacy notice for EU IP addresses [#1381](https://github.com/bigcommerce/cornerstone/pull/1381)
 
 ## 2.6.0 (2018-11-05)
 - Add support for Card Management: List, Delete, Edit, Add and Default Payment Method [#1376](https://github.com/bigcommerce/cornerstone/pull/1376)
--- a/templates/layout/base.html
+++ b/templates/layout/base.html
@@ -32,9 +32,9 @@
         {{{snippet 'header'}}}
         <svg data-src="{{cdn 'img/icon-sprite.svg'}}" class="icons-svg-sprite"></svg>
 
-        {{#if settings.privacy_cookie}}
+        {{#and settings.privacy_cookie settings.is_eu_ip_address}}
             {{> components/common/cookie}}
-        {{/if}}
+        {{/and}}
 
         {{> components/common/header }}
         {{> components/common/body }}


carsonreinke:
@bookernath Can't you be a someone from the EU and not be in the EU?  I feel like it is less confusing, either you work with the a country requiring this or you don't.  If you do want to add this, maybe make it a bit more flexible for other filtering in the future asking for consent.

bookernath:
@carsonreinke thanks for your feedback.

The [EU cookie law](https://www.cookielaw.org/the-cookie-law/) (to the best of my understanding) primarily applies to 1. EU-owned websites and 2. websites targeted at EU citizens. So when we think of the best *default* behavior of our platform (of which Cornerstone is a representation) it makes sense to only bug shoppers when absolutely necessary - when they are browsing from the EU.

An EU-owned or EU-targeted business would be best advised to modify this behavior to target a broader range of shoppers, to your point.

I had considered a future enhancement in which this would be a toggle-able setting in theme editor - what do you think about that?

carsonreinke:
@bookernath Just because you are a citizen, does not mean you are in the physical location.  A EU targeted website should not try and figure who is a citizen or not, it just needs to be on.

I could not find anything on people actually doing this.

carsonreinke:
@xenph You might be right.  I guess if there some precedence for this somewhere, please send that over.  It just sounds like the wrong path.

xenph:
Hi, I'm Christopher Beckett, BigCommerce's Data Privacy Leader :smile: If you want a further view of EDPB's recent guidance regarding the territorial scope of the GDPR, please read this: https://edpb.europa.eu/sites/edpb/files/files/file1/edpb_guidelines_3_2018_territorial_scope_en.pdf. However, typically I read these so you don't have to 😄

flair-duncan:
@bookernath As you're already doing a GeoIP lookup to generate `settings.is_eu_ip_address` could you expose additional data in a new location object, i.e. continent, country and city?

Could then do something like `{{#if settings.location.continent '===' 'eu'}}` and store owners/developers would benefit from the additional location data available without having to rely on third party services...

🤔 Just an idea 🙄

bookernath:
@flair-duncan I think that's a great idea; due to some stuff happening on the backend it would be a little bit more of a project than this change (we have an existing method to calculate "is in the EU"), but I'm willing to consider it for the future. We're kicking it around internally now.

junedkazi:
@bookernath are we waiting on anything to get this merged ? Also the PR will need a rebase.

bookernath:
@junedkazi we're waiting for https://github.com/bigcommerce/bigcommerce/pull/27154 to hit production.

junedkazi:
I think it is fine to merge it bcoz we don't plan on releasing cornerstone until we have all the backend changes rolled out completely.

bookernath:
Sure, let's do it!

bookernath:
FYI @flair-duncan we're rolling out a new `country_code` property you'll be able to access, which will include the country-level GeoIP result. It should be on all stores in a few days.

PR #1377, opened 2018-11-02
PROJECT-2272: Add support for declarative data tag analytics

lord2800:
#### What?

Add support for declarative data tag analytics events.

#### Tickets / Documentation

- CP-3026
- CP-3027
- CP-3028
- CP-3033
- CP-3055
- CP-3056
- CP-3181
- CP-3273
- CP-3275
- CP-3872
- CP-3936
- CP-3957
- CP-3960
- CP-3971
- CP-4036
- CP-4038
- CP-4042

Probably others as well.

@Ubersmake @bigcommerce/cp-dt @bigcommerce/storefront-team

PR #1384, opened 2018-11-16
changed access level for parent context- products list display

bc-zoharmuzafi:
#### What?
In Cornerstone v2.6.0 compare buttons are missing from Category pages that use a List view instead of a grid.  This pr change the access level to the parent context when generating the products list display. 

###### Before
screen shot 2018-11-16 at 3 24 11 pm

###### After
screen shot 2018-11-16 at 3 24 28 pm

### DIFF
--- a/templates/components/products/list.html
+++ b/templates/components/products/list.html
@@ -1,11 +1,7 @@
 <ul class="productList">
     {{#each products}}
     <li class="product">
-        {{#if settings.data_tag_enabled}}
-            {{>components/products/list-item   show_compare=../show_compare show_rating=../settings.show_product_rating theme_settings=../theme_settings customer=../customer event=../event position=(add @index 1)}}
-        {{else}}
-            {{>components/products/list-item show_compare=../show_compare show_rating=../settings.show_product_rating theme_settings=../theme_settings customer=../customer}}
-        {{/if}}
+		{{>components/products/list-item show_compare=../show_compare show_rating=../settings.show_product_rating theme_settings=../theme_settings customer=../customer event=../event settings=../settings position=(add @index 1)}}
     </li>
     {{/each}}
 </ul>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Draft
+- Fixed compare buttons for product list display [#1384](https://github.com/bigcommerce/cornerstone/pull/1384)
 - Remove unnecessary API call to get cookie notification status [#1380](https://github.com/bigcommerce/cornerstone/pull/1380)
 - Cart switch from quote item hash to id which is immutable [#1387](https://github.com/bigcommerce/cornerstone/pull/1387)
 - Remove extra font only used for textual store logo. [#1375](https://github.com/bigcommerce/cornerstone/pull/1375)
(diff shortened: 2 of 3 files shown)


templates/components/products/list.html
```diff
@@ -2,9 +2,9 @@
     {{#each products}}
     <li class="product">
         {{#if settings.data_tag_enabled}}
-            {{>components/products/list-item   show_compare=../show_compare show_rating=../settings.show_product_rating theme_settings=../theme_settings customer=../customer event=../event position=(add @index 1)}}
+            {{>components/products/list-item   show_compare=../../show_compare show_rating=../settings.show_product_rating theme_settings=../theme_settings customer=../customer event=../event position=(add @index 1)}}
```
mattolson: Can you explain this change? Why is it required for `show_compare` but not for others like `show_rating`?

templates/components/products/list.html
```diff
@@ -2,9 +2,9 @@
     {{#each products}}
     <li class="product">
         {{#if settings.data_tag_enabled}}
-            {{>components/products/list-item   show_compare=../show_compare show_rating=../settings.show_product_rating theme_settings=../theme_settings customer=../customer event=../event position=(add @index 1)}}
+            {{>components/products/list-item show_compare=../../show_compare show_rating=../../settings.show_product_rating theme_settings=../../theme_settings customer=../../customer event=../../event position=(add @index 1)}}
```
mattolson: Can you elaborate on this change? Why do we need to do it here and not in any of the other templates?

templates/components/products/list.html
bc-zoharmuzafi: @mattolson Entering an if block adds another level to the context, so the parent becomes ../../foo 
Looking on the grid display, I do see that the other values also will be available in the same context as show_compare so they will also should to be modified. I also tested @index and this variable is available and no change needed for it.

templates/components/products/list.html
bc-zoharmuzafi: @mattolson Entering into an if statement adds another level to the context, so the parent becomes ../../foo
Looking on the grid display, I do see that the other values also will be available in the same context as show_compare so they will also should to be modified. I also tested @index and this variable is available and no change needed for it.

mattolson: If that is the case, then many other things would be broken too. This if statement was added in https://github.com/bigcommerce/cornerstone/pull/1377 which added it in many other places as well. Are those templates experiencing issues too?

bc-zoharmuzafi: @mattolson I looked on pr 1377. The cases that I can see that we used reference to parent context inside if statements are in the following:
https://github.com/bigcommerce/cornerstone/pull/1377/files#diff-a6bce6406ef20c5a779702dccc5173b7R2
https://github.com/bigcommerce/cornerstone/pull/1377/files#diff-53b336f69d0af6f9498a97945e1a5f60R248
https://github.com/bigcommerce/cornerstone/pull/1377/files#diff-f8e0227af0d7c280e2ac9ffdc6f746d8R27
https://github.com/bigcommerce/cornerstone/pull/1377/files#diff-135cdd50e87e5a25fb7b2a86965ae978R45

I will take a look into the each case to see if we did refer to the parent context correctly.

mattolson:
@lord2800 can you take a look at this since it is a modification of some work you did earlier?

PR #1389, opened 2018-11-30
STRF-5651: Break product card titles on overflow.

Ubersmake:
#### What?

Changes to the New Products carousel in 2.5.2 changed how many products appeared as the view width became narrower and narrower. Specifically, it would always try to honor the number of columns set for the theme. This means that products with long words get less and less space as the view narrows, and particularly long words start overflowing into adjacent columns.

This change forces words to break if they hit the edge of the column.

This has also been verified to *not crash* Mobile Safari, which was the reason for #1371 in the first place.

#### Tickets / Documentation

- [STRF-5651](https://jira.bigcommerce.com/browse/STRF-5651)
- #1371 

##### Before:

before

##### After:

after


EdFromBJCS:
Is it possible to make this hyphenate word breaks?

PR #1397, opened 2018-12-06
Coming soon product image loading issue

nnandani:
#### What?
Coming soon product image loading issue on the compare page. 

#### Tickets / Documentation
There is an issue on the compare page. If the product as coming soon image it will just load loading SVG:
Before: http://prntscr.com/lrln56

After solution: http://prntscr.com/lrlnso


templates/pages/compare.html
```diff
@@ -14,7 +14,7 @@ <h1 class="page-heading">{{lang 'compare.header' products=comparisons.length}}</
                     <article class="card">
                         <figure class="card-figure">
                             <div class="card-img-container">
-                                <img class="card-image lazyload" data-sizes="auto" src="{{cdn 'img/loading.svg'}}" data-src="{{getImage image 'product_size' (cdn theme_settings.default_image_product)}}" alt="{{image.alt}}" title="{{image.alt}}">
+                                <img class="card-image lazyload" data-sizes="auto" src="{{cdn 'img/loading.svg'}}" data-src="{{getImage image 'product_size' (cdn 'img/ProductDefault.gif')}}" alt="{{image.alt}}" title="{{image.alt}}">
```
junedkazi: @nnandani instead of hardcoding the file path here. Can we update `default_image_product` in `config.json` to use the right path.

nnandani: Default image product path not working that's why I have added this path for solution.

Ubersmake:
Hi @nnandani . Thank you bringing this to our attention. We ended up merging a separate fix (#1438) which changed paths in the theme's `config.json` file, but we wouldn't have found it if this PR was never opened.

PR #1399, opened 2018-12-07
To set proper align Product thumbnail image slider arrows

NjTheCoder:
#### What?
If a product has more then 5 images. It will create slider but there is an issue about align of the slider arrows. 

**Before:** 
image

**After:**
image

### DIFF
--- a/assets/scss/components/vendor/slick/_slick.scss
+++ b/assets/scss/components/vendor/slick/_slick.scss
@@ -22,7 +22,7 @@
     z-index: 1;
     border: 1px solid $slick-arrow-borderColor;
     height: remCalc(61px);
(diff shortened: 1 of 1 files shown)


assets/scss/components/vendor/slick/_slick.scss
```diff
@@ -22,7 +22,7 @@
     z-index: 1;
     border: 1px solid $slick-arrow-borderColor;
     height: remCalc(61px);
-    margin-top: -(remCalc(30px));
+    margin-top: -(remCalc(15px));
```
NjTheCoder: This line for to fix arrow issue on product thumbnail images.

PR #1400, opened 2018-12-08
STRF-5749: Resolves price_ranges showing the same values in min & max…

jake-bc:
…. This is an update to price-range.html

#### What?
Adds if statements to check if min & max price ranges are the same. This was an issue that was found when customer group discounts are applied which end up setting the min & max values to be the same.

#### Tickets / Documentation

https://jira.bigcommerce.com/browse/STRF-5749


jake-bc:
This PR was actually probably good to go. I just need to double check customer groups on Monday.

PR #1402, opened 2018-12-11
(WIP) Don't fire cart API request if there's no cart ID

bookernath:
#### What?

If there's no cart, there's no need to check the quantity. So don't fire the API request if there's no cart_id in the context*

*Coming soon


bookernath:
Depends on https://github.com/bigcommerce/bigcommerce/pull/28078

PR #1403, opened 2018-12-11
breadcrumbs last element

bc-zoharmuzafi:
#### What?
Fixed the position for breadcrumb elements to start with 1. instead of 0.

###### Before
screen shot 2018-12-11 at 3 12 51 pm

###### After
screen shot 2018-12-19 at 2 05 35 pm

### DIFF
--- a/templates/components/common/breadcrumbs.html
+++ b/templates/components/common/breadcrumbs.html
@@ -1,13 +1,13 @@
 <ul class="breadcrumbs" itemscope itemtype="http://schema.org/BreadcrumbList">
     {{#each breadcrumbs}}
         <li class="breadcrumb {{#if @last}}is-active{{/if}}" itemprop="itemListElement" itemscope itemtype="http://schema.org/ListItem">
-            {{#if url}}
-                <a href="{{url}}" class="breadcrumb-label" itemprop="item"><span itemprop="name">{{name}}</span></a>
-            {{else}}
+            {{#or @last (if url "==" null)}}
                 <meta itemprop="item" content="{{url}}">
                 <span class="breadcrumb-label" itemprop="name">{{name}}</span>
-            {{/if}}
-            <meta itemprop="position" content="{{@index}}" />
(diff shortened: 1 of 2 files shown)


Tiggerito:
Googles docs on Breadcrumbs is not very clear on if you should include the current page in the breadcrumb:

https://developers.google.com/search/docs/data-types/breadcrumb

I personally don't include it, and it makes no difference to the rich snippets produced. Google does not include the current page as part of the trail.

Google does specify that the position should start with 1, not zero. It maybe worth fixing that, but it causes no harm.

Their example also does not include the home page. Again, I've found that including it or not makes no difference to the breadcrumbs. Google does not include the root as part of the trail. I include it for BC sites, just because the code is simpler.

On a related note, if say the product url is added as the id for the last item, then that often clashes with the id used for the Product entity on the same page. This can cause the product entity to be merged into the breadcrumb entity. One common method to avoid this id clashing is to add a # value to the end of the ids. e.g.

"@id": "https://.../product-page#Breadcrumb"

and

"@id": "https://.../product-page#Product" 

Then multiple entities can be on the same page and still have their own id based on the URI for the page.

templates/components/common/breadcrumbs.html
```diff
@@ -1,13 +1,13 @@
 <ul class="breadcrumbs" itemscope itemtype="http://schema.org/BreadcrumbList">
     {{#each breadcrumbs}}
         <li class="breadcrumb {{#if @last}}is-active{{/if}}" itemprop="itemListElement" itemscope itemtype="http://schema.org/ListItem">
-            {{#if url}}
-                <a href="{{url}}" class="breadcrumb-label" itemprop="item"><span itemprop="name">{{name}}</span></a>
-            {{else}}
+            {{#if @last}}
```
mattolson: I would change this to 
```
if (last || url.empty) {
  don't create an anchor tag
} else {
  create an anchor tag
}
```

PR #1406, opened 2018-12-13
defer loading of addthis

bc-zoharmuzafi:
#### What?

defer loading of addthis for faster loading pages.

##### Before
screen shot 2018-12-12 at 9 21 39 pm

##### After
screen shot 2018-12-12 at 9 22 54 pm

### DIFF
--- a/templates/components/common/share.html
+++ b/templates/components/common/share.html
@@ -61,11 +61,13 @@
             {{/each}}
         </ul>
         <script type="text/javascript"
-                src="//s7.addthis.com/js/300/addthis_widget.js#pubid=ra-4e94ed470ee51e32"></script>
-        <script>
-            if (typeof(addthis) === "object") {
-                addthis.toolbox('.addthis_toolbox');
-            }
+                defer src="//s7.addthis.com/js/300/addthis_widget.js#pubid=ra-4e94ed470ee51e32"></script>
+        <script>  
+            window.addEventListener('DOMContentLoaded', function() {
+                if (typeof(addthis) === "object") {
+                    addthis.toolbox('.addthis_toolbox');
+                }
+            });
         </script>
     </div>
 {{/if}}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Draft
+- Added defer tag to addThis and defered execution of related script[#1406](https://github.com/bigcommerce/cornerstone/pull/1406)
 - Fixed compare buttons for product list display [#1384](https://github.com/bigcommerce/cornerstone/pull/1384)
 - Remove unnecessary API call to get cookie notification status [#1380](https://github.com/bigcommerce/cornerstone/pull/1380)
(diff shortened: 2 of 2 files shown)


templates/components/common/share.html
```diff
@@ -61,11 +61,7 @@
...
         </ul>
         <script type="text/javascript"
-                src="//s7.addthis.com/js/300/addthis_widget.js#pubid=ra-4e94ed470ee51e32"></script>
-        <script>
-            if (typeof(addthis) === "object") {
-                addthis.toolbox('.addthis_toolbox');
-            }
-        </script>
+                defer src="//s7.addthis.com/js/300/addthis_widget.js#pubid=ra-4e94ed470ee51e32"></script>
+        <script type="text/javascript" defer src="/assets/js/theme/global/addthis.js"></script>
```
mattolson: Two things:
(1) If doing this, we should pass this path through the `cdn` helper to make sure we aren't hitting the site directly, but using the cdn.
(2) This will result in a second request which will probably hurt our time to interactive timing.

Take a look at https://www.addthis.com/academy/using-addthis-asynchronously/ for other techniques we can use.

templates/components/common/share.html
```diff
@@ -61,11 +61,13 @@
             {{/each}}
         </ul>
         <script type="text/javascript"
-                src="//s7.addthis.com/js/300/addthis_widget.js#pubid=ra-4e94ed470ee51e32"></script>
+                defer src="//s7.addthis.com/js/300/addthis_widget.js#pubid=ra-4e94ed470ee51e32"></script>
         <script>
-            if (typeof(addthis) === "object") {
-                addthis.toolbox('.addthis_toolbox');
-            }
+            window.addEventListener('DOMContentLoaded', function() {
```
mattolson: You should use https://api.jquery.com/load-event/ here to ensure that the external script has been downloaded and parsed.

templates/components/common/share.html
```diff
@@ -61,11 +61,13 @@
...
         </ul>
         <script type="text/javascript"
-                src="//s7.addthis.com/js/300/addthis_widget.js#pubid=ra-4e94ed470ee51e32"></script>
-        <script>
-            if (typeof(addthis) === "object") {
-                addthis.toolbox('.addthis_toolbox');
-            }
+                defer src="//s7.addthis.com/js/300/addthis_widget.js#pubid=ra-4e94ed470ee51e32"></script>
+        <script>  
+            window.onload = function() {
```
mattolson: This is not a good approach because it will wipe out any other onload handlers that were created. You should just jquery to attach an on load handler instead.

mattolson: Actually, turns out jQuery is not available at this time, because we load it as part of the bundle which is inserted at the end of the body (after this). So we need to use vanilla javascript here. `document.addEventListener('DOMContentLoaded', handler);`

mattolson:
@bc-zoharmuzafi 👍 LGTM but can you add a CHANGELOG entry?

PR #1407, opened 2018-12-13
Remove deprecated features - Delivery/Event Date, Configurable Fields

bookernath:
#### What?

These fields were deprecated over 6 months ago from our UIs; removing them from Cornerstone.

Product Options can be used to do all of the same things.


junedkazi:
Have we migrated all existing stores using cornerstone to not use these fields anymore ?

EdFromBJCS:
Configurable fields are absolutely still in my UI. Should I expect that to disappear?

If I were using base Cornerstone or if I had blindly applied this change, dozens of my product pages would have lost their configurable fields without warning and I would be faced with hours of work digging through hundreds of products to figure out which ones needed configurable fields and creating options for them.

Surely I'm not the only one who relies on configurable fields in my store. Has anyone warned BC customers what's in store for them if they update Cornerstone?

bookernath:
HI @EdFromBJCS , thanks for reaching out.

We bundled this change into a major version (3.0.0) since this is a breaking change from a semantic versioning perspective.

In terms of announcements, we've announced this on our [developer docs](https://developer.bigcommerce.com/changelog) and we'll also be discussing this deprecation (and the improvements in Cornerstone 3.0) in our January town hall.

Our support team is equipped with a tool that will move your configurable fields to product options for you if you'd like some assistance!

EdFromBJCS:
Also, product options do not do all the same things, because you can include a product with a configurable field in a Product Pick List option, but you cannot include a product with its own options in a Product Pick List option. On my store, this means I cannot replace configurable fields with product options.

bookernath:
@EdFromBJCS thank you for that additional context.

For what it's worth, Configurable Fields have been marked as deprecated for a while now, and won't be supported in future versions of our control panel either.

That being said, we certainly wouldn't want to alienate you or your business, so I'd like to see if there are alternative configurations of your catalog that would serve the same purpose.

Do you mind dropping us a line at community@bigcommerce.com and we can set some time to discuss it further? Thanks!

EdFromBJCS:
@bookernath email has been sent. Also included some more details of my use case for context.

PR #1415, opened 2018-12-27
Bump stencil-utils to 4.0.1

bookernath:
#### What?

There's a bug in stencil-utils that will be fixed by https://github.com/bigcommerce/stencil-utils/pull/95

Once it's released as 4.0.1, this PR is for the bump to that version, to address a bug in which the cart quantity update would not work on HTTP storefronts.

ping @Ubersmake

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "author": "BigCommerce",
   "license": "MIT",
   "dependencies": {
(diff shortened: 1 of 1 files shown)


bookernath:
Note that tests will fail until 4.0.1 is released.

PR #1418, opened 2019-01-11
Fix cart item quantity change rollback

jbruni:
#### What?

This is a bug fix.

Context: change the quantity of a cart line item by manually editing the quantity input field value.

Currently, when a manual quantity change fails to succeed (for example, because it is greater than the maximum purchasable quantity), the quantity is reverted to **zero** in the input field.

It should rollback to the **previous value**. Not to zero.

This PR fixes this bug.

<hr>

It is quite simple to see what is the problem in the code:
- inside the quantity input `focus` event handler function, **`this`** is used as a reference to the event target element (jQuery binds the event target element to the event handler function context);
- as an **arrow function** is being used for the event handler function, `this` is **not** referencing the event target element, as it should, but the enclosing context (Cart class instance)

Thus, the fix was also quite simple: only a matter of using a normal function. This way, we have the event target element properly bound to the function's `this` context. The `preVal` (previous value) variable is properly populated with the input's value.


karen-white:
Thanks @jbruni ! We'll get this reviewed.

mattolson:
@jbruni Can you add a changelog entry, and then I'll merge this.

PR #1423, opened 2019-01-16
MERC-4456 Add a page content region

bc-jcha:
#### What?

Add a region called page_content

#### Tickets / Documentation

- [Related ticket](https://jira.bigcommerce.com/browse/MERC-4456)

### DIFF
--- a/templates/pages/page.html
+++ b/templates/pages/page.html
@@ -20,7 +20,7 @@
         {{#replace '%%Syndicate%%' page.content}}
             {{> components/page/rss_feed}}
         {{else}}
-            <p>{{{page.content}}}</p>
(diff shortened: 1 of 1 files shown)


BC-EChristensen:
While I agree that ideally we'd have a way to edit these pages with older themes, I'm still not sure I fully understand the alternative. It seems sort of strange to me to have an exception to the way regions work in this one particular case. Sure, I imagine we *could* have storefront know to treat `page.content` as *also* having an invisible region below it, named `page_content`, I'm just not sure I love that solution. Possibly I'm misunderstanding the idea?

bc-jcha:
Closing this as we are discussing another approach in the long term. We are creating a private theme file for MERC team to work on with this approach for now.

PR #1431, opened 2019-01-23
STRF-4173 - Ensure SKU and UPC display correctly for Variants when th…

bc-jz:
…ey are not set on base product.

#### What?

For complex products, if you do not set SKU or UPC on the base product level, then the SKU/UPC will not display on the PDP and QuickView popup on PLP.  These changes correct this issue ensuring that variant level SKU/UPC can always display on PDP and PLP.  

The SKU/UPC label values will also now hide if there is not a SKU associated with the particular selection.

#### Tickets / Documentation
 [JIRA](https://jira.bigcommerce.com/browse/STRF-4173)

*On my branch*
Initial page load - there is not a SKU or UPC on base product:

Choosing selections that have a SKU and UPC set on Variant:

### DIFF
--- a/assets/js/theme/common/product-details.js
+++ b/assets/js/theme/common/product-details.js
@@ -216,8 +216,14 @@ export default class ProductDetails {
                 $container: $('.form-field--stock', $scope),
                 $input: $('[data-product-stock]', $scope),
             },
-            $sku: $('[data-product-sku]'),
-            $upc: $('[data-product-upc]'),
+            sku: {
+                $label: $('dt.sku-label', $scope),
+                $value: $('[data-product-sku]', $scope),
+            },
+            upc: {
+                $label: $('dt.upc-label', $scope),
+                $value: $('[data-product-upc]', $scope),
+            },
             quantity: {
                 $text: $('.incrementTotal', $scope),
                 $input: $('[name=qty\\[\\]]', $scope),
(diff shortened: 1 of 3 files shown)


assets/js/theme/common/product-details.js
```diff
@@ -216,8 +216,14 @@ export default class ProductDetails {
                 $container: $('.form-field--stock', $scope),
                 $input: $('[data-product-stock]', $scope),
             },
-            $sku: $('[data-product-sku]'),
-            $upc: $('[data-product-upc]'),
+            sku: {
+                $dt: $('dt.sku-label', $scope),
+                $dd: $('[data-product-sku]', $scope),
```
junedkazi: can we just name the keys as `label` and `value` so it is easier to read.

assets/js/theme/common/product-details.js
```diff
@@ -216,8 +216,14 @@ export default class ProductDetails {
...
+            },
+            upc: {
+                $dt: $('dt.upc-label', $scope),
+                $dd: $('[data-product-upc]', $scope),
```
junedkazi: same here

assets/js/theme/common/product-details.js
bc-jz: yea I suppose so - was just following the format already in place

PR #1434, opened 2019-01-25
fix(storefront): STRF-5948 Cleanup and XSS fix on Cart page.

Ubersmake:
#### What?

Certain product options allow for user-submitted input, which in the case of JS can be executed on the Cart page. This value is now escaped properly on the Cart page. This is also addressed on the Returns page.

This also cleans up code related to Configurable Fields, which were removed in https://github.com/bigcommerce/cornerstone/pull/1407.

This does not address Custom Fields, which are entirely determined by the merchant.

#### Tickets / Documentation

- [STRF-5948](https://jira.bigcommerce.com/browse/STRF-5948)


bc-vandana:
LGTM 👍

PR #1437, opened 2019-01-31
Bump stencil utils

junedkazi:
#### What?
Bump stencil utils

@bigcommerce/storefront-team


bc-jz:
💚 - This is looking good to me. I don't see that `search.js` being used but good to make the change there too just in case.

PR #1441, opened 2019-02-07
added package-lock.json

bc-zoharmuzafi:
#### What?
Added package-lock.json and checked to github, in order to have reproducible builds.


README.md
```diff
@@ -102,6 +102,11 @@ and run:
 npm install
 ```
 
+* Note: package-lock.json file was generated by Node version 8 and npm version 6.4.1. Although the app supports Node versions 6 and 8 as well as multiple versions of npm, we should always use those versions when installing the package-lock.json, unless decided to upgrade those, and in this case the readme should be upadted as well. If using a different version for node OR npm, please delete the package-lock.json file prior installing the node packages and also prior pushing to github.
```
mattolson: * "updating package-lock.json" instead of "installing the package-lock.json"
* "unless it is decided"
* typo: "upadted"
* "prior to installing node packages" instead of "prior installing the node packages"
* "prior to pushing to github"

README.md
```diff
@@ -102,6 +102,11 @@ and run:
...
+
+If updating or adding a dependency, please double check that you are working on Node version 8 and npm version 6.4.1 and run ```npm update <package_name>```  or ```npm install <package_name>``` (avoid running npm install for updating a package). After updating the package, please make sure that the changes in the package-lock.json reflecting only the updated/new package prior pushing the changes to github.
```
mattolson: * "reflect only" instead of "reflecting only"
* "prior to pushing the changes"

README.md
```diff
@@ -102,6 +102,12 @@ and run:
 npm install
 ```
 
+* Note: package-lock.json file was generated by Node version 8 and npm version 6.4.1. Although the app supports Node versions 6 and 8 as well as multiple versions of npm, we should always use those versions when updating package-lock.json, * unless it is decided to upgrade those, and in this case the readme should be updated as well. If using a different version for node OR npm, please delete the package-lock.json file prior to installing node packages and also prior to pushing to github.
```
mattolson: Text makes much more sense now. Just one final formatting note. I would remove the list here since this is not really a list, and also remove the `*` before `unless it is decided`.

PR #1444, opened 2019-02-13
MERC-4566 Add option to hide breadcrumbs and page title

kchu93:
#### What?
Add option to hide breadcrumbs, page heading, category heading, contact page heading and blog page heading. 

#### Tickets / Documentation
[MERC-4566](https://jira.bigcommerce.com/browse/MERC-4566)

**Turning off breadcrumbs**
feb-14-2019 12-31-38

**Turn off page heading**
feb-14-2019 12-32-33

**Turn off category page heading**
feb-14-2019 12-34-04

**Turn off contact page heading**
contact

**Turn off blog page heading**
blog

#### Images

screen shot 2019-02-14 at 4 10 13 pm

### DIFF
--- a/templates/components/common/breadcrumbs.html
+++ b/templates/components/common/breadcrumbs.html
@@ -1,13 +1,15 @@
 <ul class="breadcrumbs" itemscope itemtype="http://schema.org/BreadcrumbList">
-    {{#each breadcrumbs}}
-        <li class="breadcrumb {{#if @last}}is-active{{/if}}" itemprop="itemListElement" itemscope itemtype="http://schema.org/ListItem">
-            {{#or @last (if url "==" null)}}
-                <meta itemprop="item" content="{{url}}">
-                <span class="breadcrumb-label" itemprop="name">{{name}}</span>
-            {{else}}
-                <a href="{{url}}" class="breadcrumb-label" itemprop="item"><span itemprop="name">{{name}}</span></a>
-            {{/or}}
-            <meta itemprop="position" content="{{add @index 1}}" />
-        </li>
-    {{/each}}
+    {{#unless theme_settings.hide_breadcrumbs }}
+        {{#each breadcrumbs}}
+            <li class="breadcrumb {{#if @last}}is-active{{/if}}" itemprop="itemListElement" itemscope itemtype="http://schema.org/ListItem">
+                {{#or @last (if url "==" null)}}
+                    <meta itemprop="item" content="{{url}}">
+                    <span class="breadcrumb-label" itemprop="name">{{name}}</span>
+                {{else}}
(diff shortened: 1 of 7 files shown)


templates/components/common/breadcrumbs.html
```diff
@@ -1,5 +1,6 @@
 <ul class="breadcrumbs" itemscope itemtype="http://schema.org/BreadcrumbList">
     {{#each breadcrumbs}}
+        {{#or ../customer (if ../theme_settings.breadcrumbs '!==' true)}}
```
junedkazi: why do we need the customer ?

junedkazi: Can we push the `li` one level deep.

templates/pages/category.html
```diff
@@ -20,7 +20,9 @@
 {{#if category.image}}
     <img class="lazyload" data-sizes="auto" src="{{cdn 'img/loading.svg'}}" data-src="{{getImage category.image 'zoom_size'}}">
 {{/if}}
-<h1 class="page-heading">{{category.name}}</h1>
+{{#or ../customer (if ../theme_settings.page_title '!==' true)}}
```
junedkazi: Same here why customer ?

Tiggerito:
I think the terminology used here is confusing. Technically these are not page titles, but headings. In admin they are generally called page names.

And the Search Engine Optimisation sections use the term page title (correctly) for the title tag, which is a different thing.

I suggest you use "Hide page heading".

kchu93:
After speaking with design, we want to be very clear with which page headings we want to turn off. I have added individual options (type 'page' heading, category heading, contact us heading, and blog heading)

mattcoy-arcticleaf:
The names (id's) of the settings themselves are a bit confusing. I would think that a setting with an id of "breadcrumbs" would enable them if set to true, rather than the opposite being true, which is the case here. I would recommend changing the names to "hide_breadcrumbs", "hide_page_heading", etc.

From a code perspective, since you are only checking for a single item in the conditional statements, you do not need to use the "or" helper. A simple {{#if ../theme_settings.blog_page_heading '!==' true}}, would suffice. Also, since you are checking for an inverse, this can be written with even less code using an {{#unless}} helper. Finally, since the scope has not changed in these files (except in the case of the breadcrumbs where you put it inside the each loop), you do not need to add the '../', leaving the final code something like: {{#unless theme_settings.blog_page_heading }}code here{{/unless}}. Note that in the case of the breadcrumbs, you would need to use the ../, or move the conditional outside the each block.

kchu93:
@mattcoy-arcticleaf Thank you so much for the feedback, this was super helpful!

PR #1446, opened 2019-02-14
fix(cp): CP-4140 fixes modal scroll issue from quickview button

sacr3dc0w:
File missed in #1445


Ubersmake:
Thanks @sacr3dc0w !

PR #1449, opened 2019-02-19
MERC-4567 Add a region for page content

bc-jcha:
#### What?

Add support for placing widgets in the "page" template.

#### Why?

We will support placing widgets in the page template of all the themes going forward by implicitly injecting regions where {{{page.content}}} is placed, but for now, manual placement of "region" will help the project progress.

#### Tickets / Documentation

- [MERC-4567](https://jira.bigcommerce.com/browse/MERC-4567)


bc-jcha:
Closing this per team's discussion. We will create a stencil bundle that contains this change instead.

Issue #1416, opened 2019-01-09
**Sample** facet shows HTML as text


shotaK:
### Expected behavior
**Sample** facet filter should not show HTML as text

### Actual behavior
**Sample** facet filter should shows HTML as text
![screen shot 2019-01-09 at 12 58 59 pm](https://user-images.githubusercontent.com/3370367/50888222-9b5a6100-140e-11e9-9e31-39973c51d0e3.png)

### Steps to reproduce behavior
In BigCommerce store admin go to **Products -> Product Filtering** and activate **Sample** filter.
Check any category page **Sample** facet filter.

Ubersmake:
We've created an internal engineering ticket for this. While #1426 does fix the immediate display issue in the facet filters, there are also issues surrounding the "Refine By" text as well as the parameters that appear in the URL.

Ubersmake:
There were two PRs that accomplished similar ends, #1426 and #1455 . We ultimately ended up merging #1426 because it better deals with *preventing* HTML from appearing in product filtering options.

The differences between the two approaches can be read through here:
- https://github.com/helpers/handlebars-helpers#sanitize
- http://handlebarsjs.com/#html-escaping

1426:
![1426](https://user-images.githubusercontent.com/1546172/55906869-16f0b900-5b8a-11e9-995b-7a0e53c7479d.png)

1455:
![1455](https://user-images.githubusercontent.com/1546172/55906875-1ce69a00-5b8a-11e9-89a8-2db146f5e5e4.png)

Ubersmake:
This has been fixed in [3.4.1](https://github.com/bigcommerce/cornerstone/releases/tag/3.4.1) as a part of #1426. This should be published to all stores within the next two weeks.

PR #1426, opened 2019-01-21
Sanitize faceted search item's title

A-Jarrell:
#### What?

This pull request uses the handlebars helper {{ sanitize }} to remove HTML tags from facets' titles on category pages. 

#### Tickets / Documentation

- [Issue](https://github.com/bigcommerce/cornerstone/issues/1416)

PR #1455, opened 2019-02-20
**Sample** facet shows HTML as text

NjTheCoder:
As per #1416


junedkazi:
Please move all the changes from the other PR's closed into this PR.

Ubersmake:
Sorry for this. We ended up going with the implementation in #1426 , the reason being that `sanitize` makes sure that no HTML remains in the product filtering options.

Issue #1442, opened 2019-02-12
Back button does not "work on search results page


carsonreinke:
### Expected behavior
1.  Perform a search from the home page
2.  From the search results page, press the back button
3.  Taken back to the home page

### Actual behavior
1.  Perform a search from the home page
2.  From the search results page, press the back button
3.  Nothing, but query string `&section=product` is removed

### Steps to reproduce behavior
Same as actual behavior.

https://drive.google.com/file/d/14LZQyKg7kOExW4GVHgh4MdrBJCsK45n7/view?usp=sharing

jbruni:
Has this issue been fixed?
Is this commit a fix to this issue? :arrow_right: https://github.com/bigcommerce/cornerstone/commit/5322976e7359cd424c36229e2d7885329b13bff8

PR #1461, opened 2019-03-05
Fixed pagination and search page weirdness.

flair-conor:
#### What?

All this does is change how the system currently handles history so that you can use the back and forwards buttons native to the browser to navigate paginated pages and the search page.

#### Tickets / Documentation

#1442 search page issue.

#### Replication Steps

Just go to any category/brand/search that has pagination and traverse the pages. Clicking back on your browser will set the url back, but not process another AJAX request to reload said page.

This is probably not the best way to do this, but it fixes the issues regardless.

### DIFF
--- a/assets/js/theme/common/url-utils.js
+++ b/assets/js/theme/common/url-utils.js
@@ -8,6 +8,10 @@ const urlUtils = {
         $(window).trigger('statechange');
     },
 
+    searchGoToUrl: () => {
+        $(window).trigger('statechange');
+    },
+
     replaceParams: (url, params) => {
         const parsed = Url.parse(url, true);
         let param;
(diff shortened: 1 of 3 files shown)


assets/js/theme/common/url-utils.js
```diff
@@ -8,6 +8,10 @@ const urlUtils = {
         $(window).trigger('statechange');
     },
 
+    searchGoToUrl: () => {
```
junedkazi: How is this different from `goToUrl` method above ?

junedkazi: Also does this handle tabs on a page ?

junedkazi:
I am going to close this PR since there has been no activity on this for the past 1 year.

PR #1469, opened 2019-03-29
fixed update product qty by typing the new qty in the cart page

bc-zoharmuzafi:
#### What?
Update product qty in cart page by typing the new qty (not using the arrow), didn't work and the cart was keeping the old qty after refreshing. This pr fix this functionality.
 
#### Screenshots
Screen Shot 2019-03-29 at 2 52 41 PM
Screen Shot 2019-03-29 at 2 52 50 PM

### DIFF
--- a/assets/js/theme/cart.js
+++ b/assets/js/theme/cart.js
@@ -67,11 +67,12 @@ export default class Cart extends PageManager {
         const oldQty = preVal !== null ? preVal : minQty;
         const minError = $el.data('quantityMinError');
         const maxError = $el.data('quantityMaxError');
-        const newQty = parseInt(Number($el.attr('value')), 10);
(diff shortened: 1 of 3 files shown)


assets/js/theme/cart.js
```diff
@@ -67,11 +67,11 @@ export default class Cart extends PageManager {
         const oldQty = preVal !== null ? preVal : minQty;
         const minError = $el.data('quantityMinError');
         const maxError = $el.data('quantityMaxError');
-        const newQty = parseInt(Number($el.attr('value')), 10);
+        const newQty = parseInt(Number($el.prop('value')), 10);
```
mattolson: Is this a text box? Shouldn't we use `val()` for this?

Issue #1367, opened 2018-10-02
Product filters do not keep the previously selected color filter when the price filter is the top.


anilgautamm:
### Expected behavior
Product filters should preserve filters after range selection when price filter is the top.

### Actual behavior
Product filters do not keep the previously selected color filter when the price filter is the top.

### Steps to reproduce the behavior
when the price filter is the top.
https://www.useloom.com/share/f3bcb360539243beb01b6488d21f9e08

Ubersmake:
First mentioned in https://github.com/bigcommerce/cornerstone/pull/1232#issuecomment-426264136.

HikeOn:
Hi,

The issue got resolved for us after we replaced the code 
  
   **onRangeSubmit(event) {
        event.preventDefault();
        if (!this.priceRangeValidator.areAll(nod.constants.VALID)) {
            return;
        }
        const url = Url.parse(window.location.href);
        const queryParams = decodeURI($(event.currentTarget).serialize());
        urlUtils.goToUrl(Url.format({ pathname: url.pathname, search: `?${queryParams}` }));
    }**


in **faceted-search.js** file with

 **onRangeSubmit(event) {
         event.preventDefault();
          if (!this.priceRangeValidator.areAll(nod.constants.VALID)) {
            return;
          }
        const url = Url.parse(window.location.href);        
        const avoidSearchQuery = 'search_query=';
        const urlParams = new URLSearchParams(window.location.search);
        const avoidSortValue = 'sort='+ urlParams.get('sort');
        let queryParams = decodeURI($(event.currentTarget).serialize());
        let path = url.path;
        let urlValue;
        if(path.indexOf('?') !== -1){
            path = path.replace(avoidSortValue,'');
            queryParams = queryParams.replace(avoidSortValue,'');
            if(path.indexOf(avoidSearchQuery) !== -1){
                if(path.indexOf(avoidSearchQuery+'&m') !== -1){
                   path = path.replace(avoidSearchQuery+'&min_price='+urlParams.get('min_price')+'&max_price='+urlParams.get('max_price'),'');                  
                }else{
                    path = path.replace(avoidSearchQuery,'');                  
                }
            }
            urlValue = path + `&${queryParams}`;
            urlValue = urlValue.replace(/&&*\[\]=*/g,'').replace(/&&*/g,'&').replace(/\?&/g,'\?');
            if(urlParams.get('sort') !== null){
                urlValue = urlValue + '&' + avoidSortValue;
            }
[...]

PR #1471, opened 2019-04-05
fix(CP): ES-156 Fixed the filtering issues with price range filter selection

bc-krishsenthilraj:
#### What?

Product filtering Price Filter clears other filters below it (or after in the query URL).

#### Tickets / Documentation

https://jira.bigcommerce.com/browse/ES-156

`Note:` Please verify the URL params with selected Facets in each below image.

Image 1: **Resultant Products with all 3 brands and price range with 10-55 and `Field01` with 3 colors**.

image

Image 2: **Resultant Products with 3 brands and price range with 20-50 and `Field01` with 3 colors.**

image

Image 3: Now only with 2 brands and price range with 20-40 and `Field01` with 3 colors. But, due to the price range filter only 2 colors products are displayed.

image

@bigcommerce/cp-dt

### DIFF
--- a/assets/js/theme/common/faceted-search.js
+++ b/assets/js/theme/common/faceted-search.js
@@ -380,10 +380,17 @@ class FacetedSearch {
             return;
         }
 
-        const url = Url.parse(window.location.href);
-        const queryParams = decodeURI($(event.currentTarget).serialize());
+        const url = Url.parse(window.location.href, true);
+        let queryParams = decodeURI($(event.currentTarget).serialize()).split('&');
+        queryParams = urlUtils.parseQueryParams(queryParams);
 
-        urlUtils.goToUrl(Url.format({ pathname: url.pathname, search: `?${queryParams}` }));
+        for (const key in queryParams) {
+            if (queryParams.hasOwnProperty(key)) {
+                url.query[key] = queryParams[key];
+            }
+        }
+
+        urlUtils.goToUrl(Url.format({ pathname: url.pathname, search: urlUtils.buildQueryString(url.query) }));
     }
 
     onStateChange() {
--- a/assets/js/theme/common/url-utils.js
+++ b/assets/js/theme/common/url-utils.js
@@ -45,6 +45,26 @@ const urlUtils = {
 
         return out.substring(1);
     },
+
+    parseQueryParams: (queryData) => {
+        const params = {};
+
+        for (let i = 0; i < queryData.length; i++) {
+            const temp = queryData[i].split('=');
+
+            if (temp[0] in params) {
+                if (Array.isArray(params[temp[0]])) {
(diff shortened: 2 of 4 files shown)


assets/js/theme/common/url-utils.js
```diff
@@ -45,6 +45,28 @@ const urlUtils = {
 
         return out.substring(1);
     },
+
+    parseQueryParams: (queryData) => {
+        const params = {};
+        let temp;
```
lord2800: This can go inside the `for` loop.

assets/js/theme/common/url-utils.js
```diff
@@ -45,6 +45,28 @@ const urlUtils = {
...
+
+        for (let i = 0; i < queryData.length; i++) {
+            temp = queryData[i].split('=');
+
+            if (temp[0] in params) {
+                if (Array.isArray(params[temp[0]])) {
+                    const length = params[temp[0]].length;
+                    params[temp[0]][length] = temp[1];
```
lord2800:
```suggestion
                    params[temp[0]][length] = temp[1];
                    params[temp[0]].push(temp[1]);
```

assets/js/theme/common/url-utils.js
bc-krishsenthilraj: 👍 will do it

assets/js/theme/common/url-utils.js
```diff
@@ -45,6 +45,26 @@ const urlUtils = {
 
         return out.substring(1);
     },
+
+    parseQueryParams: (queryData) => {
```
junedkazi: Can we please add a test for this method.

bc-krishsenthilraj: sure will add it. Thanks for bringing it up :+1

assets/js/theme/common/faceted-search.js
```diff
@@ -380,10 +380,17 @@ class FacetedSearch {
             return;
         }
 
-        const url = Url.parse(window.location.href);
-        const queryParams = decodeURI($(event.currentTarget).serialize());
+        const url = Url.parse(window.location.href, true);
```
junedkazi: We should also add a test for this change https://github.com/bigcommerce/cornerstone/blob/master/assets/js/test-unit/theme/common/faceted-search.spec.js

junedkazi:
related https://github.com/bigcommerce/cornerstone/pull/1232

bc-ranji:
💚

junedkazi:
@bc-krishsenthilraj when I am looking at the 3rd screenshot I see 3 brands (38, 39, 40) in the url  but only 2 brands selected. Shouldn't the url be updated ?

Ubersmake:
This may also be a fix for #1367

PR #1472, opened 2019-04-07
Fix submenu click on mobile navigation

kylebolstad:
#### What?

Change the event handler to correctly call `this.onSubMenuClick`. Without this fix, mobile navigation is not showing/hiding siblings correctly.


Ubersmake:
Hi @kylebolstad . I'm trying to reproduce the issue this PR fixes, and I'm not sure how to set things up. How do I get mobile navigation into a state where sibling elements are not being shown or hidden correctly?

kylebolstad:
Good point! I totally forgot to mention that this seems to only happen when `"navigation_design": "alternate"` is set in `config.json`. That should be the only thing required to replicate the bug.

Ubersmake:
Thanks for this. I'm attaching GIFs so that anyone looking at this PR knows what actually changed. It took me some time to figure out what was being affected: expanding and collapsing category navigation.

Before:
before

After:
after

PR #1475, opened 2019-04-10
CHECKOUT-4039 Checkout UI Refresh

capsula4:
#### What?
Checkout UI Refresh
- Sets Monserrat as main font (weight: 500)
- Loads Monserrat bold for headers (weight: 700)
- Transparent inner shadow for input fields (not removed so it's backwards compatible)
- Refined "Order Summary" drop shadow

As per @valterfatia

#### Tickets / Documentation
CHECKOUT-4039

Screen Shot 2019-04-10 at 5 00 31 pm

Screen Shot 2019-04-10 at 5 00 45 pm


config.json
```diff
@@ -229,23 +229,23 @@
...
+    "optimizedCheckout-contentSecondary-font": "Google_Montserrat_500",
+    "optimizedCheckout-buttonPrimary-font": "Google_Montserrat_500",
     "optimizedCheckout-buttonPrimary-color": "#ffffff",
     "optimizedCheckout-buttonPrimary-colorHover": "#ffffff",
     "optimizedCheckout-buttonPrimary-colorActive": "#ffffff",
     "optimizedCheckout-buttonPrimary-backgroundColor": "#333333",
     "optimizedCheckout-buttonPrimary-backgroundColorHover": "#666666",
     "optimizedCheckout-buttonPrimary-backgroundColorActive": "#000000",
-    "optimizedCheckout-buttonPrimary-borderColor": "#cccccc",
+    "optimizedCheckout-buttonPrimary-borderColor": "transparent",
```
junedkazi: Is `transparent` a valid value for a color picker ?

schema.json
```diff
@@ -2173,7 +2173,7 @@
           {
             "group": "Montserrat",
             "label": "Montserrat",
-            "value": "Google_Montserrat_400"
+            "value": "Google_Montserrat_600"
```
junedkazi: I understand that we are changing the default from 400 to 600. But I think we should keep 400 in the list to be backward compatible ?

capsula4: what do you mean with backward compatible? Merchant would have to opt-in to the new update. Also, this is only changing `checkout` fonts, not main storefront fonts.

Also merchants don't have a way to change font weight in theme editor.

config.json
capsula4: 🤷‍♀️ not sure but I guess I could just pick the backgroundColor then so it looks as transparent

schema.json
junedkazi: When a merchant updates their theme their configuration changes are merged but their saved values are preserved. So when the merchant uses theme editor to switch font he won’t have access to his previous value since it won’t be available in the option set.

config.json
junedkazi: Can you bundle the theme and test the change inside theme editor

schema.json
capsula4: right, I see 👁  would probably need to test this scenarios to see how it works

config.json
capsula4: It definitely comes up in the editor, but once you change it to something else, you can make it transparent.

Screen Shot 2019-04-11 at 5 29 04 pm

schema.json
capsula4: Added `Google_Montserrat_400` as `Montserrat Light` for backward compatibility.

capsula4:
@junedkazi let me know if there's any outstanding issue with this PR or if it's good to go. Thanks!

config.json
```diff
@@ -502,9 +502,9 @@
         "spinner-borderColor-light": "#fab9a3",
         "optimizedCheckout-header-backgroundColor": "#555555",
         "optimizedCheckout-headingPrimary-color": "#eeeeee",
-        "optimizedCheckout-headingPrimary-font": "Google_Roboto_400",
+        "optimizedCheckout-headingPrimary-font": "Google_Montserrat_600",
         "optimizedCheckout-headingSecondary-color": "#ffffff",
-        "optimizedCheckout-headingSecondary-font": "Google_Roboto_400",
+        "optimizedCheckout-headingSecondary-font": "Google_Montserrat_600",
```
mattolson: Can we just use 500 or 700 for this? Using this will cause an additional web font to load for very little benefit.

capsula4: oooops, this was actually my mistake, it should have been 700.

schema.json
```diff
@@ -2237,9 +2247,14 @@
           },
           {
             "group": "Montserrat",
-            "label": "Montserrat",
+            "label": "Montserrat Light",
             "value": "Google_Montserrat_400"
           },
+          {
+            "group": "Montserrat",
+            "label": "Montserrat",
```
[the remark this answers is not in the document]
capsula4: I could just use a number, but not sure if that will be very friendly/understandable for the merchant. I was using the names provided by google fonts

Screen Shot 2019-04-16 at 10 39 15 am

PR #1478, opened 2019-04-15
MERC-4842 Add additional regions

kchu93:
#### What?

Enable the following regions in cornerstone for enabling adding new content via widgets onto these pages:
(1) Homepage
(2) Brand
(3) Brands
(4) Category
(5) Product Page
(6) Search Page

#### Tickets / Documentation
[MERC-4842](https://jira.bigcommerce.com/browse/MERC-4842)

#### Screenshots
**Home Page**
Screen Shot 2019-04-15 at 9 44 38 AM
Screen Shot 2019-04-15 at 9 44 41 AM
Screen Shot 2019-04-15 at 9 44 43 AM

**Brands**
Screen Shot 2019-04-15 at 9 46 18 AM
Screen Shot 2019-04-15 at 9 46 21 AM

**Brand**
Screen Shot 2019-04-15 at 9 47 06 AM
Screen Shot 2019-04-15 at 9 47 18 AM

**Category**
Screen Shot 2019-04-15 at 9 47 39 AM
Screen Shot 2019-04-15 at 9 47 47 AM

**Product Page**
Screen Shot 2019-04-15 at 9 48 39 AM

**Search**
Screen Shot 2019-04-15 at 9 49 42 AM
Screen Shot 2019-04-15 at 9 49 52 AM

### DIFF
--- a/templates/pages/brand.html
+++ b/templates/pages/brand.html
@@ -21,6 +21,7 @@ brand:
     </div>
 {{/if}}
 <h1 class="page-heading">{{brand.name}}</h1>
+{{{region name="brand_zone_1"}}}
 <div class="page">
     <aside class="page-sidebar" id="faceted-search-container">
         {{> components/brand/sidebar}}
@@ -32,6 +33,7 @@ brand:
         {{else}}
             <p>{{lang 'brands.no_products'}}</p>
         {{/if}}
+        {{{region name="brand_zone_2"}}}
     </main>
 </div>
 
--- a/templates/pages/brands.html
+++ b/templates/pages/brands.html
@@ -8,6 +8,7 @@ brands:
 
 <main class="page">
     <h1 class="page-heading">{{lang 'brand.label'}}</h1>
+    {{{region name="brands_zone_1"}}}
     <ul class="brandGrid">
         {{#each brands}}
             <li class="brand">
@@ -29,6 +30,7 @@ brands:
         {{/each}}
     </ul>
     {{> components/common/paginator pagination.brands}}
+    {{{region name="brands_zone_2"}}}
 </main>
 {{/partial}}
 {{> layout/base}}
--- a/templates/pages/category.html
+++ b/templates/pages/category.html
@@ -22,6 +22,7 @@ category:
 {{/if}}
 {{#unless theme_settings.hide_category_page_heading }}
     <h1 class="page-heading">{{category.name}}</h1>
+    {{{region name="category_zone_1"}}}
 {{/unless}}
 {{{category.description}}}
 <div class="page">
@@ -47,6 +48,7 @@ category:
         {{else}}
             <p>{{lang 'categories.no_products'}}</p>
         {{/if}}
+        {{{region name="category_zone_2"}}}
     </main>
 </div>
 
--- a/templates/pages/home.html
+++ b/templates/pages/home.html
@@ -12,9 +12,11 @@ blog:
         limit: {{theme_settings.homepage_blog_posts_count}}
 ---
 {{#partial "hero"}}
+    {{{region name="home_zone_1"}}}
     {{#if carousel}}
         {{> components/carousel arrows=theme_settings.homepage_show_carousel_arrows}}
     {{/if}}
+    {{{region name="home_zone_2"}}}
 {{/partial}}
 
 {{#partial "page"}}
@@ -27,14 +29,17 @@ blog:
     {{#if products.featured}}
         {{> components/products/featured products=products.featured columns=theme_settings.homepage_featured_products_column_count}}
     {{/if}}
+    {{{region name="home_zone_3"}}}
 
     {{#if products.top_sellers}}
         {{> components/products/top products=products.top_sellers columns=theme_settings.homepage_top_products_column_count}}
     {{/if}}
+    {{{region name="home_zone_4"}}}
 
     {{#if products.new}}
         {{> components/products/new products=products.new columns=theme_settings.homepage_new_products_column_count}}
     {{/if}}
+    {{{region name="home_zone_5"}}}
 </div>
 {{/partial}}
 {{> layout/base}}
--- a/templates/pages/product.html
+++ b/templates/pages/product.html
@@ -19,6 +19,8 @@ product:
     <div itemscope itemtype="http://schema.org/Product">
         {{> components/products/product-view schema=true  }}
 
+        {{{region name="product_zone_1"}}}
+
         {{#if product.videos.list.length}}
             {{> components/products/videos product.videos}}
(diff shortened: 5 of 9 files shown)


templates/pages/search.html
```diff
@@ -138,12 +138,13 @@ <h5 class="suggestion-title">{{lang 'forms.search.suggestions.title'}}</h5>
...
                 {{> components/common/paginator pagination.content_results reload=true}}
             </div>
-
         {{/if}}
 
         <div id="product-listing-container" {{#if forms.search.section '!=' 'product'}}class="u-hiddenVisually"{{/if}}>
             {{> components/search/product-listing}}
         </div>
+
+        {{{region name="page_content"}}}
```
bookernath: Why is this below the search results instead of above it?

templates/pages/product.html
```diff
@@ -19,6 +19,8 @@
     <div itemscope itemtype="http://schema.org/Product">
         {{> components/products/product-view schema=true  }}
 
+        {{{region name="page_content"}}}
```
bookernath: Could this be better named? I think consumers would want to know the relationship between this region and other important PDP components such as:

1. Product Price
2. Product Reviews
3. Product Description
4. Add-to-cart button

templates/pages/home.html
```diff
@@ -27,14 +28,17 @@
...
     {{/if}}
 
     {{#if products.top_sellers}}
         {{> components/products/top products=products.top_sellers columns=theme_settings.homepage_top_products_column_count}}
+        {{{region name="page_content"}}}
     {{/if}}
 
     {{#if products.new}}
         {{> components/products/new products=products.new columns=theme_settings.homepage_new_products_column_count}}
+        {{{region name="page_content"}}}
```
bookernath: Could be better named to indicate its relation to New Products

templates/pages/home.html
```diff
@@ -27,14 +28,17 @@
 <div class="main full">
     {{#if products.featured}}
         {{> components/products/featured products=products.featured columns=theme_settings.homepage_featured_products_column_count}}
+        {{{region name="page_content"}}}
     {{/if}}
 
     {{#if products.top_sellers}}
         {{> components/products/top products=products.top_sellers columns=theme_settings.homepage_top_products_column_count}}
+        {{{region name="page_content"}}}
```
bookernath: Could be better named to indicate its relation to Top Selling Products

templates/pages/home.html
```diff
@@ -27,14 +28,17 @@
 <div class="main full">
     {{#if products.featured}}
         {{> components/products/featured products=products.featured columns=theme_settings.homepage_featured_products_column_count}}
+        {{{region name="page_content"}}}
```
bookernath: Could be better named to indicate its relation to Featured Products. Also, why is this below instead of above?

templates/pages/home.html
```diff
@@ -16,6 +16,7 @@
         {{> components/carousel arrows=theme_settings.homepage_show_carousel_arrows}}
     {{/if}}
     {{{snippet 'home_content'}}}
+    {{{region name="page_content"}}}
```
bookernath: 🍹 might make sense to remove the Snippet as well in this same PR.

templates/pages/category.html
```diff
@@ -22,6 +22,7 @@
 {{/if}}
 {{#unless theme_settings.hide_category_page_heading }}
     <h1 class="page-heading">{{category.name}}</h1>
+    {{{region name="page_content"}}}
```
bookernath: 🍹 might make sense to remove the Snippet as well in this same PR.

templates/pages/brands.html
```diff
@@ -29,6 +30,7 @@ <h4 class="card-title">
         {{/each}}
     </ul>
     {{> components/common/paginator pagination.brands}}
+    {{{region name="page_content"}}}
```
bookernath: Could these be better named to give some indication of positioning? "Top" vs "bottom" etc?

templates/pages/brand.html
```diff
@@ -21,6 +21,7 @@
     </div>
 {{/if}}
 <h1 class="page-heading">{{brand.name}}</h1>
+{{{region name="page_content"}}}
```
bookernath: 🍹 might make sense to remove the Snippet as well in this same PR.

templates/pages/brands.html
```diff
@@ -8,6 +8,7 @@
 
 <main class="page">
     <h1 class="page-heading">{{lang 'brand.label'}}</h1>
+    {{{region name="page_content"}}}
```
junedkazi: How can 2 regions in the same template placed at different locations have the same name ? We need need to be specific like `page_content_header` or `page_content_sidebar`.

templates/pages/home.html
junedkazi: I am not sure if we should have regions inside very specific if conditions like these. If we do then we need to be very specific with the naming as Nathan pointed out above.

templates/pages/product.html
junedkazi: I might tend to disagree with building regions with very specific naming unless we have a very specific use case to it. I think we should keep them generic but they should be better named. The reason for keeping it generic is to make them accessible across multiple templates.

templates/pages/search.html
bc-jcruiz: The use case is to potentially display context-aware content based on the search rather than fix content above the search. This can change as we get feedback

templates/pages/category.html
```diff
@@ -44,6 +45,7 @@ <h1 class="page-heading">{{category.name}}</h1>
     <main class="page-content" id="product-listing-container">
         {{#if category.products}}
             {{> components/category/product-listing}}
+            {{{region name="category_page_content"}}}
```
BC-EChristensen: Is this the right location? It looks like it would appear below the list of products and would *not* appear if there are no products.

templates/pages/brand.html
```diff
@@ -32,6 +33,7 @@ <h1 class="page-heading">{{brand.name}}</h1>
         {{else}}
             <p>{{lang 'brands.no_products'}}</p>
         {{/if}}
+        {{{region name="brand_page_content"}}}
```
BC-EChristensen: Is it correct for this to be below the list of brand products?

kchu93:
@BC-EChristensen @pedelman ♻️ trying zone names with numbers

kchu93:
After speaking with @BC-EChristensen we are going to close this ticket for now to talk a bit more about how we handle to handle naming for regions. I can reopen this PR once we have a bit more clarity on how we want to approach this. Thanks for all the input everyone!

PR #1483, opened 2019-04-22
Fix to open Bulk Pricing modal from Quick View.

Ubersmake:
#### What?

Currently, the Bulk Pricing link doesn't work from a Quick View modal. Clicking on the "Buy in bulk and save" link effectively closes the Quick View modal.

This PR replicates the behavior for Write a Review from Quick View. Now, clicking on the "Buy in bulk and save" link from Quick View will open the Product page and immediately open the Bulk Pricing modal.

Before:
before

After:
after

### DIFF
--- a/templates/components/products/bulk-discount-rates.html
+++ b/templates/components/products/bulk-discount-rates.html
@@ -1,12 +1,12 @@
 {{#if bulk_discount_rates.length}}
     {{#if product}}<dt class="productView-info-name">{{lang 'products.bulk_pricing.title'}}</dt>{{/if}}
     <dd class="productView-info-value">
-        <a href="{{url}}"
-           {{#unless is_ajax }}data-reveal-id="bulkPricingModal{{#if id}}-{{id}}-{{$index}}{{/if}}"{{/unless}}>
+        <a href="{{product.url}}#bulk_pricing"
+           {{#unless is_ajax }}data-reveal-id="modal-bulk-pricing"{{/unless}}>
             {{lang 'products.bulk_pricing.view'}}
         </a>
     </dd>
-    <div id="bulkPricingModal{{#if id}}-{{id}}-{{$index}}{{/if}}" class="modal modal--small" data-reveal>
+    <div id="modal-bulk-pricing" class="modal modal--small" data-reveal>
         <div class="modal-header">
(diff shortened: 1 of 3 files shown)


templates/components/products/bulk-discount-rates.html
```diff
@@ -1,12 +1,12 @@
 {{#if bulk_discount_rates.length}}
     {{#if product}}<dt class="productView-info-name">{{lang 'products.bulk_pricing.title'}}</dt>{{/if}}
     <dd class="productView-info-value">
-        <a href="{{url}}"
-           {{#unless is_ajax }}data-reveal-id="bulkPricingModal{{#if id}}-{{id}}-{{$index}}{{/if}}"{{/unless}}>
+        <a href="{{product.url}}#bulk_pricing"
```
Ubersmake: This makes sure that the product URL is in the link, though it hasn't caused issues before. Follows the same convention as `#write_review`

templates/components/products/bulk-discount-rates.html
```diff
@@ -1,12 +1,12 @@
...
+           {{#unless is_ajax }}data-reveal-id="modal-bulk-pricing"{{/unless}}>
```
Ubersmake: Removed the `id` identifier because this modal is only used here. Presumably this id is here in the case of collisions. Uses the `modal-some-name` pattern used in `modal-review-form`.

bookernath: 🍹 from a "minimum viable change" perspective, is this needed? Or did this just feel like good cleanup? I believe bulk discount rates _can_ be rendered on PLPs with [certain configurations of theme config](https://github.com/bigcommerce/cornerstone/blob/master/config.json#L334) in which case this might serve a purpose (collisions, as you mentioned)

Ubersmake: It felt like cleanup. Reverting that change until I'm absolutely 100% sure it wouldn't affect anything.

Ubersmake: I was wrong. `product.url` **is** necessary. It makes no difference on the product details page, but it's necessary where multiple products are shown, like a category page. If it's just `url`, the link won't actually go anywhere. It'll go to a nonexistent anchor on the same page.

PR #1484, opened 2019-04-22
fix(storefront): STRF-6540 Fire facebook pixel AddToCart in modal

megdesko:
#### What?

Add handlebars for `remote_api_scripts` to the preview.html we use for the add to cart preview, so that the Facebook `AddToCart` pixel will fire from the modal window.

#### Tickets / Documentation

- [BC Jira ticket](https://jira.bigcommerce.com/browse/STRF-6540)

Tested this with a theme I changed & bundled locally.

Screencast showing pixel firing

### DIFF
--- a/templates/components/cart/preview.html
+++ b/templates/components/cart/preview.html
@@ -112,3 +112,4 @@
         </section>
     {{/if}}
 </div>
+{{{ remote_api_scripts }}}


carsonreinke:
@megdesko You know what I noticed, on the cart page this is only working when the cart is empty 🤔 

https://github.com/bigcommerce/cornerstone/blob/7939b238f1cb2d679b5e9aea06ffa1ad248a6be5/templates/pages/cart.html#L51

megdesko:
@carsonreinke interesting. 🤔  We're trying to make sure we get the `AddToCart` pixel firing exactly once on each add to cart event.  Can you let me know your store so I can have a look?  I haven't yet observed that particular pattern of event firing.

Thanks!

carsonreinke:
@megdesko It is right in there:

```
{{#if cart.items.length}}
            ...  
{{else}}
            <h3>{{lang 'cart.checkout.empty_cart'}}</h3>
            {{{ remote_api_scripts }}}
{{/if}}
```

megdesko:
We're going to fix this in an internal repository.

carsonreinke:
🤷‍♂️

mattolson:
Quick update on this.

`remote_api_scripts` was never meant to be exposed to Cornerstone, and these usages will be removed. We are going to replace with server side functionality that ensures that the appropriate scripts are sent along with ajax partial renders (done via stencil-options: render_with).

So please don't rely on this attribute or add it to your theme -- it is undocumented and will be removed.

PR #1486, opened 2019-04-29
Update default layout settings for PayPal smart buttons

bc-anna:
#### What?

Changes the default layout options for PayPal smart buttons based on recommendations from PayPal - vertical layout is preferred, and the size and tagline options must be acceptable matching values based on the PayPal JS SDK (tagline must be false, and buttons must be larger than small).

#### Tickets / Documentation

https://developer.paypal.com/docs/archive/checkout/how-to/customize-button/

Before:
Screen Shot 2019-04-18 at 4 49 21 PM

After:
Screen Shot 2019-04-18 at 1 00 52 PM


flyingL123:
Hi @bc-anna,

I was hoping someone could please explain how to adjust this layout? It's insane to me that PayPal thinks it's ok to take up this many lines with this much content as part of their buttons. I just want a normal "PayPal Checkout" button. I don't want credit card logos and a "Powered by PayPal" tagline on my website.

I have seen the customization options listed here in the documentation:

https://developer.bigcommerce.com/stencil-docs/template-files/customize-stencil-checkout/paypal-smart-buttons

However, nothing I do seems to make any changes to the button display. Am I missing something here? For example, if I set `"paymentbuttons-paypal-tagline": false`, the tagline is still there. If I set `"paymentbuttons-paypal-fundingicons": false`, the credit card icons are still there.

Even if I change `"paymentbuttons-paypal-shape": "rectangle"`, nothing changes. Is there something I'm missing here? Are you sure it's working?

flyingL123:
The problem seems to be that changing the value doesn't get passed through to the button. I think in order for my options to work as desired I need to use the horizontal layout. So when I change `"paymentbuttons-paypal-layout": "horizontal",`, I still see `style.layout=vertical` being passed into the button URL.

Why is that?

flyingL123:
Not sure what changed but this suddenly started working.

PR #1491, opened 2019-05-07
MERC-4842 Add additional regions

kchu93:
#### What?

Enable the following regions in cornerstone for enabling adding new content via widgets onto these pages:
(1) Homepage
(2) Brand
(3) Brands
(4) Category
(5) Product Page
(6) Search Page

#### Tickets / Documentation
[MERC-4842](https://jira.bigcommerce.com/browse/MERC-4842)

### DIFF
--- a/templates/components/common/header.html
+++ b/templates/components/common/header.html
@@ -24,8 +24,8 @@
             {{> components/common/store-logo}}
         </div>
     {{/if}}
-    {{{region name="header_bottom"}}}
     <div class="navPages-container" id="menu" data-menu>
         {{> components/common/navigation-menu}}
     </div>
 </header>
+{{{region name="header_bottom"}}}
--- a/templates/pages/brand.html
+++ b/templates/pages/brand.html
@@ -21,6 +21,7 @@ brand:
     </div>
 {{/if}}
 <h1 class="page-heading">{{brand.name}}</h1>
+{{{region name="brand_below_header"}}}
 <div class="page">
     <aside class="page-sidebar" id="faceted-search-container">
         {{> components/brand/sidebar}}
@@ -32,6 +33,7 @@ brand:
         {{else}}
             <p>{{lang 'brands.no_products'}}</p>
         {{/if}}
+        {{{region name="brand_below_content"}}}
     </main>
 </div>
 
--- a/templates/pages/home.html
+++ b/templates/pages/home.html
@@ -12,9 +12,11 @@ blog:
         limit: {{theme_settings.homepage_blog_posts_count}}
 ---
 {{#partial "hero"}}
+    {{{region name="home_below_menu"}}}
     {{#if carousel}}
         {{> components/carousel arrows=theme_settings.homepage_show_carousel_arrows}}
     {{/if}}
+    {{{region name="home_below_carousel"}}}
 {{/partial}}
 
 {{#partial "page"}}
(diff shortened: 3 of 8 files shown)


templates/pages/home.html
```diff
@@ -27,14 +29,17 @@
     {{#if products.featured}}
         {{> components/products/featured products=products.featured columns=theme_settings.homepage_featured_products_column_count}}
     {{/if}}
+    {{{region name="home_below_featured"}}}
```
bc-jcruiz: this should contain `featured_products` instead of just `featured`

templates/pages/home.html
```diff
@@ -27,14 +29,17 @@
...
 
     {{#if products.top_sellers}}
         {{> components/products/top products=products.top_sellers columns=theme_settings.homepage_top_products_column_count}}
     {{/if}}
+    {{{region name="home_below_top"}}}
```
bc-jcruiz: this should contain `top_products` instead of just `top` is confusing

templates/pages/home.html
```diff
@@ -27,14 +29,17 @@
...
 
     {{#if products.new}}
         {{> components/products/new products=products.new columns=theme_settings.homepage_new_products_column_count}}
     {{/if}}
+    {{{region name="home_below_new"}}}
```
bc-jcruiz: this should contain `new_products` instead of just `new`

bc-jcruiz:
@kchu93 did you include removing the snippets or that was done as part of a separate PR?

kchu93:
@bookernath @BC-EChristensen So for this PR we have removed the `page_content` region until discussion has been finalized, but aside from that, all other regions are ready to be merged! Thanks again for all the help / review on this

templates/components/common/header.html
```diff
@@ -24,8 +24,8 @@ <h1 class="header-logo header-logo--{{theme_settings.logo-position}}">
             {{> components/common/store-logo}}
         </div>
     {{/if}}
-    {{{region name="header_bottom"}}}
     <div class="navPages-container" id="menu" data-menu>
         {{> components/common/navigation-menu}}
     </div>
 </header>
+{{{region name="header_bottom"}}}
```
junedkazi: If this is header bottom shouldn't this be inside the header tag ?

kchu93: So i spoke with @pedelman / @BC-EChristensen and i believe they wanted to move this outside of the header tag due to its placement. I can move this back if necessary, but they may be able to speak on this as well!

junedkazi: I don't want you to move ... I am just trying to understand the naming and the reasoning behind it.

templates/pages/brand.html
```diff
@@ -21,6 +21,7 @@
     </div>
 {{/if}}
 <h1 class="page-heading">{{brand.name}}</h1>
+{{{region name="brand_below_header"}}}
```
junedkazi: Can you please explain to me the reasoning behind when to use `bottom` and when to use `below` ?

templates/components/common/header.html
BC-EChristensen: Unfortunately we already have two regions, `header_bottom` and `product_below_price`, using two different naming schemes. We opted to adopt the second, but unfortunately we can't change this one.

You're right about `menu_bottom` below, though, we're changing that to `home_below_menu`

templates/pages/home.html
```diff
@@ -12,9 +12,11 @@
         limit: {{theme_settings.homepage_blog_posts_count}}
 ---
 {{#partial "hero"}}
+    {{{region name="home_below_menu"}}}
```
kchu93: Changed `menu_bottom` to `home_below_menu`

templates/pages/brand.html
kchu93: Hey Juned, similar to what Erik was stating above, but I believe we opted for `below` / 'above' as it would specify more so where a region would be.

BC-EChristensen:
@junedkazi Kevin made some changes to the naming to be more consistent. We can't change `header_bottom` unfortunately, though. Mind taking another look at this?

PR #1496, opened 2019-05-15
Made fix to unclickable reviews link

bc-williamkwon:
#### What?

Made fix to product reviews link not being clickable when config.json file has  "show_product_reviews_tabs": true

#### Tickets / Documentation

https://jira.bigcommerce.com/browse/STRF-5521

### DIFF
--- a/assets/js/theme/product/reviews.js
+++ b/assets/js/theme/product/reviews.js
@@ -24,6 +24,7 @@ export default class {
         const $content = $('#productReviews-content', this.$reviewsContent);
 
         $('.productView-reviewLink').on('click', () => {
+            $('.productView-reviewTabLink').click();
             if (!$content.hasClass('is-open')) {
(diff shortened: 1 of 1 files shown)


assets/js/theme/product/reviews.js
```diff
@@ -24,6 +24,7 @@ export default class {
         const $content = $('#productReviews-content', this.$reviewsContent);
 
         $('.productView-reviewLink').on('click', () => {
+            $('.productView-reviewTabLink').click();
```
sacr3dc0w: Should this be `$('.productView-reviewTabLink').trigger('click')`?

Only ask because I recall the change a while back from .click().

Screen Shot 2019-05-14 at 7 01 36 PM

Screen Shot 2019-05-14 at 7 03 51 PM

bc-williamkwon: Ok, about to make another pull request using that style instead.  Thank you.

PR #1498, opened 2019-05-15
STRF-5521 PDP Product Reviews Link Not Clickable

bc-williamkwon:
#### What?

Previously when config.json had show_product_reviews_tab as true the reviews link when clicked would do nothing.  This fix should make it so that instead the reviews tab is opened and window is directed to it.

#### Tickets / Documentation

https://jira.bigcommerce.com/browse/STRF-5521

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Draft
+--Fix to PDP Product Reviews Link Not Clickable. [1498](https://github.com/bigcommerce/cornerstone/pull/1498)
 
 ## 3.4.3 (2019-05-10)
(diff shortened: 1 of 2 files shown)


junedkazi:
Will it also support the use case when some one tries to hit a url with the anchor in it ? Example http://example.com/productpage/#product-reviews

bc-williamkwon:
Yes, it does.  It moves the page down to the reviews just as if it was show_product_reviews_tab:false

mattolson:
@bc-williamkwon Can you update the changelog as well?

bc-williamkwon:
updated

mattolson:
@bc-williamkwon Please squash your commits

bc-williamkwon:
Successfully did my first rebase to squash at least I believe I did.

PR #1502, opened 2019-05-17
chore(storefront): STRF-6720 Update jQuery to 3.4.1

bookernath:
#### What?

Update jQuery version to latest.

#### Tickets / Documentation

- [STRF-6720](https://jira.bigcommerce.com/browse/STRF-6720)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 # Changelog
 
 ## Draft
---Fix to PDP Product Reviews Link Not Clickable. [1498](https://github.com/bigcommerce/cornerstone/pull/1498)
+- Fix to PDP Product Reviews Link Not Clickable. [#1498](https://github.com/bigcommerce/cornerstone/pull/1498)
+- Update jQuery to 3.4.1 [#1502](https://github.com/bigcommerce/cornerstone/pull/1502)
 
 ## 3.4.3 (2019-05-10)
 - Fix to open Bulk Pricing modal from Quick View. [#1483](https://github.com/bigcommerce/cornerstone/pull/1483)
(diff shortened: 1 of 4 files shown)


sacr3dc0w:
https://www.i-programmer.info/news/167-javascript/12693-sizzle-out-in-jquery-34.html

> Now that many of these selectors have made their way into modern browsers, it’s almost time to say goodbye to Sizzle. But in order to remove Sizzle in jQuery 4.0, we will also need to remove what we refer to as positional selectors, which are non-standard selectors.
> **Specifically, jQuery 3.4.0 is deprecating :first, :last, :eq, :even, :odd, :lt, :gt, and :nth.** When we remove Sizzle, we’ll replace it with a small wrapper around querySelectorAll, and it would be almost impossible to reimplement these selectors without a larger selector engine.
> We think this trade-off is worth it. Keep in mind we will still support the positional methods, such as .first, .last, and .eq. Anything you can do with positional selectors, you can do with positional methods instead. They perform better anyway.
> - `:first`
> - `:last `
> - `:eq`
> - `:even`
> - `:odd`
> - `:lt`
> - `:gt`
> - `:nth`

Line 42 of common/select-option-plugin.js:

Replace:
`const $elementNowAtPreviousIndex = selectElement.find(`option:eq(${previousIndex})`);`

With:
const $elementNowAtPreviousIndex = selectElement.find('option').eq(previousIndex);

bookernath:
Thanks for the callout @sacr3dc0w ! We'll update.

bc-sandeep:
💚 smoke tested manually without util bump. loosk good

will also monitor automated tests on staging before doing next cornerstone release

PR #1505, opened 2019-05-22
CHECKOUT-4118 Remove remote_api_scripts from the templates

icatalina:
#### What?

This should avoid double inclusion of some of the analytics code. (eg. facebook pixel)

#### Tickets / Documentation

https://jira.bigcommerce.com/browse/CHECKOUT-4118

Before:
image

After:
image

@bigcommerce/storefront-team @lord2800 @megdesko

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Draft
 - Fix to PDP Product Reviews Link Not Clickable. [#1498](https://github.com/bigcommerce/cornerstone/pull/1498)
 - Update jQuery to 3.4.1 [#1502](https://github.com/bigcommerce/cornerstone/pull/1502)
+- Remove remote_api_scripts to avoid double firing analytics [#1505](https://github.com/bigcommerce/cornerstone/pull/1505)
 
 ## 3.4.3 (2019-05-10)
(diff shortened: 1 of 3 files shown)


lord2800:
This looks right to me, but I'll let @bigcommerce/storefront-team make the final call here.

aleachjr:
I suspect this might require regression testing anything included in remote api scripts. From what I recall this was required for Jirafe among other things.

lord2800:
It will definitely require regression testing, but this was _not_ the right fix for Jirafe/etc. In fact, adding this caused more problems than it solved--removing it is the right fix, we just need to make sure all of our ducks are in a row before we do it.

icatalina:
can someone manage the regression testing in your side? I don't really know what's included using the `remote_api_scripts` tag. 🤷‍♂

lord2800:
@bc-sandeep you mind taking this on?

mattolson:
@icatalina You have done a great service with this PR. I think this is the correct thing to do. We'll organize some regression testing just to be sure.

bc-junbai:
I checked an integration store with Cornerstone version 3.5.0, and Problem #3 double firing addToCart event doesn't happen anymore. 

LGTM 💚

PR #1507, opened 2019-05-27
Switch image loading to use srcset

bookernath:
#### Preface
The philosophy behind modern responsive image thinking is that the *browser itself* has the best knowledge of how large an image _needs_ to be when it's painting the page, and `srcset` enables the browser to optimize the image size to download based on its expected size.

As of now, `srcset` has [94% penetration](https://caniuse.com/#feat=srcset) of browser market share.

This is the first iteration of a solution to allow merchants to use [responsive images](https://css-tricks.com/responsive-images-youre-just-changing-resolutions-use-srcset/) ([`srcset`](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images)) in Cornerstone.

This depends on a new helper: https://github.com/bigcommerce/paper-handlebars/pull/42
And a new stencil-utils method: https://github.com/bigcommerce/stencil-utils/pull/106

#### What problem are we solving?

The goal is to meet the needs of two different types of shoppers on opposite ends of the performance spectrum:

- Those on high-resolution displays (such as Retina macs) and good internet connections who want high-quality imagery
- Those on smaller-screen mobile devices on unreliable or slow connections, who want a faster loading page and can't benefit from high-resolution imagery even when it's available

Historically, merchants have had to choose which shoppers and devices they want to cater to when uploading their image assets - if I uploaded a 4k carousel image to look good on desktop, I'd harm my load speed on mobile devices. If I optimized for mobile load speeds, the image would look bad when stretched out on the desktop.

Our first effort to fix this problem was by enabling [Akamai Image Manager](https://www.bigcommerce.com/blog/automatic-image-optimization/) for all Stencil storefronts. This CDN feature optimizes file sizes and performs device specific optimizations - sending the best format based on your user agent (for example, `.webp` images on Chrome) as well as actually sending a _smaller image_ in terms of dimensions if it detects you're on a device that can't display a larger image's full width. For example, a user agent that's 480px wide (according to Akamai's heuristics) would not receive a 4k image, but instead a smaller version closer to its screen size.

This solution was nice because it was "magical" and benefitted everyone immediately, but it caused a few problems:

- Sometimes images _need_ to be high res, for example product zoom images which are usually blown up larger than the screen size of the device in order to "zoom"
  - To fix this, we created the `imbypass=on` query string to selectively disable Akamai optimizations on a given image, and we use this for zoom images
- Edge cases where the browser was expecting specific image dimensions, only to receive something smaller (this issue was usually due to poorly-written CSS that wasn't up to spec with the latest responsive practices)
- Making a sizing decision based on "user agent screen resolution" is imperfect. For example, I might only need the image to be  25% of the page's width, so downsizing to 100% of its width is under-optimizing
- In general, Akamai Image Manager takes the control away from the theme developer, where BigCommerce prefers to empower developers to make their own decisions

#### How are we solving it?

In this PR, I've introduced a `srcset` component in `templates/components/common/responsive-img.html` that acts as a drop-in replacement for generating `` tags with the `image` objects you'd typically pass to `{{getImage}}`. Exposing this as a theme component (instead of a handlebars helper) allows it be configurable by theme developers, which may be preferable for something this complex. This component aims to absorb the complexity of generating a `srcset` in such a way that it's compatible with legacy browsers and also lazy-loadable.

Cornerstone already makes use of @aFarkas's [lazysizes](https://github.com/aFarkas/lazysizes) module for lazyloading. As the default behavior, I've implemented the pattern described [here](https://github.com/aFarkas/lazysizes#lqipblurry-image-placeholderblur-up-image-technique) to allow `srcset`-capable browsers to get a low-quality placeholder image while the full resolution image is downloaded in the background. Via theme configuration, you can also do standard lazyloading (with no placeholder) via [this pattern.](https://github.com/aFarkas/lazysizes#modern-transparent-srcset-pattern) Both of these methods have graceful backwards compatibility for browsers that do not support `srcset` such as IE11.

#### Things left to do

- Get feedback from the developer community (hence this PR!)
- LOTS of testing. This is a fundamental change to how Cornerstone works, and we need to make sure we're both accomplishing our goals and not breaking anything
  - Performance testing (browser) - are we making the page load slower in our test cases? If so, we need to re-evaluate the approach.
  - Performance testing (server) - we'll be generating a lot more versions of images using this approach, and we need to make sure we can support it on infrastructure side.
  - Legacy browser testing - we still need to support IE11, let's make sure the experience there is reasonable.
  - Visual regression testing - Need to test across many browsers and screen resolutions to make sure this doesn't cause crappy styling anywhere.
- ~Allow images to be upsized, not only downsized~ - DONE!
  - This is necessary so we can generate srcsets "in the blind" without know the sizes of the original images
  - As an alternative, we could do work elsewhere to understand the size of images at runtime and thus be more intelligent with srcset generation. This might be a larger project, so perhaps it will come later.
- Create a means of getting images which are optimized by Akamai in terms of file size and best-for-device file _type_, but NOT actually _resized_ by Akamai.
  - `srcset` is intended to allow the browser to take control of resizing, so Akamai's features will interfere with this when active
- ~Figure out the best image sizes to use.~ Done! 
  - I'm inspired by this post by @pmioni: [Applying srcset: choosing the right sizes for responsive images at different breakpoints](https://medium.com/hceverything/applying-srcset-choosing-the-right-sizes-for-responsive-images-at-different-breakpoints-a0433450a4a3)

#### Demo stores

https://srcset-test.mybigcommerce.com/
https://srcset-test-lqip.mybigcommerce.com/

#### Markup Before/After

Example is from the first product card image on https://cornerstone-light-demo.mybigcommerce.com/

*Cornerstone 3.4.4 (before `srcset`):*

Raw HTML in document:
``

After lazysizes evaluates:
``

*Cornerstone with `srcset`:*

Raw HTML in document:
``

After lazysizes evaluates:
`<img src="https://cdn11.bigcommerce.com/s-n0i50vy/images/stencil/640x640/products/86/286/ablebrewingsystem4_1024x1024__07155.1456436672.jpg?c=2" alt="" title="" data-sizes="auto" srcset="https://cdn11.bigcommerce.com/s-n0i50vy/images/stencil/1280x1280/products/86/286/ablebrewingsystem4_1024x1024__07155.1456436672.jpg?c=2 1280w, https://cdn11.bigcommerce.com/s-n0i50vy/images/stencil/160x160/products/86/286/ablebrewingsystem4_1024x1024__07155.1456436672.jpg?c=2 160w, https://cdn11.bigcommerce.com/s-n0i50vy/images/stencil/1920x1920/products/86/286/ablebrewingsystem4_1024x1024__07155.1456436672.jpg?c=2 1920w, https://cdn11.bigcommerce.com/s-n0i50vy/images/stencil/2560x2560/products/86/286/ablebrewingsystem4_1024x1024__07155.1456436672.jpg?c=2 2560w, https://cdn11.bigcommerce.com/s-n0i50vy/images/stencil/320x320/products/86/286/ablebrewingsystem4_1024x1024__07155.1456436672.jpg?c=2 320w, https://cdn11.bigcommerce.com/s-n0i50vy/images/stencil/640x640/products/86/286/ablebrewingsystem4_1024x1024__07155.1456436672.jpg?c=2 640w, https://cdn11.bigcommerce.com/s-n0i50vy/images/stencil/80x80/products/86/286/ablebrewingsystem4_1024x1024__07155.1456436672.jpg?c=2 80w, https://cdn11.bigcommerce.com/s-n0i50vy/images/stencil/960x960/products/86/286/ablebrewingsystem4_1024x1024__07155.1456436672.jpg?c=2 960w, " data-srcset="https://cdn11.bigcommerce.com/s-n0i50vy/images/stencil/1280x1280/products/86/286/ablebrewingsystem4_1024x1024__07155.1456436672.jpg?c=2 1280w, https://cdn11.bigcommerce.com/s-n0i50vy/images/stencil/160x160/products/86/286/ablebrewingsystem4_1024x1024__07155.1456436672.jpg?c=2 160w, https://cdn11.bigcommerce.com/s-n0i50vy/images/stencil/1920x1920/products/86/286/ablebrewingsystem4_1024x1024__07155.1456436672.jpg?c=2 1920w, https://cdn11.bigcommerce.com/s-n0i50vy/images/stencil/2560x2560/products/86/286/ablebrewingsystem4_1024x1024__07155.1456436672.jpg?c=2 2560w, https://cdn11.bigcommerce.com/s-n0i50vy/images/stencil/320x320/products/86/286/ablebrewingsystem4_1024x1024__07155.1456436672.jpg?c=2 320w, https://cdn11.bigcommerce.com/s-n0i50vy/images/stencil/640x640/products/86/286/ablebrewingsystem4_1024x1024__07155.1456436672.jpg?c=2 640w, https://cdn11.bigcommerce.com/s-n0i50vy/images/stencil/80x80/products/86/286/ablebrewingsystem4_1024x1024__07155.1456436672.jpg?c=2 80w, https://cdn11.bigcommerce.com/s-n0i50vy/images/stencil/960x960/products/86/286/ablebrewingsystem4_1024x1024__07155.1456436672.jpg?c=2 960w, " class="card-image lazyautosizes lazyloaded" loading="lazy" sizes="263px">`

#### Tickets / Documentation

- [STRF-3251](https://jira.bigcommerce.com/browse/STRF-3251)

ping @bigcommerce/storefront-team @bigcommerce/frontend @chrisboulton


templates/components/account/returns-list.html
```diff
@@ -5,29 +5,27 @@ <h3 class="account-heading">{{lang 'account.returns.heading' }}</h3>
...
+                    lazyload=../theme_settings.lazyload_mode
+                }}
             </div>
             <div class="account-product-body">
                 <div class="account-orderStatus">
                     {{#if status '===' 'ReturnStatusPending'}}
                         <h6 class="account-orderStatus-label">{{lang 'account.returns.status.pending'}}</h6>
-                    {{/if}}
-                    {{#if status '===' 'ReturnStatusReceived'}}
+                    {{else if status '===' 'ReturnStatusReceived'}}
```
bookernath: This was just a refactoring of this logic, since it was bothering me.

assets/js/theme/common/product-details.js
```diff
@@ -271,17 +271,38 @@ export default class ProductDetails {
...
+                this.context.themeSettings.srcset_size_640,
+                // Should match fallback image size used for the main
+                // product image in components/products/product-view.html
             );
 
+            /*
+                 Check the theme config looking for any keys containing `srcset_size_` and assuming them to
+                 be image size strings (e.g. "123x123") that will be used as sizes for srcset generation
+             */
+            const srcsets = Object.keys(this.context.themeSettings)
```
jsimm-guidance: this is brilliant. it's a shame themeSettings doesn't support an Array natively.

templates/components/common/responsive-img.html
```diff
@@ -43,6 +43,8 @@
 {{#unless lazyload '==' 'disabled'}}data-{{/unless}}srcset={{!--
     Generate the srcset string using all of the srcset_size_ properties in config.json. This will allow the browser to
     pick the best-sized image based on its own understanding of how it will render the page.
---}}"{{~#each theme_settings}}{{#if (occurrences @key "srcset_size_") ">=" 1}}{{getImage ../../image @key}} {{first (split this "x")}}w, {{/if}}{{/each}}"{{else if default_image}}src="{{cdn default_image}}"{{/if}}
+--}}"{{~#each theme_settings.srcset_sizes}}{{getImage ../../image @key}} {{first (split this "x")}}w{{#unless @last}}, {{/unless}}{{/each}}"
```
jsimm-guidance: this is the first time I've seen `{{~#` whats the `~` do?

jsimm-guidance:
This is good thinking. I would like to see some before/after output of the markup generated.

How does the product admin support the theme developer w/ multiple images? Will someone maintaining the product upload multiple resolutions each image, or is the assumption still one resolution per image?

templates/components/common/responsive-img.html
bookernath: It trims whitespace: https://stackoverflow.com/questions/15953293/handlebars-whitespace-control

bookernath:
Thanks @jsimm-guidance I'll add some before/after markup examples.

In terms of the merchant experience, our guidance to merchants would simply be to upload the highest-resolution imagery they have in the CP. From there, we can generate all the resized versions for srcset.

assets/js/theme/common/product-details.js
bookernath: Yep, this was just me getting it working with existing functionality... presumably we could add an array in the theme settings to make this cleaner. It wouldn't be editable in Theme Editor without more work on our end (since it's basically a new data type)

bookernath:
@jsimm-guidance example markup added

templates/layout/base.html
```diff
@@ -11,12 +11,21 @@
...
         {{ getFontsCollection }}
         {{{stylesheet '/assets/css/theme.css'}}}
+        {{#unless theme_settings.lazyload '==' 'disabled'}}
+            {{!-- Load Lazysizes script ASAP so images will appear --}}
+            <script>
+                {{!-- Only load visible elements until the onload event fires, after which preload nearby elements. --}}
+                window.lazySizesConfig = window.lazySizesConfig || {};
+                window.lazySizesConfig.loadMode = 1;
+            </script>
+            <script src="{{cdn 'lazysizes.min.js'}}"></script>
```
bookernath: @junedkazi @mattolson curious what you think about this.

If we wait until the bundle loads for lazysizes to activate, then no images will show up until the bundle executes.

Any scripts in `head.scripts` will be *effectively* blocking at that point, because they're preventing images from loading.

If we make lazysizes load before head.scripts, we have insurance against this - but I just wanted to get your thoughts on pulling this out of the bundle in general, since we don't typically do that. Maybe there's a more "webpack-ey" way to do this?

templates/components/common/responsive-img.html
```diff
@@ -0,0 +1,50 @@
+{{!--
+This file implements the lazysizes srcset pattern (as seen on
```
davidchin: I'm wondering if it's worth adding it to `paper-handlebars` so it can be used in other themes as well? 🤔

package.json
```diff
@@ -9,11 +9,10 @@
     "@babel/polyfill": "^7.0.0",
     "@bigcommerce/stencil-utils": "^4.1.1",
     "creditcards": "^3.0.1",
-    "easyzoom": "^2.5.0",
+    "easyzoom": "^2.5.2",
     "foundation-sites": "^5.5.3",
     "jquery": "^3.4.1",
     "jstree": "vakata/jstree",
-    "lazysizes": "4.1.2",
```
davidchin: Is there a reason why we no longer want to pull this dependency as a npm package?

templates/components/common/responsive-img.html
```diff
@@ -0,0 +1,50 @@
...
+`otherAttributes`:
+    Any other HTML attributes you want on the img tag, for example "height='100' width='100'"
+--}}
+<img {{#if image}}{{!-- Get the default image for legacy browsers that do not support srcset
+--}}src="{{getImage image (default fallback_size 'srcset_size_160')}}" alt="{{image.alt}}" title="{{image.alt}}" {{!--
+  Allow lazysizes to generate the 'sizes' attribute automatically --}}data-sizes="auto"
+  {{#unless lazyload '==' 'disabled'}}
+    {{#if lazyload '==' 'lazyload+lqip'}}
+    {{!-- If we want a LQIP, include it as the default srcset (will be overwritten by lazysizes) --}}
+    srcset="{{getImage image (default lqip_size 'srcset_size_80')}}"
```
davidchin: I notice we no longer show a loading spinner. Is it because we want to show LQIP image instead?

templates/components/common/responsive-img.html
davidchin: Actually, I just saw your PR https://github.com/bigcommerce/paper-handlebars/pull/42 so I guess that could work as well 😄

templates/components/common/responsive-img.html
bookernath: Yes, either a LQIP or nothing (transparent 1px gif as inline image).

We've gotten a lot of negative feedback on the spinner.

package.json
bookernath: It's because I wanted to move lazysizes to execute before head.scripts:

https://github.com/bigcommerce/cornerstone/pull/1507#discussion_r289417748

This improves perceptual performance of the page, especially when there are things in head.scripts that are either blocking or simply hog bandwidth that could go to downloading the bundle (therefore they are _effectively_ blocking on a slow connection).

davidchin: I see, you want to include `lazysizes` separately (from the main bundle / chunks). Perhaps you just need to specify a separate `entry` in your Webpack config to produce a separate file for that dependency. One of the benefits about keeping it as a `npm` dependency is that we can easily keep track of the version we're using and if there's security vulnerability with that version, Dependabot can warn us about it and suggest an automated upgrade.

templates/components/common/responsive-img.html
davidchin: Yep, it makes sense. Not a big fan of the gigantic loading spinner as well. 😄 I think as long as we give _some_ indication that something is loading in the background (either a low res image or a blank space), we're good.

package.json
bookernath: Thanks! I'll look into this.

assets/js/theme/common/product-details.js
bookernath: I've moved away from using theme config, and instead moved the logic into a new helper and stencil-utils method. We'll have a strong set of default sizes, but allow configuration by theme developers.

assets/scss/components/citadel/cards/_cards.scss
```diff
@@ -49,7 +49,7 @@
 .card-image {
     @include lazy-loaded-img;
     border: 0;
-    width: auto;
+    width: 100%;
```
bookernath: more explicit CSS sizing is helpful for lazysizes - this prevents page flickering.

assets/scss/components/foundation/lazyLoad/_lazyLoad.scss
```diff
@@ -1,7 +1,3 @@
-.lazyload, .lazyloading {
-    height:100%;
```
bookernath: removed this as it was causing flicker during lazyloading

schema.json
```diff
@@ -1313,7 +1341,7 @@
       },
       {
         "type": "checkbox",
-        "label": "Show &quot;Shop by Price&quot; in filters",
+        "label": "Show \"Shop by Price\" in filters",
```
bookernath: just fixing this while I was in here, I think it's easier to read.

schema.json
```diff
@@ -1735,7 +1763,7 @@
         "force_reload": true,
         "options": [
           {
-            "value": "1280x1280",
+            "value": "1920x1920",
```
bookernath: This isn't completely necessary to realize the value of responsive images, but I like having the zoom be a little zoomier :)

templates/components/account/order-contents.html
```diff
@@ -21,14 +21,14 @@ <h5 class="account-listShipping-title">{{lang 'account.orders.details.ship_to_mu
             </div>
             <figure class="account-product-figure">
                 {{#if type '===' 'giftcertificate'}}
-                    <img class="lazyload" data-sizes="auto" src="{{cdn 'img/loading.svg'}}" data-src="{{cdn ../../theme_settings.default_image_gift_certificate}}" alt="Gift Certificate" title="Gift Certificate">
+                    <img src="{{cdn ../../theme_settings.default_image_gift_certificate}}" alt="Gift Certificate" title="Gift Certificate">
```
bookernath: this image doesn't need lazyloading.

templates/components/amp/products/list-item.html
```diff
@@ -1,6 +1,6 @@
 <article class="listItem">
     <figure class="listItem-figure">
-        <amp-img src="{{getImage image 'productgallery_size' (cdn theme_settings.default_image_product)}}" width="{{first (split theme_settings.productgallery_size 'x')}}" height="{{last (split theme_settings.productgallery_size 'x')}}" layout="responsive" alt="{{image.alt}}"></amp-img>
+        <amp-img src="{{getImageSrcset image (cdn theme_settings.default_image_product) use_default_sizes=true}}" layout="responsive" alt="{{image.alt}}"></amp-img>
```
bookernath: `srcset` is polyfilled by AMP, so there's no need to provide a fallback `src` for amp-img, you can just use `srcset`.

templates/components/amp/products/product-view.html
```diff
@@ -9,6 +9,7 @@
     {{/each}}
 
     {{> components/amp/products/product-view-details}}
+    {{#if product.images}}
```
bookernath: Removing this check as AMP won't validate if there aren't product images.

templates/components/products/card.html
```diff
@@ -49,14 +50,11 @@
                     {{#if show_cart_action}}
                         {{#if has_options}}
                             <a href="{{url}}" data-event-type="product-click" class="button button--small card-figcaption-button" data-product-id="{{id}}">{{lang 'products.choose_options'}}</a>
-                        {{/if}}
-                        {{#if pre_order}}
+                        {{else if pre_order}}
```
bookernath: just cleaning this up.

templates/components/carousel.html
```diff
@@ -12,8 +12,22 @@
...
+                    lazyload='disabled'
+                }}
+            {{else}}
+                {{> components/common/responsive-img
+                    image=stencil_image
+                    class="heroCarousel-image"
+                    fallback_size='1280w'
+                    lazyload='lazyload'
+                }}
+            {{/if}}
```
bookernath: It was necessary to split this because I want the first carousel slide to never lazyload, but the 2nd and beyond slides to always lazyload. I couldn't figure out how to do that with a nested handlebars expression, and this was easier to read anyways.

templates/components/cart/content.html
```diff
@@ -12,9 +12,14 @@
...
                 <td class="cart-item-block cart-item-figure">
                     {{#if type '==' 'GiftCertificate'}}
-                        <img class="cart-item-fixed-image lazyload" data-sizes="auto" src="{{cdn 'img/loading.svg'}}" data-src="{{cdn ../../theme_settings.default_image_gift_certificate}}" alt="GiftCertificate" title="GiftCertificate">
+                        <img class="cart-item-fixed-image" src="{{cdn ../../theme_settings.default_image_gift_certificate}}" alt="GiftCertificate" title="GiftCertificate">
                     {{else}}
-                        <img class="cart-item-image lazyload" data-sizes="auto" src="{{cdn 'img/loading.svg'}}" data-src="{{getImage image 'productthumb_size' (cdn ../../theme_settings.default_image_product)}}" alt="{{image.alt}}" title="{{image.alt}}">
+                        {{> components/common/responsive-img
+                            image=image
+                            class="cart-item-image"
+                            lazyload='disabled'
```
bookernath: I think once you're far enough down the purchase funnel to be looking at the cart page, it's better to show images ASAP than to optimize the load speed of the page further. There's a decent chance you already have the image in cache as well, since you viewed the product page.

templates/components/products/options/product-list.html
```diff
@@ -25,7 +25,11 @@
                 <li class="productOptions-list-item" data-product-attribute-value="{{id}}">
                     {{#if image}}
                         <figure class="productOptions-list-item-figure">
-                            <img class="productOptions-list-item-image lazyload" data-sizes="auto" src="{{cdn 'img/loading.svg'}}" data-src="{{getImage image 'productthumb_size'}}" alt="{{image.alt}}" title="{{image.alt}}">
+                            {{> components/common/responsive-img
+                                image=image
+                                class="productOptions-list-item-image"
+                                lazyload='lazyload+lqip'
```
bookernath: I'm forcing lazyloading on these because the entire structure of the product list element gets significantly modified by JS after the page loads, so there's no need to load these immediately - it's better to wait for the sizing to settle so the browser can pick the right size.

templates/layout/base.html
```diff
@@ -14,9 +14,16 @@
             {{!-- Change document class from no-js to js so we can detect this in css --}}
             document.documentElement.className = document.documentElement.className.replace('no-js', 'js');
         </script>
+        {{!-- Load Lazysizes script ASAP so images will appear --}}
+        <script>
+            {{!-- Only load visible elements until the onload event fires, after which preload nearby elements. --}}
+            window.lazySizesConfig = window.lazySizesConfig || {};
+            window.lazySizesConfig.loadMode = 1;
+        </script>
+        <script async src="{{cdn 'assets/dist/theme-bundle.head_async.js'}}"></script>
```
bookernath: Whoa, javascript in the head?!?!

Yes - we need this to make sure images actually show up on the page, so this few KB file is mission-critical.

I've marked it as `async` so it won't block render, but we want it to execute as soon as it has downloaded.

templates/pages/amp/category.html
```diff
@@ -20,7 +20,7 @@
     {{> components/amp/common/header }}
     {{#if category.image}}
         <div class="categoryView-img">
-            <amp-img src="{{getImage category.image 'zoom_size'}}" width="300" height="150" layout="responsive" alt="{{category.name}}"></amp-img>
+            <amp-img srcset="{{getImageSrcset category.image use_default_sizes=true}}" layout="responsive" alt="{{category.name}}"></amp-img>
```
bookernath: deleting dimensions as they're not necessary with `layout="responsive"` and a `srcset` 👍

templates/pages/amp/product.html
```diff
@@ -12,7 +12,9 @@
 {{/partial}}
 {{#partial "amp-scripts"}}
     <script async custom-element="amp-accordion" src="https://cdn.ampproject.org/v0/amp-accordion-0.1.js"></script>
-    <script async custom-element="amp-carousel" src="https://cdn.ampproject.org/v0/amp-carousel-0.1.js"></script>
+    {{#any product.videos product.images}}
```
bookernath: fixing another AMP validation error

templates/pages/brand.html
```diff
@@ -17,7 +17,10 @@
 {{> components/common/breadcrumbs breadcrumbs=breadcrumbs}}
 {{#if brand.image}}
     <div class="brand-image-container">
-        <img class="lazyload" data-sizes="auto" src="{{cdn 'img/loading.svg'}}" data-src="{{getImage brand.image 'thumb_size'}}">
+        {{> components/common/responsive-img
+            image=brand.image
+            lazyload='disabled'
```
bookernath: Brand image is critical above-the-fold content on the brand page, so not lazyloading this.

webpack.common.js
```diff
@@ -10,6 +10,7 @@ module.exports = {
     context: __dirname,
     entry: {
         main: './assets/js/app.js',
+        head_async: ['lazysizes'],
```
bookernath: Loading lazysizes module in the head script. This file should only be used for mission-critical head JS.

mattolson:
Do we have a sense of how much this will inflate the overall HTML document size for the average site?

bookernath:
@mattolson  About 10kb for a PLP like a category page.

Here's Cornerstone 3.5.1 on a category page with 12 products:

image

Here's this branch:

image

As you can see, it gzips quite well as to make the over-the-wire difference inconsequential.

templates/components/products/product-view.html
```diff
@@ -150,24 +150,31 @@ <h2 class="productView-brand"{{#if schema}} itemprop="brand" itemscope itemtype=
...
     <section class="productView-images" data-image-gallery>
+        {{!--
+            Note that these image sizes are coupled to image sizes used in /assets/js/theme/common/product-details.js
+            for variant/rule image replacement
+        --}}
         <figure class="productView-image"
                 data-image-gallery-main
                 {{#if product.main_image}}
-                data-zoom-image="{{getImage product.main_image 'zoom_size' (cdn theme_settings.default_image_product)}}"
+                data-zoom-image="{{getImageSrcset product.main_image (cdn theme_settings.default_image_product) 1x="1280w" }}"
```
mattolson: Shouldn't we continue to use the `zoom_size` defined in the theme settings rather than hard code it here?

templates/components/amp/products/list-item.html
mattolson: This is inconsistent with the usage in `templates/components/amp/products/card.html`. That uses the attribute `srcset` where this one uses `src`. Which is correct?

assets/js/theme/common/product-details.js
```diff
@@ -269,19 +269,27 @@ export default class ProductDetails {
 
     showProductImage(image) {
         if (_.isPlainObject(image)) {
-            const zoomImageUrl = utils.tools.image.getSrc(
+            const zoomImageUrl = utils.tools.image.getSrcset(
```
mattolson: We are relying on a trick in getSrcset where it returns a string rather than an array if only one size is passed in. Perhaps we should have a comment to that effect here.

templates/components/common/responsive-img.html
```diff
@@ -0,0 +1,51 @@
...
+    - If set to 'lazyload+lqip', will load a low-res image which will be replaced by lazysizes - ensuring that SOME
+    content is shown immediately on page load. Without LQIP, a transparent gif will be used instead, which reduces
+    the number of requests needed to load the page, making it even faster.
+    - If set to 'lazyload', the browser will not load the full-size image at first, instead the lazysizes plugin will
+        load it progressively after page load. This improves the load speed of the page, but it should be disabled for
+        high-priority above-the-fold images.
+    - If set to 'disabled', will not attempt to lazyload the image and instead tell the browser to load it immediately.
+        Use this for high-priority images that are above the fold.
+
+`fallback_size`:
```
mattolson: You should document `lqip_size` and `default_image` as well.

templates/components/common/responsive-img.html
```diff
@@ -0,0 +1,51 @@
...
+    {{/if}}
+{{/unless}}
+{{!-- If lazyloading is disabled, just use "srcset". If it is not disabled, use "data-srcset" so lazysizes will use
+      this after page load --}}
+{{!--
+Generate the srcset string using the default sizes defined in the helper. This will allow the browser to
+pick the best-sized image based on its own understanding of how it will render the page.
+--}}
+{{#unless lazyload '==' 'disabled'}}data-{{/unless}}srcset="{{getImageSrcset image use_default_sizes=true }}"
+{{else if default_image}}src="{{cdn default_image}}"{{/if}}
```
mattolson: Shouldn't we apply all the same logic above for default_image as well? The way I read it, we should use `image` if it is present, otherwise `default_image`, but all the other logic around sizing, srcset, lazy loading, etc, should be the same. One way to handle that might be to push that up a layer and just pass in the right `image` into this component. i.e. `image=(or image theme_settings.default_image_product)` (don't know if that works, just a thought)

bookernath:
@mattolson I didn't get rid of any of the sizes, and in most cases they still affect the sizing of the images via CSS; they're simply decoupled from the pixel dimensions of the image, because with srcset, that's less important. We're not going for a perfect match between CSS sizing and pixel sizing with this methodology.

templates/components/common/responsive-img.html
bookernath: Assumption is a default image is much more likely to already be in the browser cache, and if it isn't, it's probably tiny (ours is a few KB). So there's no point loading a 1kb LQIP for a 4kb image.

Also, those images can't be resized in the same way product images can (to my knowledge) as they're just theme assets.

templates/components/common/responsive-img.html
bookernath: Good callout, will add that.

templates/components/amp/products/list-item.html
bookernath: This file was a miss, I've fixed it.

bookernath:
In the interest of reducing the changeset (and thus the risk), I've removed any updates to AMP templates.

templates/components/products/product-view.html
bookernath: Updated here and in product-details.js.

templates/components/common/responsive-img.html
```diff
@@ -0,0 +1,61 @@
...
+    srcset="data:image/gif;base64,R0lGODlhAQABAIAAAP///wAAACH5BAEAAAAALAAAAAABAAEAAAICRAEAOw=="
+    {{/if}}
+{{/unless}}
+{{!-- If lazyloading is disabled, just use "srcset". If it is not disabled, use "data-srcset" so lazysizes will use
+      this after page load --}}
+{{!--
+Generate the srcset string using the default sizes defined in the helper. This will allow the browser to
+pick the best-sized image based on its own understanding of how it will render the page.
+--}}
+{{#unless lazyload '==' 'disabled'}}data-{{/unless}}srcset="{{getImageSrcset image use_default_sizes=true }}"
```
mattolson: Is the thinking that in the future, we'd give theme developers a hook here to define their own sizes?

templates/pages/category.html
```diff
@@ -18,7 +18,11 @@
 
 {{> components/common/breadcrumbs breadcrumbs=breadcrumbs}}
 {{#if category.image}}
-    <img class="lazyload" data-sizes="auto" src="{{cdn 'img/loading.svg'}}" data-src="{{getImage category.image 'zoom_size'}}">
+    {{> components/common/responsive-img
+        image=category.image
+        fallback_size='1280w'
```
mattolson: Shouldn't this be defined by `zoom_size`?

templates/pages/brands.html
```diff
@@ -16,7 +16,13 @@ <h1 class="page-heading">{{lang 'brand.label'}}</h1>
                     <figure class="card-figure">
                         <a href="{{url}}">
                             <div class="card-img-container">
-                                <img class="card-image lazyload" data-sizes="auto" src="{{cdn 'img/loading.svg'}}" data-src="{{getImage image 'brand_size' (cdn ../theme_settings.default_image_brand)}}" alt="{{name}}" title="{{name}}">
+                                {{> components/common/responsive-img
+                                    image=image
+                                    class="card-image"
+                                    fallback_size='320w'
```
mattolson: Shouldn't this be defined by `brand_size`?

templates/components/products/modals/writeReview.html
```diff
@@ -12,7 +12,12 @@ <h2 class="modal-header-title">{{lang 'products.reviews.write_a_review'}}</h2>
     <div class="modal-body">
         <div class="writeReview-productDetails">
             <div class="writeReview-productImage-container">
-                <img class="lazyload" data-sizes="auto" src="{{cdn 'img/loading.svg'}}" data-src="{{getImage product.main_image 'product_size' (cdn theme_settings.default_image_product)}}">
+                {{> components/common/responsive-img
+                    image=product.main_image
+                    fallback_size='320w'
```
mattolson: Shouldn't this be defined by `product_size`?

templates/components/products/list-item.html
```diff
@@ -4,7 +4,13 @@
     <article class="listItem">
 {{/if}}
     <figure class="listItem-figure">
-        <img class="listItem-image lazyload" data-sizes="auto" src="{{cdn 'img/loading.svg'}}" data-src="{{getImage image 'productgallery_size' (cdn theme_settings.default_image_product)}}" alt="{{image.alt}}" title="{{image.alt}}">
+        {{> components/common/responsive-img
+            image=image
+            class="listItem-image"
+            fallback_size='320w'
```
mattolson: Shouldn't this be defined by `productgallery_size`?

templates/components/products/card.html
```diff
@@ -16,17 +16,18 @@
...
             {{/if}}
         {{/or}}
-        {{#if demo}}
+        {{#if demo}}<a href="{{url}}" {{#if settings.data_tag_enabled}} data-event-type="product-click" {{/if}}>{{/if}}
             <div class="card-img-container">
-                <img class="card-image lazyload" data-sizes="auto" src="{{cdn 'img/loading.svg'}}" data-src="{{getImage image 'productgallery_size' (cdn theme_settings.default_image_product)}}" alt="{{image.alt}}" title="{{image.alt}}">
+                {{> components/common/responsive-img
+                    image=image
+                    class="card-image"
+                    fallback_size='320w'
```
mattolson: Shouldn't this be defined by `productgallery_size`?

templates/components/common/cart-preview.html
```diff
@@ -5,9 +5,14 @@
                 <li class="previewCartItem">
                     <div class="previewCartItem-image">
                         {{#if type '==' 'GiftCertificate'}}
-                            <img class="lazyload" data-sizes="auto" src="{{cdn 'img/loading.svg'}}" data-src="{{cdn ../../theme_settings.default_image_gift_certificate}}" alt="GiftCertificate" title="GiftCertificate">
+                            <img src="{{cdn ../../theme_settings.default_image_gift_certificate}}" alt="GiftCertificate" title="GiftCertificate">
                         {{else}}
-                            <img class="lazyload" data-sizes="auto" src="{{cdn 'img/loading.svg'}}" data-src="{{getImage image 'productthumb_size' (cdn ../../theme_settings.default_image_product)}}" alt="{{image.alt}}" title="{{image.alt}}">
+                            {{> components/common/responsive-img
+                                image=image
+                                fallback_size='160w'
```
mattolson: Shouldn't this be defined by `productthumb_size`?

templates/components/common/responsive-img.html
bookernath: We could; the helper already supports it.

My best-guess assumption going into this is if you're advanced enough to want to tweak your srcset sizes, you're probably on a custom theme anyways.

templates/pages/category.html
bookernath: Yep, I'll make these updates. I think my thinking at the time is that it seemed like we were sort of overloading those sizes (based on what they say they do in Theme Editor), but I can certainly put them back for consistency. I'll do that now and resolve these.

templates/components/products/product-view.html
```diff
@@ -183,11 +190,16 @@ <h2 class="productView-brand"{{#if schema}} itemprop="brand" itemscope itemtype=
                 <li class="productView-thumbnail">
                     <a
                         class="productView-thumbnail-link"
-                        href="{{getImage this 'product_size' (cdn ../theme_settings.default_image_product)}}"
+                        href="{{getImageSrcset this (cdn ../theme_settings.default_image_product) 1x=../theme_settings.zoom_size}}"
```
bookernath: href in case you open in new tab should return a large image.

VH-WRK:
@bookernath This was a brilliant idea!
I'm struggling to implement this correctly for my store. I'm using an edited version of cornerstone and the problem lies in our images taking on different aspect ratios because we sell books and they come in landscape and portrait. 
What happens is our portrait images get stretched as a result of `width: 100%;` change in assets/scss/components/citadel/cards/_cards.scss
For .card-image we had both width and height set to auto and max-width and max-height set to 100%. This allowed the src attribute create the correct size and aspect ratio and the image would fill the image container.
Now the image loaded is the default lqip_size of width 80px. Is there away around this?

bookernath:
@VH-WRK that was a bug with this PR actually. It's been fixed in https://github.com/bigcommerce/cornerstone/pull/1559 if you want to pull in that fix!

PR #1512, opened 2019-05-31
chore(storefront): STRF-6780 clean up returns-list.html conditional logic

bookernath:
#### What?

This should be else if as this is basically a switch statement.


bookernath:
@junedkazi ♻️

PR #1513, opened 2019-05-31
Clean up whitespace in layout/base.html

bookernath:
The <head> of the storefront contains a bunch of empty stuff when you llook at it in prod - this removes some of that which makes it a little easier to develop on Cornerstone.

#### Before

image

#### After

image

### DIFF
--- a/templates/layout/base.html
+++ b/templates/layout/base.html
@@ -11,7 +11,7 @@
         <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1">
 
         <script>
-            // Change document class from no-js to js so we can detect this in css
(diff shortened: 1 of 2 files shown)


templates/layout/base.html
```diff
@@ -11,7 +11,7 @@
         <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1">
 
         <script>
-            // Change document class from no-js to js so we can detect this in css
+            {{!-- Change document class from no-js to js so we can detect this in css --}}
```
lord2800: 🙇 for this and the js --> handlebars comment below.

PR #1522, opened 2019-06-13
Clean up share.html conditional logic

sacr3dc0w:
**What?**

This should be else if as this is basically a switch statement.

Similar reason for logic cleanup in #1512.


bookernath:
Thanks!

PR #1525, opened 2019-06-23
STRF-3591 add return instructions in return-saved.html

bc-williamkwon:
#### What?
Previously return instructions did not appear after submitting a return request and one would need to edit the template file in return-saved.html to display return instructions even if they had configured returns instructions in the control panel.  This pull request is to fix that issue.

#### Tickets / Documentation

- [Link 1](https://jira.bigcommerce.com/browse/STRF-3591)

image
image

### DIFF
--- a/templates/pages/account/return-saved.html
+++ b/templates/pages/account/return-saved.html
@@ -1,12 +1,16 @@
 {{#partial "page"}}
 <div class="account">
     {{> components/common/breadcrumbs breadcrumbs=breadcrumbs}}
-    <h2 class="page-heading">{{lang 'account.returns.heading' }}</h2>
+    <h2 class="page-heading">{{lang 'account.returns.heading'}}</h2>
     {{> components/account/navigation account_page='returns'}}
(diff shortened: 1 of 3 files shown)


templates/pages/account/return-saved.html
```diff
@@ -7,6 +7,10 @@ <h2 class="page-heading">{{lang 'account.returns.heading' }}</h2>
         <p>
             {{lang 'account.returns.successful'}}
         </p>
+        {{#if this.instructions}}
+            <h3>Return Instructions:</h3>
```
bookernath: Can you make this a language string in https://github.com/bigcommerce/cornerstone/blob/master/lang/en.json ? This will make it easier to customize and translate; we should avoid hardcoding human-readable language in the theme HTML.

bc-williamkwon: Ok, will do.  Thanks for pointing that out.

bc-williamkwon:
rebased again.

PR #1526, opened 2019-06-24
STRF-5665 remove google plus

bc-williamkwon:
#### What?

Google plus is deprecated so we should remove it.  This pull request is part of that that removal.

#### Tickets / Documentation

- [Link 1](https://jira.bigcommerce.com/browse/STRF-5665)

PR #1530, opened 2019-06-27
STRF-3382 display facebook like button when enabled

bc-williamkwon:
#### What?

Display like button when it is enabled through control panel.  This pull request is to fix a bug where the facebook like button would not display when enabled.

#### Tickets / Documentation

- [Link 1](https://jira.bigcommerce.com/browse/STRF-3382)

Screen Shot 2019-06-26 at 9 54 27 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Add return instructions in return-saved.html [#1525](https://github.com/bigcommerce/cornerstone/pull/1525)
 - Remove Google Plus [#1526](https://github.com/bigcommerce/cornerstone/pull/1526)
 - Fix broken conditional statement in share.html [#1533](https://github.com/bigcommerce/cornerstone/pull/1533)
+- Enable Facebook like button to be displayed on product page if enabled [#1530](https://github.com/bigcommerce/cornerstone/pull/1530)
 
 ## 3.5.1 (2019-06-24)
 - Fix conditional logic in share.html [#1522](https://github.com/bigcommerce/cornerstone/pull/1522)
--- a/templates/components/common/share.html
+++ b/templates/components/common/share.html
@@ -33,10 +33,6 @@
                             <svg>
                                 <use xlink:href="#icon-pinterest"/>
                             </svg>
-                        {{else if service '===' 'facebook_like'}}
-                            <svg>
-                                <use xlink:href="#icon-facebook_like"/>
-                            </svg>
                         {{/if}}
                     </a>
                 </li>
@@ -52,4 +48,14 @@
             });
         </script>
     </div>
+{{/if}}
+{{#if settings.facebook_like_button.enabled}}
+    <iframe 
+        class="facebook{{#if settings.facebook_like_button.verb '===' 'recommend'}}Recommend{{else}}Like{{/if}}Btn" 
+        style="border:none; overflow:hidden; margin: 0; padding: 0; position:absolute" 
+        src="https://www.facebook.com/plugins/like.php?href={{settings.facebook_like_button.href}}&amp;layout=button_count{{#if settings.facebook_like_button.verb '===' 'recommend'}}&amp;action=recommend{{/if}}&amp;colorscheme=light&amp;height=20" 
+        scrolling="no" 
+        frameborder="0" 
+        allowTransparency="true">
+    </iframe>
 {{/if}}


templates/components/products/product-view.html
```diff
@@ -230,6 +230,7 @@ <h2 class="productView-brand"{{#if schema}} itemprop="brand" itemscope itemtype=
             {{/if}}
         </div>
         {{> components/common/share}}
+        {{#if settings.facebook_like_button}}{{{settings.facebook_like_button}}}{{/if}}
```
mattolson: Should this go in the `common/share` component? Do we only want it to show on this page, or anywhere we have the share buttons?

templates/components/amp/common/share.html
```diff
@@ -3,11 +3,7 @@
         {{#each settings.add_this.buttons}}
             {{#all (if service '!==' 'google') (if service '!==' 'print') (if service '!==' 'facebook')}}
                 <li class="socialLinks-item socialLinks-item--{{service}}">
-                  {{#if service '===' 'google_plusone'}}
-                        <amp-social-share class="amp-social-share" data-vars-product-link="{{url}}" data-vars-product-id="{{id}}" data-vars-product-name="{{name}}" data-vars-social-type="gplus" type="gplus"></amp-social-share>
-                    {{else}}
-                        <amp-social-share class="amp-social-share" data-vars-product-link="{{url}}" data-vars-product-id="{{id}}" data-vars-product-name="{{name}}" data-vars-social-type="{{service}}" type="{{service}}"></amp-social-share>
-                  {{/if}}
+                    <amp-social-share class="amp-social-share" data-vars-product-link="{{url}}" data-vars-product-id="{{id}}" data-vars-product-name="{{name}}" data-vars-social-type="{{service}}" type="{{service}}"></amp-social-share>
```
mattolson: It looks like you are blending this change with https://github.com/bigcommerce/cornerstone/pull/1526. Can you separate them?

templates/components/products/product-view.html
bc-williamkwon: Facebook like button isn't the same as the other share buttons it's an iframe and not an icon for one.  I think it's good to keep them separate.

bc-williamkwon:
Just moved the markup into its own component where it is being generated on client side.  Thanks for the suggestion.

templates/components/common/facebook-like-button.html
```diff
@@ -0,0 +1,10 @@
+{{#if settings.facebook_like_button.enabled}}
```
mattolson: You should check for the existence of `settings.facebook_like_button` before using `enabled`.

bc-williamkwon: fixed.

templates/components/common/facebook-like-button.html
```diff
@@ -0,0 +1,12 @@
...
+        <iframe 
+            class="facebook{{#if settings.facebook_like_button.verb '===' 'recommend'}}Recommend{{else}}Like{{/if}}Btn" 
+            style="border:none; overflow:hidden; margin: 0; padding: 0; position:absolute" 
+            src="http://www.facebook.com/plugins/like.php?href={{settings.facebook_like_button.href}}&amp;layout=button_count{{#if settings.facebook_like_button.verb '===' 'recommend'}}&amp;action=recommend{{/if}}&amp;colorscheme=light&amp;height=20" 
+            scrolling="no" 
+            frameborder="0" 
+            allowTransparency="true">
+        </iframe>
+    {{/if}}
+{{/if}}
```
junedkazi: EOL is missing.

templates/components/products/product-view.html
```diff
@@ -230,6 +230,7 @@ <h2 class="productView-brand"{{#if schema}} itemprop="brand" itemscope itemtype=
             {{/if}}
         </div>
         {{> components/common/share}}
+        {{> components/common/facebook-like-button}}
```
junedkazi: Is there a reason why we can't pull this component inside `{{> components/common/share}}`

bc-williamkwon: I thought it would be better to separate it out since everything else is a svg instead of an iframe and they are about sharing whereas this is about liking but no real reason.

templates/components/common/facebook-like-button.html
bc-williamkwon: fixed.

templates/components/products/product-view.html
junedkazi: @bookernath do you have an opinion on this one ? If now this PR is good to go.

templates/components/common/facebook-like-button.html
```diff
@@ -0,0 +1,12 @@
+{{#if settings.facebook_like_button}}
+    {{#if settings.facebook_like_button.enabled}}
```
bookernath: 🍹 could switch this to `{{and}}` or just drop the outside `{{if}}`, if the object is not defined handlebars will treat that as falsy.

templates/components/products/product-view.html
bookernath: I like the idea of putting this new component _inside_ the share component.

If I were a theme developer that wanted to move my share icons around on the page, I'd probably expect Facebook to come along for the ride, so I think that would be more convenient.

bc-williamkwon: Made the requested changes @bookernath

PR #1537, opened 2019-07-12
Remove nanobar

bookernath:
#### What?

This loading bar just makes the site seem slow, removing it.


bc-williamkwon:
Didn't even know we had nanobar.  Seems like a good idea to remove.

mattolson:
Seems fine to me. However, the diff in package-lock is concerning. Can you try with Node 8 and npm 6.4.1?

oggede:
Just some feedback on this change.

I have a website with a lot of images that have text in them. See here https://www.nameitlabels.com/

As you can see, when you have images with a lot of text, the nanobar is a better option rather than the pre-blured images. When images load on a lower quality, the text is blurred. I have found that the nanobar is more pleasing.

For 95% of people, I'm sure the blurred option is better than the nanobar. However, there are positives and negatives to all systems and changes. One way doesn't always work for everyone.

I recommend leaving the nanobar as an option for people to re-enable or disable.

Cheers.

PR #1539, opened 2019-07-17
chore(storefront): STRF-7084 remove RSS links from head

bookernath:
#### What?

This variable does not return anything, so it should not be here.

There is no head.rsslinks in the context.

#### Tickets / Documentation

- [STRF-7084](https://jira.bigcommerce.com/browse/STRF-7084)


bookernath:
ping @bigcommerce/storefront-team

bookernath:
@junedkazi some of those things still work, and some of them don't. We can take them one at a time.

I have high confidence in the change within this PR, because as far as I can tell, this variable has _never_ done anything.

PR #1542, opened 2019-07-19
Add Amazon Pay and Google Pay to payment icons in footer

AlbertJBurton:
#### What?

Merchants should have the ability to advertise acceptance of payments via alternate means other than credit cards or PayPal. Two popular payment methods available through BigCommerce are Amazon Pay and Google Pay. This change adds the icons for these payment methods and allows customization of the theme to display these icons in the footer if desired.

#### Tickets / Documentation

N/A

New Amazon Pay and Google Pay icons with existing payment icons

image

New options in the Header & Footer customization menu

image


config.json
```diff
@@ -43,284 +43,286 @@
...
-    "supported_card_type_icons": [
-      "american_express",
-      "diners",
-      "discover",
-      "mastercard",
-      "visa"
-    ]
-  },
+    "settings": {
+        "hide_breadcrumbs": false,
```
bookernath: @AlbertJBurton it looks like your editor converted spaces to tabs or something like that here - can you fix that to minimize this changeset to just the functional bits?

schema.json
```diff
@@ -489,426 +489,438 @@
...
+                }
+            ]
+        },
+        {
+            "type": "select",
+            "label": "Logo font size",
+            "id": "logo_fontSize",
+            "options": [
+                {
+                    "value": 14,
```
bookernath: Same thing here - we should limit this changeset to just the new feature.

config.json
AlbertJBurton: Updated the config.json file with an editor that doesn't reformat the spacing and tabs in the file.

schema.json
AlbertJBurton: Updated the schema.json file with an editor that doesn't reformat the spacing and tabs in the file.

bookernath:
@AlbertJBurton oops, it looks like schema.json is minified now - would you mind fixing that?

bookernath:
This looks good now. Thanks for your contribution!

ebuni123:
Adding the Apple Pay Logo to the payment types would also be beneficial.

PR #1545, opened 2019-07-19
Prevent heading from displaying when no price ranges are available

AlbertJBurton:
#### What?

It is possible for a category to have an insufficient number of items to generate price ranges. If the theme is customized to display the "Shop by Price" sidebar and the category does not generate a price range, only the heading is displayed. This change ensures there are price ranges available before displaying the heading and price ranges. 

#### Tickets / Documentation

N/A

Example of heading displayed with no price ranges.

image

### DIFF
--- a/templates/components/category/shop-by-price.html
+++ b/templates/components/category/shop-by-price.html
@@ -1,4 +1,5 @@
 {{#if theme_settings.shop_by_price_visibility}}
+  {{#if shop_by_price}}
     <div class="sidebarBlock">
         <h5 class="sidebarBlock-heading">{{lang 'category.shop_by_price'}}</h5>
         <ul class="navList">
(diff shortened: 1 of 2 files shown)


templates/components/category/shop-by-price.html
```diff
@@ -1,4 +1,5 @@
 {{#if theme_settings.shop_by_price_visibility}}
+  {{#if shop_by_price}}
```
bookernath: you could consider combining this into one check which would be a little cleaner:

`{{#and theme_settings.shop_by_price_visibility shop_by_price}}`

AlbertJBurton: Done...

bookernath:
This looks good, can you rebase to resolve the conflict with CHANGELOG.md?

Thanks again for your contribution.

PR #1550, opened 2019-07-24
Remove "new products" carousel, use same display used for featured products and best sellers

bookernath:
#### What?

Standardize "New Products" to use the same display treatment as other homepage panels, to reduce complexity of the home page and fix a few bugs.

### DIFF
--- a/templates/components/products/featured.html
+++ b/templates/components/products/featured.html
@@ -3,7 +3,7 @@
 <ul class="productGrid productGrid--maxCol{{ columns }}" data-product-type="featured" {{#if settings.data_tag_enabled}} data-list-name="Featured Products" {{/if}}>
     {{#each products}}
         <li class="product">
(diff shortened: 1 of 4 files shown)


templates/components/products/featured.html
```diff
@@ -3,7 +3,7 @@ <h2 class="page-heading">{{lang 'products.featured' }}</h2>
 <ul class="productGrid productGrid--maxCol{{ columns }}" data-product-type="featured" {{#if settings.data_tag_enabled}} data-list-name="Featured Products" {{/if}}>
     {{#each products}}
         <li class="product">
-                {{>components/products/card settings=../settings theme_settings=../theme_settings customer=../customer event="list" position=(add @index 1)}}
```
bookernath: just making spacing identical across these files.

bookernath:
This may not be necessary if https://github.com/bigcommerce/cornerstone/pull/1458 solves the same problems.

PR #1551, opened 2019-07-25
Fix bug where image is not clickable in product card

bookernath:
Removing unneeded `demo` conditional logic which is no longer used. Fixes a bug which was preventing product card images from being clicked.


bookernath:
@bigcommerce/storefront-team turns out the `demo` key isn't even specified anywhere, and I can't find any record of it ever being real. So I've just removed it.

PR #1552, opened 2019-07-25
mattcoy-arcticleaf's patch/product carousel responsiveness

bookernath:
#### What?

Pulling in @mattcoy-arcticleaf 's product carousel fix from https://github.com/bigcommerce/cornerstone/pull/1458 together with a slick carousel iOS/Safari fix from https://github.com/kenwheeler/slick/issues/2834#issuecomment-389590661.

Fixes styling issues in several spots where slick carousel is used.

See original PR.

### DIFF
--- a/templates/components/products/carousel.html
+++ b/templates/components/products/carousel.html
@@ -4,13 +4,29 @@
         "dots": true,
         "infinite": false,
         "mobileFirst": true,
-        "slidesToShow": {{ columns }},
-        "slidesToScroll": 3
+        "slidesToShow": 2,
+        "slidesToScroll": 2,
+        "responsive": [
+            {
+                "breakpoint": 800,
+                "settings": {
+                    "slidesToShow": {{ columns }},
+                    "slidesToScroll": 3
+                }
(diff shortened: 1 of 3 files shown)


bookernath:
I've confirmed that this does not freeze on iOS safari in Store Design.

templates/components/products/carousel.html
```diff
@@ -4,13 +4,29 @@
...
         "infinite": false,
         "mobileFirst": true,
-        "slidesToShow": {{ columns }},
-        "slidesToScroll": 3
+        "slidesToShow": 2,
+        "slidesToScroll": 2,
+        "responsive": [
+            {
+                "breakpoint": 800,
+                "settings": {
```
ncheikh: Are these the standard breakpoints we use platform wide?

ncheikh: Confirmed these are in alignment

bookernath: Yes, 800px and 550px are breakpoints for cornerstone.

PR #1553, opened 2019-07-26
Put nanobar back

bookernath:
#### What?

Add back nanobar because we rely on it for tests.

But hide it with CSS to still get the perceptual benefit we were going for in https://github.com/bigcommerce/cornerstone/pull/1537

We can fully remove nanobar later.

ping @bc-sandeep


assets/scss/components/vendor/nanobar/_nanobar.scss
```diff
@@ -0,0 +1,20 @@
+// =============================================================================
+// NANOBAR - AJAX LOADING PROGRESS
+// =============================================================================
+
+.nanobar {
+    display: none;
```
bookernath: this is a straight revert except for this change to hide the bar.

PR #1559, opened 2019-08-13
Fix tall images on PDPs and PLPs

bookernath:
#### What?

Cornerstone 4.0 introduced a bug in which images which are particularly tall (portrait) are stretched on PDPs and PLPs:

image

This fixes that using `object-fit: contain`.

#### Tickets / Documentation

image
image

### DIFF
--- a/assets/js/theme/global/object-fit-polyfill.js
+++ b/assets/js/theme/global/object-fit-polyfill.js
@@ -1,0 +1,5 @@
+import objectFitImages from 'object-fit-images';
+
+export default function () {
+    objectFitImages();
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "bigcommerce-cornerstone",
-  "version": "4.0.0-rc.2",
+  "version": "4.0.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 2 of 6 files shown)


bookernath:
Looks like this doesn't work on IE11 and Edge. Need to look for a better solution.

bookernath:
@mattolson ♻️

bookernath:
I've added this polyfill for IE11: https://github.com/fregante/object-fit-images

We can remove it when we no longer need to support IE11

bookernath:
It's 3kb before gzip: https://github.com/fregante/object-fit-images/blob/master/dist/ofi.min.js

bookernath:
@mattolson I've added a conditional to make it only execute for IE11. Let me know if you think we should further optimize.

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "bigcommerce-cornerstone",
-  "version": "4.0.0-rc.2",
+  "version": "4.0.0",
```
junedkazi: Something is wrong here .. your PR shouldn't have this change.

junedkazi: Please ignore this comment. I have created a follow up ticket to fix this issue.

assets/js/theme/global/object-fit-polyfill.js
```diff
@@ -0,0 +1,8 @@
+import objectFitImages from 'object-fit-images';
+
+export default function () {
+    /* Polyfill object-fit for IE11 only */
+    if (navigator.userAgent.match(/Trident.*rv\:11\./)) {
```
mattolson: Most likely this kind of logic is already in the polyfill code, we shouldn't need it here. In my other comment, I was referring to using some kind of webpack magic to make sure the file is only downloaded when using IE11. This would require moving it out of global and into a separate entry point. However, if this is only 3kb uncompressed, I'm fine with just including it in the main bundle.

bookernath: Okay, I've dropped that change.

PR #1561, opened 2019-08-13
Simple GraphQL example using Apollo Client

bookernath:
#### What?

Adding a simple GraphQL request example as a branch so we can link to it as sample code.


bc-williamkwon:
Don't have much to say about this example as it kind of lean besides the query itself.  However, using the api via cornerstone seems very straightforward with this example.

PR #1577, opened 2019-09-24
STRF-6450 - Updates to Product Page Template for Google Structured Data

bc-jz:
#### What?
It was found that we are not properly marking up some fields that are available on a product page for Google Structured Data.  This PR fixes some of our markup and adds some additional markup.

The current errors seen on most product pages (which have reviews added to the product) is:

Also found a bug where if a price range is being displayed then the "offer availability" property does not display correctly:

#### Tickets / Documentation
JIRA ticket:
https://jira.bigcommerce.com/browse/STRF-6450

Google Structured Data (Product) information:
https://developers.google.com/search/docs/data-types/product

Screenshot having applied my updated cornerstone bundle to my production store that I used to generate the error images above.  The product is displaying a price range:

Note only 1 warnings is now seen. This remaining warning for `priceValidUntil` is not something we can currently address.

### DIFF
--- a/templates/components/products/price-range.html
+++ b/templates/components/products/price-range.html
@@ -29,8 +29,11 @@
             <abbr title="{{lang 'products.including_tax'}}">{{lang 'products.price_with_tax' tax_label=price.price_range.min.tax_label}}</abbr>
         {{/and}}
         {{#if schema_org}}
-            <meta itemprop="availability" content="{{product.availability}}">
+            <meta itemprop="availability" itemtype="http://schema.org/ItemAvailability"
+                    content="http://schema.org/{{#if product.pre_order}}PreOrder{{else if product.out_of_stock}}OutOfStock{{else if product.can_purchase '===' false}}OutOfStock{{else}}InStock{{/if}}">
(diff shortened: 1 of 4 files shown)


templates/components/products/price-range.html
```diff
@@ -29,8 +29,11 @@
             <abbr title="{{lang 'products.including_tax'}}">{{lang 'products.price_with_tax' tax_label=price.price_range.min.tax_label}}</abbr>
         {{/and}}
         {{#if schema_org}}
-            <meta itemprop="availability" content="{{product.availability}}">
+            <meta itemprop="availability" itemtype="http://schema.org/ItemAvailability"
```
bookernath: Should this be solved in cornerstone or server side?

bc-jz:
I think we will need changes to template logic either way.  What I added to the `price-ranges` template was following an established pattern in `prices` which is was missing for parity.  If we alter our backend behavior to return already calculated `availabilityStatus` (or something like that) value then templates will still need changes to use this new value while removing their existing logic (at least in `prices` template).

PR #1590, opened 2019-11-01
chore: bumping stencil-cli forward for node 10 support

ncheikh:
#### What?

Updating the stencil-cli dependency version to latest in order to support node 10.


ncheikh:
@junedkazi
- Node: 10.16.3
- NPM: 6.9.0

PR #1606, opened 2019-12-03
Fix to make browser back button navigate pagination properly when faceted search is enabled

bc-williamkwon:
#### What?
Tested on store: https://williamkwon1576004500-testly-the-third.my-integration.zone/
Previously, the browser back button would not navigate pagination properly when faceted search was enabled but now it does.

#### Tickets / Documentation

- [Link 1](https://jira.bigcommerce.com/browse/STRF-7232)

**Screenshots (if appropriate)**
https://imgur.com/a/MtmIeIR

### DIFF
--- a/assets/js/theme/common/faceted-search.js
+++ b/assets/js/theme/common/faceted-search.js
@@ -298,6 +298,7 @@ class FacetedSearch {
 
         // DOM events
         $(window).on('statechange', this.onStateChange);
+        $(window).on('popstate', this.onPopState);
         $(document).on('click', this.options.showMoreToggleSelector, this.onToggleClick);
         $(document).on('toggle.collapsible', this.options.accordionToggleSelector, this.onAccordionToggle);
         $(document).on('keyup', this.options.facetedSearchFilterItems, this.filterFacetItems);
@@ -312,6 +313,7 @@ class FacetedSearch {
     unbindEvents() {
         // DOM events
         $(window).off('statechange', this.onStateChange);
+        $(window).off('popstate', this.onPopState);
         $(document).off('click', this.options.showMoreToggleSelector, this.onToggleClick);
         $(document).off('toggle.collapsible', this.options.accordionToggleSelector, this.onAccordionToggle);
         $(document).off('keyup', this.options.facetedSearchFilterItems, this.filterFacetItems);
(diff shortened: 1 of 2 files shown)


mattolson:
Please provide proof that the change works as expected.

bc-williamkwon:
Added a screenshot and linked to store with theme that has changes

bookernath:
💚

bookernath:
I see the issue is still happening with a filter applied, e.g.:

https://williamkwon1575504304-testsworthy.my-integration.zone/shop-all/?_bc_fsnf=1&brand=36&limit=2

From here, you can navigate to page 2 and then click the back button - you'll notice the page number in the content footer does not change, and the products do not update.

bc-williamkwon:
Made changes to fix issues.

bookernath:
💚 all cases seem fixed now. Nice work!

bookernath:
The issue with the page not reloading seems fixed now.

However, it appears that we've introduced a new bug with this change:

Now if I'm on page 1 and I directly click on page 3, and then click the back button, I'll be taken to page 2, even though I never visited that page so it shouldn't be in my history.

bc-williamkwon:
Fixed said bug described in last comment

assets/js/theme/common/faceted-search.js
```diff
@@ -408,6 +410,18 @@ class FacetedSearch {
...
         }
     }
+
+    onPopState() {
+        const currentUrl = window.location.href;
+        const searchParams = new URLSearchParams(currentUrl);
+        // If searchParams does not contain a page value then modify url query string to have page=1
+        if (!searchParams.has('page')) {
+            const linkUrl = $('.pagination-link').attr('href');
+            const url = `${linkUrl.slice(0, linkUrl.indexOf('page'))}page=1`;
```
mattolson: Why does `linkUrl.indexOf('page')` return something if `searchParams.has('page')` is false?

bc-williamkwon: linkUrl is taken from the dom searchParams is taken from the url.  Each page has a corresponding href on the dom.

mattolson: I think we need to use something more robust that updates just the page parameter. `linkUrl` could contain other search parameters when facets are selected, and this code assumes that `page` is the last parameter, which may not always be the case.

bc-williamkwon:
Made change to only update page parameter

bc-williamkwon:
Updated solution to use regex instead of split.

bookernath:
I've tested all the cases I tested before, and this looks good now 💚

PR #1609, opened 2019-12-13
STRF-7755: added support to multiple template engines

bc-zoharmuzafi:
#### What?
Added a new variable to config file to allow specify template engine.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,6 +1,6 @@
 {
   "name": "Cornerstone",
-  "version": "4.3.0",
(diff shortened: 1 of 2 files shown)


fthobe:
Hey @bc-zoharmuzafi , is this commit backwards compatible with older versions of the cornerstone theme? I don't see any issue but I wanted to be sure!

junedkazi:
I am going to close this one for now as won't fix. We can always open it when we do decide to proceed with it.

PR #1631, opened 2020-02-03
PDP: Enable buy now button functionality

sacr3dc0w:
### What?

We've utilized this on our ecommerce sites with much success the past few years. Finally got around to cleaning it up for a PR. Makes use of the add to cart via URL functionality.

### Screenshots

**PDP - Buy Now**
01 PDP - Buy Now
**PDP - Buy Now - Mobile**
02 PDP - Buy Now - Mobile
**PDP - Buy Now - HTML5 Validation**
03 PDP - Buy Now - HTML5 Validation
**PDP - Buy Now - Alert**
Screen Shot 2020-02-03 at 6 59 51 PM


bookernath:
FYI @bigcommerce/checkout

assets/js/theme/common/product-details.js
```diff
@@ -764,4 +765,51 @@ export default class ProductDetails {
...
+            // `attr` is false.  Check for both.
+            if (typeof attr !== typeof undefined && attr !== false) {
+                const message = $form.find('.alertBox-column.alertBox-message').text();
+                showAlertModal(message);
+                return false;
+            }
+            if ($form[0].checkValidity()) {
+                const formStr = $form.serialize();
+                const replacedString = this.replaceBulk(formStr, ['%5B', '%5D', 'add', 'qty%5B%5D'], ['[', ']', 'buy', 'qty']);
+                window.location.href = `/cart.php?${replacedString}&source=buy-now-button`;
```
serroba: as we discussed in the channel, IMO, we should be using the storefront APIs instead if this is something we want to use in cornerstone. I am fine keeping around the existent endpoint for email marketing, but for any client JS-enabled like this one, we should be using our public APIs instead.

serroba: To be more clear, if this was not clear the first time and due to the lack of added context of my original comment. It would be ideal if we can work on top of the publicly documented APIs we have built for JS clients https://developer.bigcommerce.com/api-docs/cart-and-checkout/working-sf-apis

The `/cart.php` endpoint will likely stay around for a while, but is more limited on some capabilities and is not ideal for many scenarios. The storefront APIs are the place where we want to keep adding features and functionality. If there are some features lacking in those APIs, we should probably enhance those endpoints to reach feature parity.

The use case where I think keeping around a `GET` endpoint is mostly for email marketing purposes, where is much less common to have JS available to our advantage. For other cases, like in is in this case, where we do have access to JS capabilities, then working on top of the APIs would be ideal.

sacr3dc0w: @serroba Will work on refactoring to make use of storefront APIs. Created this a long time ago and am amazed we've rode it this far and wanted to share it with the community. Very much appreciate the feedback!

bookernath:
@sacr3dc0w I had a chat with my colleague on the BC Product team and we're fine to add this feature into Cornerstone, and we thank you once again for your contribution. This is of course pending the code review and approval from our engineering team, which you're probably quite familiar with by now.

I'm not sure this feature should be enabled by default, for 2 reasons:

1. It's a rather jarring behavioral change for folks updating from previous versions of Cornerstone, who may not expect this change to their PDP behavior
2. In general, Cornerstone has to be a good "out-of-the-box" default theme for all kinds of merchants across many different verticals. I can imagine this being an excellent feature for many merchants, but I can also see it alienating certain kinds of merchants as well such as B2B merchants or those who do a lot of cross-sell, up-sell, bulk quantities, and so on. In general, I imagine the desire for this feature will correlate with average cart size for a particular business.

How would you feel about defaulting it to off? We'll definitely give you a shout-out on our next Town Hall to make sure folks are aware of it. :)

sacr3dc0w:
@bookernath Not a problem at all with the defaulting to off. Makes total sense and don't want to alienate or confuse other merchants.

Our ecommerce sites generally have a lot of repeat B2B customers, but we've noticed that this has captured a lot of B2C that we may not have captured without since they're generally not our target audience. Not sure if that feedback is helpful for the product team on possible outcomes merchants might expect.

Going to work on refactoring to make use of storefront APIs. Appreciate all the feedback!

assets/js/theme/common/product-details.js
serroba: Thank you @sacr3dc0w for your contribution and sorry for the my first comment, I didn't realize this was coming from the community, hence my lack of context in the first comment.

PR #1636, opened 2020-02-11
Remove "cookie notification" feature as it is no longer useful

bookernath:
#### What?

Removing the "EU Cookie Notification" feature which would show a notice to EU visitors to the site that cookies are used as part of the experience.

This is being removed as BigCommerce has adapted to changing interpretations of GDPR, and have determined that this feature is no longer useful to merchants who are seeking to be compliant with data privacy regulations.

Merchants are recommended to consider our new [Shopper Consent Tracking feature
](https://support.bigcommerce.com/s/article/Store-Settings#security) and consider our other guidance related to local laws:

https://support.bigcommerce.com/s/article/General-Data-Protection-Regulation#learning
https://support.bigcommerce.com/s/article/Implementing-CCPA-with-BigCommerce

As always, merchants are advised to seek their own counsel regarding any data privacy requirements that may apply to their business.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -867,9 +867,6 @@
         "view_more": "View more details",
         "view_less": "Hide details"
     },
-    "cookies": {
-        "accept": "Accept Cookies"
-    },
     "maintenance": {
         "down": "Down for Maintenance"
     }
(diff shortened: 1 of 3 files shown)


mattcoy-arcticleaf:
@bookernath  would it be safe to also remove the cookieNotification.js file and the CSS for the cookie banner?

bookernath:
Yes, and nice catch. Not sure how I missed those.

PR #1637, opened 2020-02-12
feat(search): ES-97 fixed the html special chars issue in facet names

bc-krishsenthilraj:
#### What?
As part of fixing the encoding issues for ES-97, we found that html characters are being displayed in stencil theme. This fix will address it.

#### Tickets / Documentation
- [https://jira.bigcommerce.com/browse/ES-97](ES-97)

Broken "RefineBy" facet names
---------------------------------
image

Fixed "RefineBy" facet names
-------------------------------
image

### DIFF
--- a/templates/components/faceted-search/facets/hierarchy-children.html
+++ b/templates/components/faceted-search/facets/hierarchy-children.html
@@ -7,7 +7,7 @@
             rel="nofollow"
             data-id="{{ id }}"
             data-faceted-search-facet>
-                {{title}}
+                {{{sanitize title}}}
                 {{#if ../show_product_counts}}
                     {{#if count}}
                         <span>({{count}})</span>
--- a/templates/components/faceted-search/facets/multi.html
+++ b/templates/components/faceted-search/facets/multi.html
@@ -30,7 +30,7 @@
                         class="navList-action navList-action--checkbox {{#if selected }} is-selected {{/if}}"
                         rel="nofollow"
                         data-faceted-search-facet>
-                        {{ sanitize title }}
+                        {{{ sanitize title }}}
                         {{#if ../show_product_counts}}
                             <span>({{ count }})</span>
                         {{/if}}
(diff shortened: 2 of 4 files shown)


templates/components/faceted-search/facets/multi.html
```diff
@@ -30,7 +30,7 @@ <h5 class="accordion-title">
                         class="navList-action navList-action--checkbox {{#if selected }} is-selected {{/if}}"
                         rel="nofollow"
                         data-faceted-search-facet>
-                        {{ sanitize title }}
+                        {{{ title }}}
```
junedkazi: We should also update the label across each of the facets.

templates/components/faceted-search/selected-facets.html
```diff
@@ -15,7 +15,7 @@ <h5 class="sidebarBlock-heading">
                         {{#if facet '===' 'rating'}}
                             {{lang 'search.faceted.selected.rating-label' rating=title}}
                         {{else}}
-                            {{ title }}
+                            {{{ title }}}
```
junedkazi: I think we should do it across all these files https://github.com/bigcommerce/cornerstone/tree/master/templates/components/faceted-search/facets

bc-krishsenthilraj: I have made changes to all applicable files. Please review them.

templates/components/faceted-search/facets/hierarchy-children.html
```diff
@@ -7,7 +7,7 @@
             rel="nofollow"
             data-id="{{ id }}"
             data-faceted-search-facet>
-                {{title}}
+                {{{title}}}
```
lord2800: We should probably `sanitize` all of the titles--there's little to no reason anyone should be using html in a custom field or option value name or value.

bc-krishsenthilraj: added them.. pls review it.

Issue #1312, opened 2018-07-12
Search Page Pagination issues.


hiteshvaghasiya83:
There seems to be an issue with pagination on search page.
**Case:** More than 1 page for a particular search result set.
Tab 1: "Products"
Tab 2: "News & Information"

**Issue:**
Switching from Tab 1 to Tab 2 it retains the page number selected.
Which could be an issue when Tab 2 doesn't have page 2 resulting in showing 0 results.
It's a vice versa situation for Tab 1 and Tab 2.

### Steps to reproduce the behavior
Add multiple products with the same name and search with the same name. also, add the same content on the page.
"Products" tab have pagination and "News & Information" tab also result.

Goto second page of "Products" and click on "News & Information" you can show blank on  "News & Information" tab.

### Suggestion
Add different pagination for "Products"  and "News & Information".
so it will managed by separately

Ubersmake:
Thanks for writing this. There's an engineering ticket open to track this and the similar https://github.com/bigcommerce/cornerstone/issues/1170.

PR #1641, opened 2020-02-16
Search Results: fix navigation issues

### LINKED ISSUE #1170: Search pagination displays no content when switching between sections and one section has more pages then the other
Note, the [Cornerstone demo](http://cornerstone-light-demo.mybigcommerce.com/search.php?search_query=blog&section=content&page=2), does not have enough content to view this issue, but is reproducible.

### Expected behavior
When switching between sections on a search, if one section has 100 items, and the other has 10, if you're viewing content on page=4, and switch to the other section, it should show either it's last page, or reset the pagination.

### Actual behavior
When using the pagination, and switching between sections, the page counter is not adjusted or reset, and wrongly displays no content.

### Steps to reproduce behavior
Search a term that has pages worth of products, and a few blog posts, and use the pagination til it's above the page amount of the blog section ( or vice versa ), and switch sections. The page counter doesn't adjust itself to reflect the page amount discrepancy, and shows no items.
image

This is more readily viewable when both sections have a discrepancy in page numbers.

jbruni:
#### What?

This PR contains a number of bug fixes related to the Search Results page navigation.

If fixes #1170 and #1312 by separating the current navigation **page** between "Products" and "News & Information" tabs. It becomes possible, for example, to be at "page 3" in one tab, and at "page 1" in the other tab.

It also improves the fix to #1442, provided by commit 5322976e. Now, the back button also works when there is a switch between "Products" and "News & Information" tabs, when navigating backwards.

And it also fixes the dynamic loading of search results for the "News & Information" tab. The AJAX load/update only worked for the "Products" tab. Now, it works also for the "News & Information" tab.

It also fixes the annoying "re-load" of the search results immediately after the first load of the search results page.

While working in this PR, another search results navigation issue was found, but the problem is at `stencil-utils`. It has been reported at https://github.com/bigcommerce/stencil-utils/pull/110#issuecomment-586749147 and a PR for the fix has been submitted at https://github.com/bigcommerce/stencil-utils/pull/117

-------------------------------------------

##### Technical Details

In this section, I will explain the code changes in this PR.

:one: [Changes to `search.html` template](https://github.com/jbruni/cornerstone/commit/1d95e3b3450171f598528ec297594103dae8e41c#diff-d1a81eff432508a1eff09cf7a636130e):

- We already had `{{> components/search/product-listing}}` for the products results and `{{> components/search/product-count}}` for the products count, as partials for the **Products** tab.

- Now, we introduce `{{> components/search/content-listing}}` and `{{> components/search/content-count}}` partials for the **News & Information** tab results and count.

- Note that we **_always want to render the container elements_**, in order to support toggling between a "zero results page" and a "page with results". We can't keep rendering the containers conditionally (only if there are results) at this upper level (`search.html`)... if the containers are not rendered, when receiving non-empty search results from a future AJAX call, we are not be able to update the DOM.

:two: [New `content-count.html` partial](https://github.com/jbruni/cornerstone/commit/1d95e3b3450171f598528ec297594103dae8e41c#diff-844a40c0ef56c7b0826890d381978209) and [changes to `product-count.html` partial](https://github.com/jbruni/cornerstone/commit/1d95e3b3450171f598528ec297594103dae8e41c#diff-84773ea3021829e48ce4b49c2a51c0df):

- As part of the strategy of keeping the page number of each tab separated, we update only the respective section (`content` or `product`) when updating the DOM. In other words: when the `product` section is updated, the `content` section will NOT be updated; and when the `content` section is updated, the `product` section will NOT be updated. Makes sense.

- To achieve the expected results, we output the search `url` and `count` values as `data` attributes in a dummy wrapper `span` element, for later retrieval and usage in the JavaScript code.

:three: [New `content-listing.html` partial](https://github.com/jbruni/cornerstone/commit/1d95e3b3450171f598528ec297594103dae8e41c#diff-1d01b2f36dea1b5f96c8da7c616d4469):

- This has been extracted from `search.html` to its own partial, so it is possible to update it dynamically.

:four: [Changes to `search.js` code](https://github.com/jbruni/cornerstone/commit/1d95e3b3450171f598528ec297594103dae8e41c#diff-c6f8ca603e58d1d49d5479308da4fac0):

- Let's see the changes to `initFacetedSearch()` first:
    1. We add jQuery wrappers to the "News & Information" listing and count elements.
    2. We include the new `content-*` components in the list of partials to retrieve.
    3. In the callback which processes the AJAX call results, there are three changes:
        - We added updates for the two new `content-*` partials;
        - We only update the `content` or `product` elements, according to the requested section (instead of updating everything);
        - We call the `showProducts` or `showContent` method, to switch to the proper tab; this ensures the proper tab is displayed after using the **back button** to browse backwards.

- The changes to `showProducts` and `showContent` are similar. Note that these functions do two things: update the UI, showing the product or content tab; and they also trigger the dynamic load/update of results.
    1. We introduced a boolean `navigate` parameter (with default to `true`, to keep the previous behaviour). This way, it is possible to "separate" the two things that these functions do. If we want to switch the UI tab only, without doing AJAX request to load contents, we can do it by passing `false` as argument to the `navigate` parameter.
    2. When `navigate` is `true`, instead of using the `window.location.href`, which can point to a different tab, we are looking into the `data-url` and `data-count` attributes introduced to the `*-count` partials. The query string parameter for `section` will be always correct. And, if the `count` is zero, we force the query string parameter `page` to `1`. (The use case here is when the initial load was for a page greater than `1`, and the tab didn't contain results for that page.)

-------------------------------------------

These are the minimal changes to make the existing navigation in a working state. There is room for further improvements.

#### Tickets / Documentation

- Issue #1312  
- Issue #1170 
- Issue #1442 
- Commit 5322976e
- stencil-utils issue https://github.com/bigcommerce/stencil-utils/pull/110#issuecomment-586749147
- stencil-utils PR https://github.com/bigcommerce/stencil-utils/pull/117

N/A

### DIFF
--- a/assets/js/theme/search.js
+++ b/assets/js/theme/search.js
@@ -33,11 +33,7 @@ export default class Search extends CatalogPage {
         return nodeData;
     }
 
-    showProducts() {
-        const url = urlUtils.replaceParams(window.location.href, {
-            section: 'product',
-        });
-
+    showProducts(navigate = true) {
         this.$productListingContainer.removeClass('u-hiddenVisually');
         this.$facetedSearchContainer.removeClass('u-hiddenVisually');
         this.$contentResultsContainer.addClass('u-hiddenVisually');
@@ -48,14 +44,19 @@ export default class Search extends CatalogPage {
         $('[data-product-results-toggle]').removeClass('navBar-action');
         $('[data-product-results-toggle]').addClass('navBar-action-color--active');
 
+        if (!navigate) {
+            return;
+        }
+
+        const searchData = $('#search-results-product-count span').data();
+        const url = (searchData.count > 0) ? searchData.url : urlUtils.replaceParams(searchData.url, {
+            page: 1,
+        });
+
         urlUtils.goToUrl(url);
     }
 
-    showContent() {
-        const url = urlUtils.replaceParams(window.location.href, {
-            section: 'content',
-        });
-
+    showContent(navigate = true) {
         this.$contentResultsContainer.removeClass('u-hiddenVisually');
         this.$productListingContainer.addClass('u-hiddenVisually');
         this.$facetedSearchContainer.addClass('u-hiddenVisually');
@@ -65,6 +66,15 @@ export default class Search extends CatalogPage {
 
         $('[data-content-results-toggle]').removeClass('navBar-action');
         $('[data-content-results-toggle]').addClass('navBar-action-color--active');
+
+        if (!navigate) {
+            return;
+        }
+
+        const searchData = $('#search-results-content-count span').data();
+        const url = (searchData.count > 0) ? searchData.url : urlUtils.replaceParams(searchData.url, {
+            page: 1,
+        });
 
         urlUtils.goToUrl(url);
     }
@@ -102,9 +112,9 @@ export default class Search extends CatalogPage {
         });
 
         if (this.$productListingContainer.find('li.product').length === 0 || url.query.section === 'content') {
-            this.showContent();
+            this.showContent(false);
         } else {
-            this.showProducts();
+            this.showProducts(false);
         }
 
         const validator = this.initValidation($searchForm)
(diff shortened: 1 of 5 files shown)


junedkazi:
@jbruni thank you for fixing this issue.

PR #1647, opened 2020-03-12
fix(storefront): STRF-5678 fixed the check for max length in create account fields

bc-nataliya:
#### What?

Top 20 bug reports that while user can set the max length for the account sign up fields it is not respected in the storefront

#### Tickets / Documentation

https://jira.bigcommerce.com/browse/STRF-5678

Screen Shot 2020-03-12 at 2 00 08 PM

To test:
1. Navigate to Control Panel -> Advanced Settings -> Account Sign up 
2. Change max length on any text field (for example Address 1 field)
3. Navigate to storefront and click on Register
4. Type in the field with defined max length and observe how it is not allowing you enter more characters than specified in max length

ping @bigcommerce/storefront-team @bigcommerce/artemis-dt

### DIFF
--- a/templates/components/common/forms/text.html
+++ b/templates/components/common/forms/text.html
@@ -2,5 +2,5 @@
     <label class="form-label" for="{{id}}_input">{{label}}
         {{#if required}}<small>{{lang 'common.required' }}</small>{{/if}}
     </label>
-    <input type="text" id="{{id}}_input" data-label="{{label}}" name="{{name}}" {{#if value}} value="{{value}}"{{/if}} {{#if placeholder}} placeholder="{{placeholder}}"{{/if}} class="form-input" aria-required="{{required}}" {{#if private_id}}data-field-type="{{private_id}}"{{/if}}>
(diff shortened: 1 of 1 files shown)


bc-nataliya:
@junedkazi  multiline doesn't have an option of setting max length

templates/components/common/forms/text.html
```diff
@@ -2,5 +2,5 @@
     <label class="form-label" for="{{id}}_input">{{label}}
         {{#if required}}<small>{{lang 'common.required' }}</small>{{/if}}
     </label>
-    <input type="text" id="{{id}}_input" data-label="{{label}}" name="{{name}}" {{#if value}} value="{{value}}"{{/if}} {{#if placeholder}} placeholder="{{placeholder}}"{{/if}} class="form-input" aria-required="{{required}}" {{#if private_id}}data-field-type="{{private_id}}"{{/if}}>
+    <input type="text" id="{{id}}_input" data-label="{{label}}" name="{{name}}" {{#if value}} value="{{value}}"{{/if}} {{#if placeholder}} placeholder="{{placeholder}}"{{/if}} class="form-input" aria-required="{{required}}" {{#if private_id}}data-field-type="{{private_id}}"{{/if}} {{#if maxlength}}maxlength="{{maxlength}}"{{/if}}>
```
bookernath: You could consider putting the space inside the `{{#if}}` block so it only makes it to the markup if this argument is used 🍹

bc-aodvak:
tested on integration
Set length for all fields to 5, except State - set to 1 and zip - set to 2.

Screen Shot 2020-03-12 at 3 18 35 PM

looks good 💚

PR #1654, opened 2020-04-01
feat(search): ES-466 Facets with & symbol works now

bc-krishsenthilraj:
#### What?
Faceted search filters that include ampersand (&) breaks 'Show More' links. 

With this fix, now we can have Facets with `&` symbol and `Show More` link works fine.

#### Tickets / Documentation

- [ES-466](https://jira.bigcommerce.com/browse/ES-466)

#### Testing Steps:
1. Login to CP
2. Navigate to Products
3. Edit few Products and add a `Custom Field` (with same name like _ Custom & Color _) and with different values for each product and save the Products.
4.  Navigate to Products->Product Filtering
5. Enable the newly created facet.
6. Click the settings of this facet filter and change the value for `show` to 5 items and save
6. Go to SF and click on the `Show More` link for this new filter and the popup window shows all the filter values.

1. List of all Facets:
image

2. Facet with `&` in the name & value
image

3. `Show More`  link displays the popup with filterable values
image

4. Other Facet's `Show More` link works fine too.. 
image

@bigcommerce/artemis-dt  @lord2800 @kwang30

### DIFF
--- a/templates/components/faceted-search/facets/hierarchy.html
+++ b/templates/components/faceted-search/facets/hierarchy.html
@@ -3,7 +3,7 @@
     <div
         class="accordion-navigation toggleLink {{#unless ../start_collapsed }} is-open {{/unless}}"
         role="button"
-        data-collapsible="#facetedSearch-content--{{hyphenate facet }}">
+        data-collapsible="#facetedSearch-content--{{#replace '&' (hyphenate facet) }}{{/replace}}">
         <h5 class="accordion-title">
             {{ title }}
         </h5>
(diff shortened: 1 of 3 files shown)


templates/components/faceted-search/facets/hierarchy.html
```diff
@@ -3,7 +3,7 @@
     <div
         class="accordion-navigation toggleLink {{#unless ../start_collapsed }} is-open {{/unless}}"
         role="button"
-        data-collapsible="#facetedSearch-content--{{hyphenate facet }}">
+        data-collapsible="#facetedSearch-content--{{#replace '&' (hyphenate facet) }}{{/replace}}">
```
junedkazi, line 6: I really don't like the idea of doing special characters replaces at the handlebars level. Today we are doing `&` tomorrow it is going to be some other special characters.

We should either strip it or escape it at the backend level. Right now

lord2800: You can't escape `&` to be placed in a fragment. There are no other special characters like this.

junedkazi: Since we are just doing a replace for & in the string for the fragment can't we do that on the backend itself ?

bc-aodvak:
Tested on staging by uploading and applying theme from the ticket

Screen Shot 2020-04-06 at 12 54 05 PM

Looks good 💚

PR #1655, opened 2020-04-02
Account form classes

PeteyRev:
#### What?

When editing Account Signup Fields in the Control Panel, there are fields for 
`CSS Class Name, Inline CSS Style, Field Size` that can be customized by the user, however these do not currently work. The data is present , it just was never added via handlebars

oBundle - Lead Dev Sandbox - BigCommerce Control Panel
oBundle - Lead Dev Sandbox - Create Account

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9631,6 +9631,20 @@
             "boom": "2.x.x",
             "hoek": "2.x.x",
             "joi": "5.x.x"
+          },
+          "dependencies": {
+            "joi": {
+              "version": "5.1.0",
+              "resolved": "https://registry.npmjs.org/joi/-/joi-5.1.0.tgz",
+              "integrity": "sha1-FSrQfbjunGQBmX/1/SwSiWBwv1g=",
+              "dev": true,
+              "requires": {
+                "hoek": "^2.2.x",
+                "isemail": "1.x.x",
+                "moment": "2.x.x",
+                "topo": "1.x.x"
+              }
+            }
           }
         },
         "hoek": {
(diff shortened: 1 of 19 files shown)


package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "bigcommerce-cornerstone",
-  "version": "4.4.1",
+  "version": "4.5.0",
```
junedkazi, line 3: @PeteyRev thank you for your contribution. Can you please rebase and drop the `package-lock.json` change. We do need to fix it and we are going to fix our release script to make sure this does not happen again.

PeteyRev: @junedkazi Should be gtg

package-lock.json
```diff
@@ -9846,6 +9860,7 @@
             "hoek": "2.x.x"
           }
         },
+<<<<<<< HEAD
```
junedkazi, line 9863: @PeteyRev I don't believe the rebase went well since I am seeing these line.

package-lock.json
```diff
@@ -9862,6 +9877,12 @@
           "version": "6.0.0",
           "resolved": "https://registry.npmjs.org/wreck/-/wreck-6.0.0.tgz",
           "integrity": "sha1-T0CGaWHl14rOBPMqa38x8/PFFjg=",
+=======
```
junedkazi, line 9880: same here

package-lock.json
```diff
@@ -9862,6 +9877,12 @@
...
+        "wreck": {
+          "version": "6.3.0",
+          "resolved": "https://registry.npmjs.org/wreck/-/wreck-6.3.0.tgz",
+          "integrity": "sha1-oTaXafB7u2LWo3gzanhx/Hc8dAs=",
+>>>>>>> 4bd50865cf4e39814a9a0c485c5643cc99799020
```
junedkazi, line 9885: same here

PR #1661, opened 2020-04-22
Reset cart quantity to 0 if we get a 404 for the cart

bookernath:
#### What?

If we get a 404 for the cart, set the quantity to 0 for the cart preview.

Fixes a bug where the quantity will not update back to 0 after a successful checkout.

#### Tickets / Documentation

- [STRF-8295](https://jira.bigcommerce.com/browse/STRF-8295)

### DIFF
--- a/assets/js/theme/global/cart-preview.js
+++ b/assets/js/theme/global/cart-preview.js
@@ -69,7 +69,12 @@ export default function (secureBaseUrl, cartId) {
         const cartQtyPromise = new Promise((resolve, reject) => {
             utils.api.cart.getCartQuantity({ baseUrl: secureBaseUrl, cartId }, (err, qty) => {
                 if (err) {
-                    reject(err);
+                    // If this appears to be a 404 for the cart ID, set cart quantity to 0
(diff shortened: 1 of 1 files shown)


assets/js/theme/global/cart-preview.js
```diff
@@ -69,7 +69,12 @@ export default function (secureBaseUrl, cartId) {
         const cartQtyPromise = new Promise((resolve, reject) => {
             utils.api.cart.getCartQuantity({ baseUrl: secureBaseUrl, cartId }, (err, qty) => {
                 if (err) {
-                    reject(err);
+                    // If this appears to be a 404 for the cart ID, set cart quantity to 0
+                    if (err === 'Not Found') {
```
bookernath, line 73: Why the messy string comparison?

Our request implementation only returns the string of the response code, not the actual error object:

https://github.com/bigcommerce/stencil-utils/blob/master/src/lib/request.js#L127

As such, I have nothing else to match on without a significant refactor.

Fortunately, the fallback behavior here is pretty elegant.

PR #1665, opened 2020-04-30
fix(catalog): CATALOG-5557 Special characters are not rendered for product review

BC-DNikanov:
Special characters are not rendered for product review titles and text on storefront

#### What?
Special characters in product review titles and texts are displaying html entities. The entities that are occurring are generally for single quotes and ampersand.

#### Tickets / Documentation

https://jira.bigcommerce.com/browse/CATALOG-5557

#### Screenshots
Screen Shot 2020-04-30 at 2 39 30 PM

### DIFF
--- a/templates/components/products/reviews.html
+++ b/templates/components/products/reviews.html
@@ -35,7 +35,7 @@
                             </p>
                         {{/if}}
                     </header>
-                    <p itemprop="reviewBody" class="productReview-body">{{nl2br text}}</p>
+                    <p itemprop="reviewBody" class="productReview-body">{{{ sanitize text }}}</p>
                 </article>
             </li>
             {{/each}}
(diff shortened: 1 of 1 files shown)


templates/components/products/reviews.html
```diff
@@ -35,7 +35,7 @@ <h5 itemprop="name" class="productReview-title">{{ title }}</h5>
                             </p>
                         {{/if}}
                     </header>
-                    <p itemprop="reviewBody" class="productReview-body">{{nl2br text}}</p>
+                    <p itemprop="reviewBody" class="productReview-body">{{{ sanitize text }}}</p>
```
junedkazi, line 38: can we please do the same for the review title as well

BC-DNikanov: Sure. I've done the same for the review title.

olysenko:
💚

templates/components/products/reviews.html
fthobe: The same issue is also present on facets

hoofbootique:
This doesn't work for me, if I add the new line and remove the old one my reviews no longer paragraph properly and instead show as big chunks of text with no breaks. Yes it removed the weird special characters ' cause, but the paragraph breaks are gone. Am I doing something wrong?

PR #1668, opened 2020-05-07
feat(checkout): MCU-427 Display modal before switching currencies

mjhmatt:
#### What?
Before switching currencies in the case of a cart containing a gift certificate/promotion we will need to notify the customer as they may be removed from cart once the currency is changed.

#### Tickets / Documentation
https://jira.bigcommerce.com/browse/MCU-427

Screen Shot 2020-05-07 at 4 07 54 pm
Screen Shot 2020-05-07 at 3 51 57 pm

@bigcommerce/checkout


templates/components/common/currency-selector.html
```diff
@@ -5,16 +5,23 @@
         <ul class="dropdown-menu" id="currencySelection" data-dropdown-content aria-hidden="true" tabindex="-1">
             {{#each currency_selector.currencies}}
             <li class="dropdown-menu-item">
-                <a href="{{{switch_url}}}">
-                    {{#if is_active}}
-                    <strong>{{name}}</strong>
-                    {{else}}
-                    {{name}}
+                <a href="{{{switch_url}}}"
+                   {{#if ../currency_selector.has_discount}}
```
capsula4, line 9: what if they have discounts + GC?

mjhmatt: I was going to ask @bc-charlesho this tomorrow. The current implementation just shows discount warning

bc-charlesho: Hi Matt, Cart Level Discounts going forwards will be marketed as Promotions. So we can say this:

"Promotions and gift certificates that don't apply to the new currency will be removed from your cart. Are you sure you want to continue?"

bookernath:
@mjhmatt please rebase

package.json
```diff
@@ -64,4 +64,4 @@
     "lighthouse": "npx lighthouse --config-path=lighthouse-config.js --quiet --output json --chrome-flags=\"--headless\" $URL | jq '.audits | map_values(.rawValue)'",
     "test": "npx jest"
   }
-}
+}
```
davidchin, line 67: Missing EOL 💅

templates/components/common/currency-selector.html
```diff
@@ -5,16 +5,20 @@
...
+                {{name}}
+                {{/if}}
                 </a>
             </li>
             {{/each}}
         </ul>
     </li>
 </ul>
-{{/if}}
+{{/if}}
```
davidchin, line 24: Missing EOL 💅

config.json
```diff
@@ -781,4 +781,4 @@
       }
     }
   ]
-}
+}
```
davidchin, line 784: Missing EOL 💅

assets/js/theme/global/currency-selector.js
```diff
@@ -1,5 +1,57 @@
...
+        }).then(() => {
+            window.location.reload();
+        });
+    }
+
     $(document.body).on('click', '.currencySelector', () => {
         $('.currency-selection-list').toggleClass('active');
     });
+
+    $(document.body).on('click', '.dropdown-menu-item a', event => {
```
davidchin, line 19: I think it'd be better if you use a more specific selector rather than targeting a generic CSS selector. In this case, you can just target `[data-cart-currency-switch-url]`.

assets/js/theme/global/currency-selector.js
```diff
@@ -1,5 +1,57 @@
...
+        if (!cartId) {
+            return;
+        }
+
+        event.preventDefault();
+
+        utils.api.cart.getCart({ cartId }, (err, response) => {
+            let showWarning = false;
+            response.discounts.forEach((discount) => {
```
davidchin, line 28: You can write `response.discounts.some(discount => discount > 0)`. It exits the loop as soon as there's a match. With that, you can combine all checks into one so you don't need to do other checks if one of those conditions is true. i.e.:

```js
const showWarning = response.discounts.some(discount => discount > 0) ||
  response.coupons.length > 0 ||
  response.lineItems.giftCertificates.length > 0;
```

assets/js/theme/global/currency-selector.js
```diff
@@ -1,5 +1,57 @@
...
+            }
+
+            if (showWarning) {
+                swal.fire({
+                    text: $(event.target).attr('data-warning'),
+                    icon: 'warning',
+                    showCancelButton: true,
+                }).then(result => {
+                    if (result.value && result.value === true) {
+                        changeCurrency($(event.target).attr('data-cart-currency-switch-url'), $(event.target).attr('data-currency-code'));
```
davidchin, line 49: I believe jQuery has a `data` method that can be used to retrieve `data-*` attributes. i.e.: `.data('cart-currency-switch-url')`.

assets/js/theme/global/currency-selector.js
```diff
@@ -1,5 +1,57 @@
...
+
+export default function (cartId) {
+    function changeCurrency(url, currencyCode) {
+        $.ajax({
+            url,
+            method: 'POST',
+            data: JSON.stringify({ currencyCode }),
+        }).then(() => {
+            window.location.reload();
+        });
```
davidchin, line 12: We should probably add a `catch` here and display an error modal if something goes wrong in the request.

assets/js/theme/global/currency-selector.js
```diff
@@ -1,5 +1,57 @@
-export default function () {
+import swal from './sweet-alert';
+import utils from '@bigcommerce/stencil-utils';
+
+export default function (cartId) {
+    function changeCurrency(url, currencyCode) {
+        $.ajax({
+            url,
+            method: 'POST',
```
davidchin, line 8: I think by default, the `content-type` for Ajax requests made with jQuery is `application/x-www-form-urlencoded`. Considering you're sending a JSON body back, it'd be good if you can pass `contentType: 'application/json'`.

assets/js/theme/global/currency-selector.js
```diff
@@ -1,5 +1,57 @@
...
+            data: JSON.stringify({ currencyCode }),
```
davidchin, line 9: You probably don't need to `stringify` yourself. jQuery should do it for you if the content-type is JSON.

assets/js/theme/global.js
```diff
@@ -22,7 +22,7 @@ export default class Global extends PageManager {
         } = this.context;
         cartPreview(secureBaseUrl, cartId);
         quickSearch();
-        currencySelector();
+        currencySelector(this.context.cartId);
```
bookernath, line 25: 🍹 as cartId is already destructured above, you could just reference `cartId`

junedkazi:
@mjhmatt @davidchin is this good to go ?

mjhmatt:
Sorry for the delay. Yes @junedkazi this is finished now, was just waiting to finalise the backend PR.

PR #1671, opened 2020-05-11
Bump jQuery version, expose jquery globally so it doesn't need to be loaded separately

bookernath:
#### What?

Bump jQuery version to [address a CVE](https://blog.jquery.com/2020/04/10/jquery-3-5-0-released/). I've confirmed the main change (related to [htmlPrefilter()](https://api.jquery.com/jQuery.htmlPrefilter/)) does not affect us.

I also exposed `$` on the window object - because I see that lots of custom themes feel the need to separately include jQuery, which will be bad for performance. Exposing our existing bundle jQuery gets rid of the need to do this.

image


webpack.common.js
```diff
@@ -36,6 +36,13 @@ module.exports = {
                     },
                 },
             },
+            {
+                test: require.resolve('jquery'),
+                use: [{
+                    loader: 'expose-loader',
+                    options: '$',
```
junedkazi, line 43: should we also expose `jQuery` as an option ?

junedkazi: Something like this 
```
{
    test: require.resolve('jquery'),
    use: [{
      loader: 'expose-loader',
      options: 'jQuery'
    },{
      loader: 'expose-loader',
      options: '$'
    }]
  }
```

bookernath: Sure!

bookernath: Couldn't get it working, will fix in a follow up.

PR #1676, opened 2020-05-15
Update package-lock.json

rowleyaj:
#### What?

This was created by running npm install in the root of the repo

node version: 10.17.0
as required by stencil-cli

npm version: 6.11.3

I noticed that there is a section in the README that states that node version 8 and npm version 6.4.1 should be used. Is this accurate since stencil-cli requires node 10? If not I can raise a PR to update this as well.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "bigcommerce-cornerstone",
-  "version": "4.5.3",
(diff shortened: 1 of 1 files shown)


rowleyaj:
@bc-williamkwon tagging you as the last person to update the package.json

let me know if you'd welcome a PR for the README as well

PR #1681, opened 2020-05-21
Account: Product option special characters are not rendered on invoice, order contents and returns

sacr3dc0w:


sacr3dc0w:
**Invoice - Before**
01 invoice-before
**Invoice - After**
02 invoice-after
**Order Contents - Before**
03 order-list-before
**Order Contents - After**
04 order-list-after

PR #1689, opened 2020-06-12
feat(search): ES-1192 fixed the replace helper to expand and collapse the product filters

bc-krishsenthilraj:
#### What?
Product filters are not expanded or collapsed in category listing, brand listing & search results page when product filters for SF are turned on. 

### Why
Currently, we are not able to expand/collapse the filters since, `replace` helper returns empty string when the input does not have the special character `&` in it. Thus collapsible divs are having same id and thus its broken this feature in the product filters. 

In the below image, both the `data-collapsible` in the html source is same for both `Catrgory` & `Brand` filters.

image

#### Tickets / Documentation

- [https://jira.bigcommerce.com/browse/ES-1192](ES-1192)

Steps:
1. Login to CP
2. Use the theme file (`Cornerstone-4.6.1_fix_ES-1192.zip`) uploaded in the above issue
3. Upload & Publish the theme by navigating to Storefront
4. Navigate to Products, choose any product 
5. Create a `custom field` with special character `&` in it. Ex: `Custom & Filter`  and save the product.
6. Navigate to Products -> Product Filters and enable product filters
7. Make sure to enable above created custom filter and save.
8. Go to SF and check by expand & collapsing all the product filters.
9. All the filters should be collapsible or expandable.

Images:

1. Before collapasing any filter

image

2. Category filter collapsed

image

3. Category & Brand filters collapsed

image

4. Category Expanded, but Brand collapsed

image

5. Other & Custom filter expanded 
image

6. Other collapsed and Custom expanded
image

7. Both Other & Custom filters collapsed
image

8. Other Expanded again & Custom filter collapsed 
image

9. Both Other & Custom expanded again
image

10. All the filters collapsed
image

### DIFF
--- a/templates/components/faceted-search/facets/hierarchy.html
+++ b/templates/components/faceted-search/facets/hierarchy.html
@@ -3,7 +3,7 @@
     <div
         class="accordion-navigation toggleLink {{#unless ../start_collapsed }} is-open {{/unless}}"
         role="button"
-        data-collapsible="#facetedSearch-content--{{#replace '&' (hyphenate facet) }}{{/replace}}">
+        data-collapsible="#facetedSearch-content--{{#replace '&' (hyphenate facet) }}{{else}}{{facet}}{{/replace}}">
         <h5 class="accordion-title">
             {{ title }}
         </h5>
@@ -21,8 +21,8 @@
         </div>
     </div>
 
-    <div id="facetedSearch-content--{{#replace '&' (hyphenate facet) }}{{/replace}}" class="accordion-content {{#unless ../start_collapsed }} is-open {{/unless}}">
-        <ul id="facetedSearch-navList--{{#replace '&' (hyphenate facet) }}{{/replace}}" class="navList" data-facet="{{facet}}" data-has-more-results="{{ has_more_results }}">
+    <div id="facetedSearch-content--{{#replace '&' (hyphenate facet) }}{{else}}{{facet}}{{/replace}}" class="accordion-content {{#unless ../start_collapsed }} is-open {{/unless}}">
+        <ul id="facetedSearch-navList--{{#replace '&' (hyphenate facet) }}{{else}}{{facet}}{{/replace}}" class="navList" data-facet="{{facet}}" data-has-more-results="{{ has_more_results }}">
             {{#each items}}
                 <li class="navList-item">
                     <a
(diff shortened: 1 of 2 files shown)


templates/components/faceted-search/facets/hierarchy.html
```diff
@@ -3,7 +3,7 @@
     <div
         class="accordion-navigation toggleLink {{#unless ../start_collapsed }} is-open {{/unless}}"
         role="button"
-        data-collapsible="#facetedSearch-content--{{#replace '&' (hyphenate facet) }}{{/replace}}">
+        data-collapsible="#facetedSearch-content--{{#replace '&' (concat '&' (hyphenate facet)) }}{{/replace}}">
```
junedkazi, line 6: Sorry I am not sure I understand what is going on here and what are we trying to do. Can you please an explanation to the description as to why were they broken in the first place. Also I am not sure I understand why we need to add `concat '&'`. Can you please explain using this example `Custom & Filter`. If I am reading correctly it will look something like this;
`Custom & Filter` -> hyphenate `Custom-&-Filter` -> concat `&Custom-&-Filter` -> replace `Custom-Filter`

Please correct me if I am wrong.

bc-krishsenthilraj: replace helper, returns empty string if it doesn't find any `&` in it. Thus all the standard filters where return as `empty string`.  So, we fixed it by `concatenating &` to all the input facets and replace them later. 

Since custom filters are part of this loop, we ended up adding and removing the `&` to it.

lord2800: Open to suggestions here but what happens is if `replace` doesn't find the replacement character, it returns an empty string. So what happens is `Custom Filter` becomes: `'Custom Filter'` -> hyphenate `'Custom-Filter'` -> replace `''`. The `concat` forces there to be at least one replacement token, so it always exists.

junedkazi: It feels so hacky and for input's like this `Custom & Filter` it will generate output with double dash`Custom--Filter`.

lord2800: I agree, but the whole way the expand/collapse system works using css selectors also feels just as hacky to me. I'm open to suggestions here, once again.

junedkazi: I think it should no longer return empty string based on https://github.com/bigcommerce/paper-handlebars/pull/92/files

lord2800: OK, I figured out what the disconnect was here. We needed to use `{{replace}}` with `{{else}}`, which is beyond confusing. 😕

bc-aodvak:
Executed steps above for search, filter by category, filter by brand. Looks good 💚

PR #1693, opened 2020-06-15
fix(search): ES-1322 fixed the collapse and expand issue for custom filter without no special characters

bc-krishsenthilraj:
#### What?
Custom filter without '&' is not collapsable

Since custom filter without `&` and if it has `empty space` between the words (`Ex: Custom Size`), the `data-collapsible` element is broken in the html source. With this fix, 

#### Tickets / Documentation
- [https://jira.bigcommerce.com/browse/ES-1322](ES-1322)

1. All the filters are collapsed
image

2. Standard filters are expanded 
image

3. Both custom filters are expanded
image

4. One custom collapsed (custom filter without `&` in it) and other one is expanded (custom filter with `&` in it)
image

5. One custom collapsed (custom filter with`&` in it) and other one is expanded (custom filter without `&` in it)
image

### DIFF
--- a/templates/components/faceted-search/facets/hierarchy.html
+++ b/templates/components/faceted-search/facets/hierarchy.html
@@ -3,7 +3,7 @@
     <div
         class="accordion-navigation toggleLink {{#unless ../start_collapsed }} is-open {{/unless}}"
         role="button"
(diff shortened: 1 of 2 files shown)


bc-aodvak:
Tested with attached to the ticket theme. Looks good 💚

tvlgiao:
Doesn't work if the filter name contains special characters %+&:()?
I'd suggest using {{dashcase ...}} to remove all non-word characters.

PR #1699, opened 2020-06-23
fix(catalog): BCTHEME-40 Cornerstone - Product Reviews Not Showing If…

BC-tymurbiedukhin:
… Tabs Disabled

#### What?

On the most recent version of Cornerstone, 3.2.2, if you have tabs disabled and reviews set to show in the theme editor, the product reviews will not be displayed.

In theme setting we have two checkboxes "Show descriptions tabs"(let it be D) and "Show product reviews"(let it be R). And we have to handle any combinations of these checkbox enabling ('+' means turned on, '-' means turned off)

Previous logic:
+D+R = both description and reviews tabs
+D-R = description tab and expandable review section
-D-R = text description and expandable review section
-D+R = only text description (that is an issue because when R checkbox is enabled we can’t see any reviews)

Current logic:
+D+R = both description and review tabs,
+D-R = description tab,
-D-R = text description,
-D+R = text description + expandable review section

#### Tickets / Documentation

https://jira.bigcommerce.com/browse/BCTHEME-40

Screenshot 2020-06-23 at 20 58 28
Screenshot 2020-06-23 at 20 58 43
Screenshot 2020-06-23 at 20 59 01
Screenshot 2020-06-23 at 20 59 15

ping @junedkazi @yurytut1993

### DIFF
--- a/config.json
+++ b/config.json
@@ -81,7 +81,7 @@
     "show_accept_googlepay": false,
     "show_accept_klarna": false,
     "show_product_details_tabs": true,
-    "show_product_reviews_tabs": false,
+    "show_product_reviews": true,
     "show_custom_fields_tabs": false,
     "show_product_weight": true,
     "show_product_dimensions": false,
--- a/templates/pages/product.html
+++ b/templates/pages/product.html
@@ -27,7 +27,7 @@ product:
             {{> components/products/videos product.videos}}
         {{/if}}
 
-        {{#all settings.show_product_reviews (if theme_settings.show_product_reviews_tabs '!==' true)}}
+        {{#all settings.show_product_reviews theme_settings.show_product_reviews (if theme_settings.show_product_details_tabs '!==' true)}}
             {{> components/products/reviews reviews=product.reviews product=product urls=urls}}
         {{/all}}
 
--- a/schema.json
+++ b/schema.json
@@ -1381,13 +1381,13 @@
         "type": "checkbox",
         "label": "i18n.ShowProductReviews",
         "force_reload": true,
-        "id": "show_product_reviews_tabs"
+        "id": "show_product_reviews"
       },
       {
         "type": "select",
         "label": "i18n.NumberOfProductReviews",
         "id": "productpage_reviews_count",
-        "reference": "show_product_reviews_tabs",
+        "reference": "show_product_reviews",
         "reference_default": false,
         "force_reload": true,
         "options": [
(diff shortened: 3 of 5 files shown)


config.json
```diff
@@ -81,7 +81,7 @@
     "show_accept_googlepay": false,
     "show_accept_klarna": false,
     "show_product_details_tabs": true,
-    "show_product_reviews_tabs": false,
+    "show_product_reviews": true,
```
junedkazi, line 84: I have a couple of questions here;
- Is there a reason to change the key name here specially when for product desc we still use `show_product_details_tabs` ? I like the old name was explicit and clear.
- Why are we changing the default from `false` to `true` ? Is this a product decision cc @bc-as  ?

templates/pages/product.html
```diff
@@ -27,7 +27,7 @@
             {{> components/products/videos product.videos}}
         {{/if}}
 
-        {{#all settings.show_product_reviews (if theme_settings.show_product_reviews_tabs '!==' true)}}
+        {{#all settings.show_product_reviews theme_settings.show_product_reviews (if theme_settings.show_product_details_tabs '!==' true)}}
```
junedkazi, line 30: Can we also remove the `'!==' true` here for the `show_product_details_tabs` ?

config.json
BC-tymurbiedukhin: * in theme builder we have checkbox with name "Show product reviews" but previous logic was depending on show_products_reviews_tab. Now logic have been changed and "Show product reviews" checkbox is responsible for product reviews display. So, for better code semantic I decided to change its name to current
* In case of false we won't see any product reviews, so in case of testing we need to set it to true anyway. It's not a decision of @bc-as

templates/pages/product.html
BC-tymurbiedukhin: Are there any more elegant way to check if the flag is falsy? If not, it will break a condition logic

config.json
bc-as: @junedkazi @BC-tymurbiedukhin 

- We should keep the same key here for consistency and to ensure that merchants don't have an issue when they update their theme.  That being said, the option in page builder/theme editor is "Show product reviews" and not "Show product reviews tab".
- Product reviews should be turned on by default unless the merchant wants to opt out

junedkazi:
@BC-tymurbiedukhin I will wait for a confirmation if this is tested and good to merge.

PR #1703, opened 2020-06-30
Suppress 3rd party Related Videos on YouTube videos

tipleyg:
This adds to product page YouTube videos a YouTube URL parameter that suppresses 3rd party "Related Videos" at the end of each video's run. During the June Town Hall Q&A, this was named by one customer as a problem that forced them to move their videos to Vimeo. With this URL parameter, only same-channel videos will appear as related videos.

For an Ecommerce site, this is an improvement and should reduce the possibility to sending traffic to someone else's store through YouTube 3rd party videos.

Since video.html is used by both AMP and non-AMP product pages, this should be effective on both.


tipleyg:
My apologies, this is my first pull request and it went all wrong.

PR #1708, opened 2020-07-01
feat(bctheme): BCTHEME-51 Optional Custom Options (Modifiers)

BC-tymurbiedukhin:
#### What?

For Custom Options (modifiers) that are not required for the shopper to fill out add the word "Optional" next to the field.

#### Tickets / Documentation

https://jira.bigcommerce.com/browse/BCTHEME-51

Screenshot 2020-07-01 at 19 37 02

ping @yurytut1993 @golcinho


templates/components/common/requireness-msg.html
```diff
@@ -0,0 +1,3 @@
+<small>
+    {{#if required}}{{lang 'common.required'}}{{else}}{{lang 'common.optional'}}{{/if}}
+</small>
```
junedkazi, line 3: Can we add an End of Line here.

templates/components/common/requireness-msg.html
```diff
@@ -0,0 +1,3 @@
+<small>
+    {{#if required}}{{lang 'common.required'}}{{else}}{{lang 'common.optional'}}{{/if}}
```
junedkazi, line 2: @bc-as are we sure we want to surface `Optional` as a flag across all not not required fields.

bc-as: @junedkazi Yes, let's add it.  We can re-evaluate this into a page builder toggle if necessary

templates/components/common/requireness-msg.html
BC-tymurbiedukhin: End of the line was added

templates/components/common/requireness-msg.html
junedkazi: @bc-as can we create a ticket to explore the `Optional` being `Optional` via Page Builder.

Issue #1677, opened 2020-05-15
Hero carousel images stretched on large screens


rowleyaj:
### Expected behavior
Setting `homepage_stretch_carousel_images` to false should prevent the carousel images being stretched on large screens

Help pages referenced:
https://support.bigcommerce.com/s/article/Carousel#carousel-dimensions
https://support.bigcommerce.com/s/article/Fixing-Distorted-Images#faq

### Actual behavior
It stretches the image and letterboxes due to `object-fit: cover` regardless of the setting

### Steps to reproduce behavior
Load page on a large browser window

---

I've tried to include further information below, as to what I believe the cause of this issue is:

The setting controls the addition of the class `stretch` in carousel.html:
https://github.com/bigcommerce/cornerstone/blob/d681ca929f24243a2284c15e4513fedc989686ec/templates/components/carousel.html#L13 

Searching for stretch in the repo:
```
$ ag stretch
config.json
57:    "homepage_stretch_carousel_images": false,
384:        "homepage_stretch_carousel_images": false,

schemaTranslations.json
926:  "i18n.AllowImageToStretchOn": {
927:    "default": "Allow image to stretch on large screens",

schema.json
980:        "label": "i18n.AllowImageToStretchOn",
982:        "id": "homepage_stretch_carousel_images"

templates/components/carousel.html
13:        <div class="heroCarousel-slide {{#if ../theme_settings.homepage_stretch_carousel_images}}stretch{{/if}} {{#if @first}}heroCarousel-slide--first{{/if}}">

assets/scss/components/stencil/heroCarousel/_heroCarousel.scss
103:        &.stretch {
```

The only place I can find this class configured is in `assets/scss/components/stencil/heroCarousel/_heroCarousel.scss`

https://github.com/bigcommerce/cornerstone/blob/d681ca929f24243a2284c15e4513fedc989686ec/assets/scss/components/stencil/heroCarousel/_heroCarousel.scss#L103-L107

but this compiles to a selector that requires `compat-object-fit`: `.heroCarousel-slide.compat-object-fit.stretch` so I'm not sure this is the right one to be looking at?

frontend code isn't my speciality so i might be way off the mark here, but this functionality appears to have been changed with this commit https://github.com/bigcommerce/cornerstone/commit/970d0e5ca774bcec5cf8226a6672b2d576fbac53

[...]

PR #1711, opened 2020-07-03
fix(Cornnerstone): BCTHEME-52 - Allowing carousel image to stretch distorts image

yurytut1993:
#### What?

homepage_stretch_carousel_images option from config.json didn’t work. We always had streched images and homepage_stretch_carousel_images value is not affected. Now it works - I have updated carousel scss file for to fix it.

Now images are stretching when homepage_stretch_carousel_images: true and contain in case when homepage_stretch_carousel_images: false.

Tested on Chrome, Safari, Firefox, ie11, IOS Safari and Android Chrome.

compat-object-fit class didn’t add in ie11 - I have updated it and now it works. homepage_stretch_carousel_images: false mode looks fine on ie11, homepage_stretch_carousel_images: true stretching images, but images looks distorted (as it is now on prod). Maybe we can find a way to update it in other task. Carousel looks okey even on small width on ie11.

Also I added some javascript to detect vertical or square images and make them looking more nicely and informative on mobile - if image is square or vertical and it has no content block in its slide, we have space to make image wrapper higher.

#### Tickets / Documentation

[Ticket](https://jira.bigcommerce.com/browse/BCTHEME-52)

In Ticket I added videos with prefix final for both homepage_stretch_carousel_images: true and homepage_stretch_carousel_images: false modes. So you can see the implementation. On videos we have the worst case of usage - vertical images, horizontal images, some with text and some without. And all them must be good looking in one carousel 

@BC-tymurbiedukhin @golcinho @junedkazi  ready for your questions


assets/scss/components/stencil/heroCarousel/_heroCarousel.scss
```diff
@@ -117,17 +169,21 @@
...
     text-align: center;
 
-    @include breakpoint("medium") {
+    @include breakpoint("small") {
         @include carouselOpaqueBackgrounds;
+        width: remCalc(700px);
+        padding: spacing("single") * 1.5;
+
+
+        width: 70%;
```
golcinho, line 178:
```suggestion
        width: 70%;
        padding: spacing("single") * 1.5;
```

assets/scss/components/stencil/heroCarousel/_heroCarousel.scss
```diff
@@ -79,35 +87,79 @@
...
+        height: 100%;
+        object-position: 50% 0%;
+
+        @include breakpoint("small") {
+            object-position: 50% 50%;
+        }
+    }
+
+    &.stretch {
+
```
golcinho, line 104:
```suggestion
```

assets/scss/components/stencil/heroCarousel/_heroCarousel.scss
```diff
@@ -79,35 +87,79 @@
...
     }
 
     .heroCarousel-image-wrapper {
+
+        height: remCalc(250px);
+        display: flex;
+        justify-content: center;
+        align-items: flex-start;
+
+        height: 56.25vw;
```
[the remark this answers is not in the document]
golcinho: you use `vw` (viewport width) in `height` property is it correct? maybe you need `vh` (viewport height)?

assets/js/theme/common/carousel.js
```diff
@@ -2,20 +2,44 @@ import 'slick-carousel';
...
-        $('.heroCarousel-slide').each((index, element) => {
-            const $container = $(element);
-            const imgUrl = $container.find('img').data('lazy');
-
-            if (imgUrl) {
-                $container.css('backgroundImage', `url(${imgUrl})`).addClass('compat-object-fit');
-            }
+    if (document.documentElement.style.objectFit === undefined) {
+        $slidesNodes.each((index, element) => {
+            $(element).addClass('compat-object-fit');
```
golcinho, line 42: Looks like you could move this code into the above `each` method.

assets/js/theme/common/carousel.js
```diff
@@ -2,20 +2,44 @@ import 'slick-carousel';
...
     const $carousel = $('[data-slick]');
-    if ($carousel.length) {
-        const multipleSlides = $carousel[0].childElementCount > 1;
-        $carousel.slick({ dots: multipleSlides });
-    }
+
+    if ($carousel.length === 0) return;
+
+    const multipleSlides = $carousel[0].childElementCount > 1;
+    $carousel.slick({ dots: multipleSlides });
```
golcinho, line 9: 🍹 You use `multipleSlides` variable only in one place, so, I think you could combine these two rows.

```
$carousel.slick({ dots: $carousel[0].childElementCount > 1 });
```

assets/js/theme/common/carousel.js
```diff
@@ -2,20 +2,44 @@ import 'slick-carousel';
...
+
+    const $slidesNodes = $('.heroCarousel-slide');
+
+    $slidesNodes.each((index, element) => {
```
golcinho, line 13: if you will merge two `each` methods, you will not need `$slidesNodes` variable

```suggestion
    $('.heroCarousel-slide').each((index, element) => {
```

assets/js/theme/common/carousel.js
```diff
@@ -2,20 +2,44 @@ import 'slick-carousel';
...
+
+                const isImageSquare = imageAspectRatio > 0.8 && imageAspectRatio <= 1.2;
+                const isImageVertical = imageAspectRatio > 1.2;
+
+                const slideClass = isImageSquare && 'is-square-image-type'
+                    || isImageVertical && 'is-vertical-image-type'
+                    || '';
+
+                $element.addClass(slideClass);
+            });
```
golcinho, line 36: 🍹 small proposal
```suggestion
            // You don't need to specify name of call back function in this case.
            .load(function () {
                const imageAspectRatio = this.height / this.width;
                
                switch (true) {
                    case imageAspectRatio > 0.8 && imageAspectRatio <= 1.2:
                        $element.addClass('is-square-image-type');
                        break;
                    case imageAspectRatio > 1.2:
                        $element.addClass('is-vertical-image-type');
                        break;
                }
            });
```

BC-tymurbiedukhin: My small proposal for small proposal:
```suggestion
            .load(function () {
                const imageAspectRatio = this.height / this.width;
                
                $element.addClass(() => {
                    switch(true){
                        case imageAspectRatio > 0.8 && imageAspectRatio <= 1.2:
                            return 'is-square-image-type';
                        case imageAspectRatio > 1.2:
                            return 'is-vertical-image-type'
                        default: 
                            return ''
                    }
                })
            });
```

golcinho: Good idea, but I think that adding of the empty row is a little bit redundant.

BC-tymurbiedukhin: addClass method interface expects string as classname type, so in my opinion it's good to follow it

golcinho: You right, but why we need to call `addClass` method for the case when we don't need to add a class name? 🤔 

I marked this comment as 🍹 (optional), so I do not have a strict position on this piece of code. Anyway, if this code will be optimized I will be happy 😄

BC-tymurbiedukhin: Thank you for a great review!)

assets/scss/components/stencil/heroCarousel/_heroCarousel.scss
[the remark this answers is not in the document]
yurytut1993: we really need **vw** here - this is for aspect ratio 16x9

assets/js/theme/common/carousel.js
yurytut1993: I consciously separated this 2 loops. It won't have an influence for performance, but will make code more understandable. You will not have ie11 unnecessary checks in main loop, so it is easier to read. Also the second loop will  work only in ie11 and similar, so form most of browsers in it better even in context of performance. The most expensive operation is finding nodes - that is why I saved it in **$slidesNodes** variable.  But if you insist of course I will combine all logic in one loop

assets/js/theme/common/carousel.js
yurytut1993: Thanks for your proposals!

yurytut1993:
@golcinho  @BC-tymurbiedukhin 
To be honest I suppose that my version of implementation is little bit more readable, but you both voted for switch, so let it be switch.
Also I added some new logic. As you may remember for vertical and square images we make container bigger. The problem is images loading faster then this code works and image is jumping when its height changes. I recorded video (bug_image_jump) and put it to the [ticket](https://jira.bigcommerce.com/browse/BCTHEME-52) - better one time to see. New logic fix this problem (fix_image_jump)

junedkazi:
https://github.com/bigcommerce/cornerstone/issues/1677

PR #1712, opened 2020-07-03
fix(Cornerstone): BCTHEME-2 Sale badge "Burst" does not change colour on hover

yurytut1993:
#### What?
Cornerstone has 3 types of sale badge - sash, topleft and burst. On hover product with sale badge it must change color that user can setup in storefront panel (color_hover_product_sale_badges settings option in config.json). sash & topleft affecting this value, but burst not. I have added styles for burst sale badge and now hover works on all sale badges

@BC-tymurbiedukhin  @golcinho @junedkazi 

#### Tickets / Documentation

[Ticket](https://jira.bigcommerce.com/browse/BCTHEME-2)

### DIFF
--- a/assets/scss/layouts/products/_productSaleBadges.scss
+++ b/assets/scss/layouts/products/_productSaleBadges.scss
@@ -64,6 +64,16 @@
     z-index: zIndex("high");
 }
 
+.product:hover .starwrap {
+    
+    .sale-flag-star,
+    .sale-flag-star:before,
+    .sale-flag-star:after {
+        background: stencilColor("color_hover_product_sale_badges");
+        transition: 800ms ease;        
+    }    
+}
+
 // -----------------------------------------------------------------------------
 // TOP LEFT BADGE
 // -----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


assets/scss/layouts/products/_productSaleBadges.scss
```diff
@@ -64,6 +64,16 @@
     z-index: zIndex("high");
 }
 
+.product:hover .starwrap {
+    
+    .sale-flag-star,
+    .sale-flag-star:before,
+    .sale-flag-star:after {
+        background: stencilColor("color_hover_product_sale_badges");
+        transition: 800ms ease;
```
golcinho, line 73: I recommend moving the `transition` property to the place where declaring the main styles of these selectors.

Also, you need to specify `transition-property`, the next example would be helpful

```
transition: background-color 800ms ease;
```

yurytut1993: Update scss file following through your comments. We have 3 types of sale badges as I mentioned earlier, so I applied such changes for other badges too

junedkazi:
@yurytut1993 the changes look sane. Can you please attach some screenshots to show before and after.

yurytut1993:
> @yurytut1993 the changes look sane. Can you please attach some screenshots to show before and after.

@junedkazi in the [ticket](https://jira.bigcommerce.com/browse/BCTHEME-2) you can see 2 videos with bug and fix

PR #1714, opened 2020-07-07
fix(bctheme): BCTHEME-20 Stencil Themes - Non-required options do not…

BC-tymurbiedukhin:
… have a 'none' choice on the storefront

#### What?

Historically, when product options were set to not required a shopper would have the ability to select 'none' on the option. Stencil themes not provide this choice, so if a shopper selects an option that isn't required, they do not have the ability to de-select it by choosing 'none'. This can be frustrating for shoppers, especially when there are price change rules attached to the non-required option.

#### Tickets / Documentation

https://jira.bigcommerce.com/browse/BCTHEME-20

Screenshot 2020-07-07 at 13 18 20

ping @yurytut1993 @golcinho @junedkazi

### DIFF
--- a/assets/js/theme/common/utils/api.js
+++ b/assets/js/theme/common/utils/api.js
@@ -1,0 +1,49 @@
+/**
+ * This function removes any empty string values from the formData
+ * @param formData: FormData object
+ * @returns FormData object
+*/
+export const filterEmptyValuesFromForm = formData => {
+    const res = new FormData();
+
+    try {
+        for (const [key, val] of formData) {
+            if (val !== '') {
+                res.append(key, val);
+            }
+        }
+    } catch (e) {
+        console.log(e); // eslint-disable-line no-console
+    }
+
+    return res;
+};
+
+/**
+ * https://stackoverflow.com/questions/49672992/ajax-request-fails-when-sending-formdata-including-empty-file-input-in-safari
+ * Safari browser with jquery 3.3.1 has an issue uploading empty file parameters. This function removes any empty files from the form params
+ * @param formData: FormData object
+ * @returns FormData object
+ */
+export const filterEmptyFilesFromForm = formData => {
+    const res = new FormData();
+
+    try {
+        for (const [key, val] of formData) {
+            if (!(val instanceof File) || val.name || val.size) {
+                res.append(key, val);
+            }
+        }
+    } catch (e) {
+        console.error(e); // eslint-disable-line no-console
+    }
+
+    return res;
+};
+
+/**
+ * This function removes empty string values and empty files from the formData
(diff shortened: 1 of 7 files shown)


templates/components/products/options/set-rectangle.html
```diff
@@ -6,6 +6,21 @@
...
         {{/if}}
     </label>
+
+    {{#unless required}}
+        <input
+            class="form-radio"
+            type="radio"
+            id="attribute_rectangle__{{../id}}_none"
+            name="attribute[{{../id}}]"
+            value=""
```
golcinho, line 16: Should it be empty? or it should be `value="{{lang 'products.none'}}"`?

templates/components/products/options/swatch.html
```diff
@@ -8,6 +8,20 @@
         {{/if}}
     </label>
 
+    {{#unless required}}
+        <input
+            class="form-radio"
+            type="radio"
+            name="attribute[{{../id}}]"
+            value=""
```
golcinho, line 16: same as above

templates/components/products/options/set-rectangle.html
BC-tymurbiedukhin: I see, but I have wrote a util function which prevents sending empty values to backend. Change its work with 'none' value will make it not reusable

golcinho: Got it, thank you for clarifying 🙇

assets/js/theme/common/utils/api.js
```diff
@@ -0,0 +1,49 @@
...
+ * Safari browser with jquery 3.3.1 has an issue uploading empty file parameters. This function removes any empty files from the form params
+ * @param formData: FormData object
+ * @returns FormData object
+ */
+export const filterEmptyFilesFromForm = formData => {
+    const res = new FormData();
+
+    try {
+        for (const [key, val] of formData) {
+            if (!(val instanceof File) || val.name || val.size) {
```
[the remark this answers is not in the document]
BC-tymurbiedukhin: This is util function which can be used by its own because I believe sending empty values not only issue for product options fields

assets/js/theme/common/utils/api.js
```diff
@@ -0,0 +1,49 @@
+/**
+ * This function removes any empty string values from the formData
+ * @param formData: FormData object
+ * @returns FormData object
+*/
+export const filterEmptyValuesFromForm = formData => {
+    const res = new FormData();
```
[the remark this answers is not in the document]
BC-tymurbiedukhin: This is pure function which makes result of executing predictable and easy testable. I have faced some issues with changing form data by link so decided to make it in this way

PR #1715, opened 2020-07-08
feature(bctheme): BCTHEME-50 ability to load subcategories and expand them if there are other categories beneath it

yurytut1993:
#### What?

@BC-tymurbiedukhin @junedkazi @golcinho 

In Cornerstone we have mobile menu with nested categories if they exist. For now we can not go to the subcategory page if it has its own subcategories (has arrow to dropdown). I have created a wrapper for arrow and replaced data attributes from link on this wrapper. Javascript uses this attributes to open/close nested subcategories. 

So now link is responsible for redirect and arrow responsible for open/close nested subcategories. Also I updated icon wrapper styles to stretch it to the end of the row for comfortable toggling nested categories. 

Styles wrote carefully and not affecting alternate menu styles (by default in config.json navigation_design option value is simple).

#### Tickets / Documentation

In [Ticket](https://jira.bigcommerce.com/browse/BCTHEME-50) you can see 2 videos (before_menu_links_update and after_menu_links_update) to better understand the result

### DIFF
--- a/assets/scss/components/stencil/navPages/_navPages.scss
+++ b/assets/scss/components/stencil/navPages/_navPages.scss
@@ -240,7 +240,7 @@
 
 .navPages-action-moreIcon {
     @include square(8px);
-    margin-left: spacing("quarter");
+    margin-left: spacing("half");
     transform: rotate(-90deg);
 
     @include breakpoint("medium") {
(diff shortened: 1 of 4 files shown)


yurytut1993:
> Everything is ok, but I suggest to move arrow on the right side of menu, @bc-as what do you think?

it will not be a problem

yurytut1993:
> Everything is ok, but I suggest to move arrow on the right side of menu, @bc-as what do you think?

@BC-tymurbiedukhin moved arrow to the right. made some code updates for it

Screenshot 2020-07-08 at 18 42 05

yurytut1993:
@BC-tymurbiedukhin @bc-as @golcinho 

made updates for mobile menu styles that were discussed with designers 

Screenshot 2020-07-13 at 12 43 55

bc-as:
@yurytut1993 lgtm

PR #1722, opened 2020-07-13
fix(bctheme): BCTHEME-8 Cornerstone v 4.1.1 - Product Image dimension…

BC-tymurbiedukhin:
…s not constrained proportionally on Cart page which can lead to element overlapping

#### What?

On the latest version of Cornerstone, larger product images are failing to be constrained proportionally to their parent div on the Cart page. This can lead to element overlapping with other line-items in the cart.

Expanding this ticket to mention that these symptoms are also present in the Cart Quick-View modal.

#### Tickets / Documentation

https://jira.bigcommerce.com/browse/BCTHEME-8

Screenshot 2020-07-13 at 12 27 07
Screenshot 2020-07-13 at 12 22 40
Screenshot 2020-07-13 at 12 22 51
Screenshot 2020-07-13 at 12 23 08

ping @yurytut1993 @bc-alexsaiannyi @junedkazi @golcinho

### DIFF
--- a/assets/scss/components/foundation/lazyLoad/_lazyLoad.scss
+++ b/assets/scss/components/foundation/lazyLoad/_lazyLoad.scss
@@ -16,3 +16,15 @@
         padding-bottom: get-padding(stencilString($size-param));
     }
 }
+
+@mixin cart-item-lazy-load-img-placeholder() {
+    &:after {
+        content: '';
+        display: block;
+        position: absolute;
+        top: 0;
+        left: 0;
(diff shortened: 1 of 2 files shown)


BC-tymurbiedukhin:
Changes has been applied for classes which are used only in components/cart/content.html file

junedkazi:
@BC-tymurbiedukhin I understand that the changes are specific to cart. But I am wondering if we need to make this change globally instead of doing a targeted fix.

PR #1723, opened 2020-07-13
feature(bctheme): BCTHEME-56 Long Product Descriptions Cause Large White Space Under Product Image

yurytut1993:
#### What?

To avoid space on the left side of the screen when we have a huge description on a product, as suggested in the task, I made description block on the whole screen. So I just removed styles that make this block on the half of the screen width on large displays

#### Tickets / Documentation
[Ticket](https://jira.bigcommerce.com/browse/BCTHEME-56)

before
before-bc-56

after
after-bc-56

### DIFF
--- a/assets/scss/layouts/products/_productView.scss
+++ b/assets/scss/layouts/products/_productView.scss
@@ -52,12 +52,6 @@
     position: relative;
     z-index: 1;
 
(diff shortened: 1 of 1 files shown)


bc-as:
@junedkazi Yes, we're going to make this consistent at all screen widths

PR #1726, opened 2020-07-15
feat(bctheme): BCTHEME-38 Cornerstone - Add Client Validation for Pho…

BC-tymurbiedukhin:
…ne Number on Account Creation/Edit Pages

#### What?

The new validation confirms that after removing -"_()/:#.,'*\s user input will match with E164 phone number standard, but with at least 3 numbers

#### Tickets / Documentation

https://jira.bigcommerce.com/browse/BCTHEME-38

Screenshot 2020-07-15 at 16 37 23
Screenshot 2020-07-15 at 16 37 33

ping @junedkazi @yurytut1993 @bc-alexsaiannyi @golcinho

### DIFF
--- a/assets/js/theme/account.js
+++ b/assets/js/theme/account.js
@@ -313,6 +313,7 @@ export default class Account extends PageManager {
         const formEditSelector = 'form[data-edit-account-form]';
         const editValidator = nod({
             submit: '${formEditSelector} input[type="submit"]',
+            delay: 0,
         });
         const emailSelector = `${formEditSelector} [data-field-type="EmailAddress"]`;
         const $emailElement = $(emailSelector);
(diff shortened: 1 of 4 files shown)


BC-YIvashchenko (approved):
Looks good to me

junedkazi (commented):
Is there a reason why we are trying to add validation here instead of just switching over to html5 input type `tel` https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/tel

BC-tymurbiedukhin:
Pattern attribute of html5 tag isn't available to validate complex logic, I have investigated this question and found that the only way to let validation works in different countries to create custom regexp logic. Lmk if there is a workaround, I will be glad to have a look on it

aleachjr:
I think this will cause issues as we have some merchants who use this field to store non phone numbers. I would talk to Derrek Pearson and see if he remembers. The last time we did this on customers team it had to be rolled back as people had saved data that broke checkout.

golcinho (approved):

BC-tymurbiedukhin:
@aleachjr what is current status for this issue?

PR #1727, opened 2020-07-16
feature(themes): BCTHEME-76 Carousel bullets should be buttons (paging func update)

yurytut1993:
#### What?

@BC-tymurbiedukhin @junedkazi @bc-alexsaiannyi 

I added buttons in dots list of slick-carousel. As I understood from library code we can not use dots not in the <ul> tag. In screenshots I added the proof of it - <ul> tag is hardcoded in function where dots are creating. But as I understand from [slick demos](https://kenwheeler.github.io/slick/) this is not a problem to create accessible dots. Now buttons in dots list can be used for sliding carousel. Buttons have no visual focus now - it will be implemented in one of the next tasks.

Also I turned off accessibility of slick, because current version of the library has a bug with setting tabindex="-1" for all inactive bullets. So you can not use tabbing to jump from one bullet to other. Also by default slick adds role="tabs" for bullets container which is not correct. Because of this bugging behaviour it is better to turn off slick's accessibility and make it by our own. So I added function that controls tabindex of slides (0 for active slide and -1 for other). And dot buttons now have default semantic behaviour as we need 

#### Tickets / Documentation

[Ticket](https://jira.bigcommerce.com/browse/BCTHEME-76)

slick <ul> tag hardcoded for dots
proof

current dot button with wrong role and tabindex
wrong-button

new button that has correct aria role and tabindex by default
semantic-button


assets/js/theme/common/carousel.js
```diff
@@ -14,7 +24,24 @@ export default function () {
 
     if ($carousel.length === 0) return;
 
-    $carousel.slick({ dots: $carousel[0].childElementCount > 1 });
+    const isMultipleSlides = $carousel[0].childElementCount > 1;
+
+    const slickSettingsObj = isMultipleSlides
+        ? {
+            dots: true,
+            customPaging: () => (
```
BC-tymurbiedukhin, line 32: Why do we need to return empty button here?

yurytut1993: For current ticket it is not absolutely necessary, because slick will generate it by default. But I still added it manually for 2 reasons: to make obvious that we have buttons here, and the main - in the next implementations we need to set aria-labels for this generated dot buttons, so we will need this function for it. But if you insist I can remove customPaging function for now

BC-tymurbiedukhin: It's fine, thank you for clarification

PR #1730, opened 2020-07-21
Optimize themeSettings injection

bookernath:
#### What?

We currently inject the entire `theme_settings` context object which is wasteful, especially since we only use 3 keys from it. I've optimized this import to be [only what's used.](https://github.com/bigcommerce/cornerstone/search?q=themeSettings&unscoped_q=themeSettings)

Before:
image

After:
image


BC-tymurbiedukhin:
Great, thank you

mattolson:
👏

PR #1733, opened 2020-07-22
As a shopper I want to have the quick view modal immediately have keyboard focus

yurytut1993:
#### What?

@BC-tymurbiedukhin @golcinho @bc-alexsaiannyi 

As a shopper I want to have the quick view modal immediately have keyboard focus.

Now when modal opening focus goes to its first focusable element - close button as usual. By the way this close button is a link tag now with button role. As I know it is better practice to use native tags, what do you think, maybe we should change it?

When user has focus  on last focusable element and tabbing - he will be redirected to the first. And when user on the first focusable element and tabbing back using shift - he will be redirected to the last element. That is focus trap we need.

Each tab press on modal is under js control. We have different layouts in the modal. For example nested modal with id modal-review-form. To be sure that user will not focus on unexpected element each tabbing is under control while modal opened.

Also not all focusable elements can be focused because of their own issues. By default browsers skip such elements and focusing on the next valid focusable element. This behaviour is also repeated

#### Tickets / Documentation

[Ticket](https://jira.bigcommerce.com/browse/BCTHEME-73)

tabbing demonstration (focused styles in this video added just for better visibility of focused element)
[ezyzip.zip](https://github.com/bigcommerce/cornerstone/files/4959086/ezyzip.zip)


assets/js/theme/global/modal.js
```diff
@@ -196,12 +210,65 @@ export class Modal {
...
+
+        const $collection = focusableElements[modalType]();
+        $collection.each((index, element) => $(element).removeClass(focusedClass)); 
+        const $elementToFocus = $($collection.get(0));
+        $elementToFocus.focus().addClass(focusedClass);
+
+        $('#modal').on('keydown', this.onTabbing.bind(null, modalType));
+    }
+
+    onTabbing(modalType, event) {
```
BC-tymurbiedukhin, line 224: I suggest to use currying for onTabbing func modalType => event => {}

assets/js/theme/global/modal.js
```diff
@@ -196,12 +210,65 @@ export class Modal {
...
 
+    setupFocusableElements(modalType) {
+        this.$lastPageFocusedElement = $(document.activeElement);
+
+        const $collection = focusableElements[modalType]();
+        $collection.each((index, element) => $(element).removeClass(focusedClass)); 
+        const $elementToFocus = $($collection.get(0));
+        $elementToFocus.focus().addClass(focusedClass);
+
+        $('#modal').on('keydown', this.onTabbing.bind(null, modalType));
```
BC-tymurbiedukhin, line 221: Change use for curried func to this.onTabbing(modalType)

assets/js/theme/global/modal.js
```diff
@@ -196,12 +210,65 @@ export class Modal {
...
+
+        if (!isTAB) return;
+
+        const $collection = focusableElements[modalType]();
+        const collectionLastIdx = $collection.length - 1
+
+        const $currentElement =  $collection.filter(`.${focusedClass}`);
+        const currentElementIdx = $collection.index($currentElement);
+
+        const direction = event.which === TAB && event.shiftKey ? 'backwards' : 'forwards';
```
BC-tymurbiedukhin, line 235: Direction var could be changed to boolean type

assets/js/theme/global/modal.js
```diff
@@ -196,12 +210,65 @@ export class Modal {
...
+
+    onTabbing(modalType, event) {
+        const isTAB = event.which === TAB;
+
+        if (!isTAB) return;
+
+        const $collection = focusableElements[modalType]();
+        const collectionLastIdx = $collection.length - 1
+
+        const $currentElement =  $collection.filter(`.${focusedClass}`);
```
BC-tymurbiedukhin, line 232: Better to use "each" to get the element and index during one iteration

assets/js/theme/global/quick-view.js
```diff
@@ -22,6 +22,8 @@ export default function (context) {
 
             modal.$content.find('[data-slick]').slick();
 
+            modal.setupFocusableElements('forQuickView');
```
BC-tymurbiedukhin, line 25: Use modalType map value instead of srtring

assets/js/theme/global/modal.js
```diff
@@ -196,12 +210,65 @@ export class Modal {
...
+        } else if (direction === 'backwards') {
+            nextElementIdx = currentElementIdx === 0
+                ? collectionLastIdx
+                : currentElementIdx - 1;
+            $candidatesCollection = $($collection.slice(0, nextElementIdx + 1).get().reverse());
+        }
+
+        $candidatesCollection.each((index, element) => {
+            const $focusCanditate = $(element);
+            $focusCanditate.focus();
```
BC-tymurbiedukhin, line 253: Why do we need to trigger focus event handler on every collection element?

assets/js/theme/global/modal.js
```diff
@@ -196,12 +210,65 @@ export class Modal {
         this.$content.html('');
     }
 
+    setupFocusableElements(modalType) {
+        this.$lastPageFocusedElement = $(document.activeElement);
```
BC-tymurbiedukhin, line 214: Better to name like preModalFocusedEl

assets/js/theme/global/modal.js
yurytut1993: Because not all elements in focusable elements collection truly can be focused - because of their own reasons. For example in video I added we have wishlist form. Inputs in it should be focusable, but they are not. And when you trying to set focus to first of this input - focus wont affect to this element and will stay on previous. But in such case we need to focus on next focusable element. So I am trying to find first truly focusable element in collection

assets/js/theme/global/quick-view.js
yurytut1993: Do you mean to create mapping object in modal.js and using it in quick-view.js after creating a modal?

assets/js/theme/global/modal.js
yurytut1993: I can not. Because in that case first function argument will be event and modalType will be lost

assets/js/theme/global/quick-view.js
BC-tymurbiedukhin: It's about map with modal types constants (example ModalTypes.QUICK_VIEW)

assets/js/theme/global/modal.js
BC-tymurbiedukhin: I think in this case we need to change the logic of collection creation, we need create a collection of truly focusable elements

yurytut1993: Maybe. But in that new logic we will need to check all elements parents on aria-hidden, tabindex=-1 and maybe other attributes to detect they are not focusable. It can be even more painful for performance and readability

BC-tymurbiedukhin: In result of call with Yury, we have made a decision to proceed with current algorithm

assets/js/theme/global/modal.js
```diff
@@ -4,13 +4,26 @@ const bodyActiveClass = 'has-activeModal';
...
+const focusedClass = 'focused';
 
 const SizeClasses = {
     small: 'modal--small',
     large: 'modal--large',
     normal: '',
 };
 
+const focusableElements = {
+    'forQuickView': () => $('#modal')
```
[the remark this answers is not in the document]
yurytut1993: hope I understood you correctly. please check

assets/js/theme/global/modal.js
```diff
@@ -4,13 +4,26 @@ const bodyActiveClass = 'has-activeModal';
 const loadingOverlayClass = 'loadingOverlay';
 const modalBodyClass = 'modal-body';
 const modalContentClass = 'modal-content';
+const focusableElementsSelector = 'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])';
+const uselessFocusableElementsSelector = '[tabindex="-1"], [type="hidden"]'
+
+const TAB = 9;
```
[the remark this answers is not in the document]
yurytut1993: agree. done

assets/js/theme/global/modal.js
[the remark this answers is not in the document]
yurytut1993: done

assets/js/theme/global/modal.js
yurytut1993: agree. done

assets/js/theme/global/modal.js
yurytut1993: done

assets/js/theme/global/modal.js
yurytut1993: If you don't mind I will leave it in such way. Suppose that it is little bit more readable

assets/js/theme/global/modal.js
yurytut1993: updated code. And now it is even not necessary performance will be little bit better

assets/js/theme/global/modal.js
yurytut1993: agree. done

assets/js/theme/global/modal.js
```diff
@@ -4,13 +4,29 @@ const bodyActiveClass = 'has-activeModal';
...
+const tabKeyCode = 9;
 
 const SizeClasses = {
     small: 'modal--small',
     large: 'modal--large',
     normal: '',
 };
 
+const focusableElements = {
+    'forQuickView': () => $('#modal')
```
BC-tymurbiedukhin, line 19: Now you could use [ModalTypes.QUICK_VIEW]: () => $('#modal')...

assets/js/theme/global/modal.js
```diff
@@ -4,13 +4,29 @@ const bodyActiveClass = 'has-activeModal';
 const loadingOverlayClass = 'loadingOverlay';
 const modalBodyClass = 'modal-body';
 const modalContentClass = 'modal-content';
+const focusableElementsSelector = 'button, [href], input, select, textarea, [tabindex]';
+const uselessFocusableElementsSelector = '[tabindex="-1"], [type="hidden"]'
```
BC-tymurbiedukhin, line 8: Better to use inactive instead of useless

assets/js/theme/global/modal.js
```diff
@@ -196,12 +213,89 @@ export class Modal {
...
+        const isValidElementActive = currentElementIdx !== -1;
+
+        let startingPoint;
+        if (isValidElementActive) {
+            startingPoint = currentElementIdx
+        } else if (direction === 'forwards') {
+            startingPoint = collectionLastIdx;
+        } else if (direction === 'backwards') {
+            startingPoint = 0;
+        }
```
BC-tymurbiedukhin, line 267: For me, it's look like a separate small function with switch(true){} statement if you want to use string direction var instead of boolean

assets/js/theme/global/modal.js
```diff
@@ -196,12 +213,89 @@ export class Modal {
...
+
+        const $collection = focusableElements[modalType]();
+        const collectionLastIdx = $collection.length - 1
+
+        const $currentElement = $(document.activeElement);
+        const currentElementIdx = $collection.index($currentElement);
+
+        const direction = event.which === tabKeyCode && event.shiftKey ? 'backwards' : 'forwards';
+        
+        /* to jump to the first or last element (depends on direction)
```
BC-tymurbiedukhin, line 255: Why do we need to jump on start or the end if we can find the nearest focusable element depending on direction

assets/js/theme/global/modal.js
```diff
@@ -196,12 +213,89 @@ export class Modal {
...
+                return false;
+            };
+            // if no appropriate candidate - find appropriate in full collection
+            if (index === $candidatesCollection.length - 1 && $reserveCollection) {
+                this.findElementToFocus($reserveCollection);
+            };
+        });
+    }
+
+    onTabbing(event, modalType, findElementToFocus) {
```
BC-tymurbiedukhin, line 242: Why do we need to pass class method in this case? And currying here is still a better idea https://learn.javascript.ru/currying-partials

assets/js/theme/global/modal.js
```diff
@@ -196,12 +213,89 @@ export class Modal {
...
     }
 
+    setupFocusableElements(modalType) {
+        this.$preModalFocusedEl = $(document.activeElement);
+        
+        const $collection = focusableElements[modalType]();
+
+        this.findElementToFocus($collection);
+
+        const onTabbingWithArgs = (event) => this.onTabbing(event, modalType, this.findElementToFocus.bind(this));
```
BC-tymurbiedukhin, line 223: We can pass onTabbingWithArgs func directly in event listener, no need in separate creating

assets/js/theme/global/modal.js
```diff
@@ -196,12 +216,64 @@ export class Modal {
...
+    setupFocusableElements(modalType) {
+        this.$preModalFocusedEl = $(document.activeElement);
+
+        const $collection = focusableElements[modalType]();
+        $collection.get(0).focus();
+
+        $('#modal').on('keydown', (event) => this.onTabbing(event, modalType));
+    }
+
+    onTabbing(event, modalType) {
```
BC-tymurbiedukhin, line 228: Still better to use currying)

PR #1737, opened 2020-07-23
fix(storefront): BCTHEME-126 'Write a Review' should have role 'button'

BC-tymurbiedukhin:
#### What?

On PDPs the "Write a Review" link should be labeled as a "button" rather than a "link".

#### Tickets / Documentation

[Ticket](https://jira.bigcommerce.com/browse/BCTHEME-126)

Screenshot 2020-07-23 at 14 19 25

### DIFF
--- a/templates/components/products/product-view.html
+++ b/templates/components/products/product-view.html
@@ -50,12 +50,14 @@
                     </span>
                 {{/if}}
                 {{#if settings.show_product_reviews}}
-                    <span class="productView-reviewLink">
+                    <button class="productView-reviewLink">
                         <a href="{{product.url}}{{#if is_ajax}}#write_review{{/if}}"
-                           {{#unless is_ajax }}data-reveal-id="modal-review-form"{{/unless}}>
+                           {{#unless is_ajax}}data-reveal-id="modal-review-form"{{/unless}}
+                           role="button"
+                        >
                            {{lang 'products.reviews.new'}}
                         </a>
-                    </span>
+                    </button>
                     {{> components/products/modals/writeReview}}
                 {{/if}}
             </div>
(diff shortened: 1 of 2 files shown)


templates/components/products/product-view.html
```diff
@@ -50,9 +50,11 @@ <h2 class="productView-brand"{{#if schema}} itemprop="brand" itemscope itemtype=
                     </span>
                 {{/if}}
                 {{#if settings.show_product_reviews}}
-                    <span class="productView-reviewLink">
+                    <span class="productView-reviewLink" role="button">
```
bc-alexsaiannyi, line 53: Are there any limitations for using a button instead of span ?

BC-tymurbiedukhin: No limitations, have been changed, thank you

templates/components/products/product-view.html
```diff
@@ -50,12 +50,14 @@ <h2 class="productView-brand"{{#if schema}} itemprop="brand" itemscope itemtype=
...
+                    <button class="productView-reviewLink">
                         <a href="{{product.url}}{{#if is_ajax}}#write_review{{/if}}"
-                           {{#unless is_ajax }}data-reveal-id="modal-review-form"{{/unless}}>
+                           {{#unless is_ajax}}data-reveal-id="modal-review-form"{{/unless}}
+                           role="button"
+                        >
                            {{lang 'products.reviews.new'}}
                         </a>
-                    </span>
+                    </button>
```
yurytut1993, line 60: We have link nested in button here. As I know it can be issue for Firefox. Can you please check it? 
And here is the link that also can be helpful
https://stackoverflow.com/questions/16280684/nesting-a-inside-button-doesnt-work-in-firefox
Inside button our link also have role button. Maybe we can just use span for productView-reviewLink and change inner link to button?

BC-tymurbiedukhin: Tested in FF 78 and it works fine. In this case we need to use link because of href attribute usage and set role='button' for correct voiceover handling

PR #1738, opened 2020-07-23
feature(themes): BCTHEME-71 Add To Cart & Quick View buttons now appear when tabbing on product cards

yurytut1993:
#### What?

@BC-tymurbiedukhin  @bc-alexsaiannyi 

Add To Cart & Quick View buttons now appear when tabbing on product cards. Unfortunately this feature can not be done using only css because we can not affect parent elements. So when we have one button focused we can not remove transparency from class card-figcaption-body to make all buttons visible.

So I created event listener to check if focused element in card and show card-figcaption-body if it it true.

Also I changed tag of quick view button from <a> to <button>, because it is truly button

#### Tickets / Documentation

[Ticket](https://jira.bigcommerce.com/browse/BCTHEME-71)

[card_focus.zip](https://github.com/bigcommerce/cornerstone/files/4965499/card_focus.zip)


assets/js/theme/common/product-card.js
```diff
@@ -0,0 +1,30 @@
...
+
+    const isFocusInCard = $cardContainer.length;
+
+    const { $previous } = previousCard;
+    
+    if (isFocusInCard) {
+        $previous && $previous.removeClass(focusedCardClass); 
+        $cardContainer.addClass(focusedCardClass);
+        previousCard.$previous = $cardContainer;
+        return;
```
BC-tymurbiedukhin, line 20: I would be nice to remove 17 and 20 code lines

assets/js/theme/common/product-card.js
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    $previous && $previous.removeClass(focusedCardClass);
+};
+
+export default function () {
+    const $allFocusableElements = $(focusableElementsSelector);
```
BC-tymurbiedukhin: Maybe it would be better idea to listen focus events only on card focusable elements?

assets/js/theme/common/product-card.js
```diff
@@ -0,0 +1,30 @@
+const focusedCardClass = 'focused-card';
+const focusableElementsSelector = '[href], button, input, textarea, select, details, [tabindex]';
+const previousCard = {
```
BC-tymurbiedukhin, line 3: I can see only usage of "$previous" link. Why do we need object in this case?

assets/js/theme/common/product-card.js
yurytut1993: > Generally, I think https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-within would be helpful in this situation (polyfill for IE will be needed)

Maybe I can add focus-within, detect IE and use my script in ie like polifill?

assets/js/theme/common/product-card.js
yurytut1993: No, because we need to hide buttons when we focusing not inside the card

assets/js/theme/common/product-card.js
yurytut1993: if we will use simple variable - we will scope its value, but we need to scope the link to the value for repeat usage

assets/js/theme/common/product-card.js
yurytut1993: > > Generally, I think https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-within would be helpful in this situation (polyfill for IE will be needed)
> 
> Maybe I can add focus-within, detect IE and use my script in ie like polifill?

test my code in ie 11 and it works

assets/js/theme/common/product-card.js
BC-tymurbiedukhin: It does not depend or what we are using, it depends on data type that we write into variable

assets/js/theme/common/product-card.js
BC-tymurbiedukhin: I prefer https://www.npmjs.com/package/focus-within-polyfill

assets/js/theme/common/product-card.js
yurytut1993: yes, that is why I added object as variable value

BC-tymurbiedukhin: It is unnecessary as you write jquery object in a $previous var

yurytut1993: at the beginning I have no $previous element. Of course I can use {} instead of null, but null easier to check so I decided to do it in react useRef style:) 
but anyway now it is removed - polyfil is realy better idea. thanks)

Issue #1705, opened 2020-06-30
PDP image switching not working on IE11


chippleh1392:
### Expected behavior
On the product page, clicking an image thumbnail should swap the main image to the selected thumbnail image.

### Actual behavior
On IE11, clicking the thumbnail results in nothing happening

### Steps to reproduce behavior
On IE11, go to any product page that has more than 1 image, click a thumbnail image that is different from the currently selected image.

### Comments
I investigated this and it appears that IE11's way of setting the main image using inline css (background-image property) is not compatible with the EasyZoom swap() function which is called in the swapMainImage() function in image-gallery.js. EasyZoom is specifically looking to update the image's src and srcset attributes. I implemented a fix just using an IE11 detection check and updating the needed image attributes with jQuery. 

    swapMainImage() {
        // ie11 fix
        let isIE11 = !!window.MSInputMethodContext && !!document.documentMode; // detect if ie11
        
        if (isIE11 == true) {
            this.$mainImage.find('img').css('background-image', `url('${this.currentImage.mainImageUrl}')`);
            this.$mainImage.find('a').attr('href', this.currentImage.mainImageUrl);
        } else { // if not ie11, perform easyzoom swap
            this.easyzoom.data('easyZoom').swap(
                this.currentImage.mainImageUrl,
                this.currentImage.zoomImageUrl,
                this.currentImage.mainImageSrcset,
            );
        }

        this.$mainImage.attr({
            'data-zoom-image': this.currentImage.zoomImageUrl,
        });
    }

PR #1748, opened 2020-07-28
fix(storefront):fix product Images don't change on click in IE11

bc-alexsaiannyi:
#### What?

@BC-tymurbiedukhin @yurytut1993 
this PR fixes issue with viewing a Product with multiple images and clicking on one of its images in Internet Explorer 11 , it appears that image doesn't change.

#### Tickets / Documentation

- [BCTHEME-140](https://jira.bigcommerce.com/browse/BCTHEME-140)

[IE11 no images changes before and after fix video](https://github.com/bigcommerce/cornerstone/files/4989688/IE11.no.images.changes.before_after.fix.zip)


assets/js/theme/product/image-gallery.js
```diff
@@ -74,6 +76,18 @@ export default class ImageGallery {
...
+        if (isBrowserIE) {
+            const fallbackStylesIE = {
+                'background-image': `url(${this.currentImage.mainImageUrl}&ampimbypass=on)`,
+                'background-position': 'center',
+                'background-repeat': 'no-repeat',
+                'background-origin': 'content-box',
+                'background-size': 'contain',
+            };
+
+            this.$mainImageNested.css(fallbackStylesIE);
```
BC-tymurbiedukhin, line 89: Could we use css class instead of js obj?

bc-alexsaiannyi: good point, but we need to change bg-image url depending on chosen image in gallery plus these styles are inline and bad suited for rewriting via css class.

bookernath:
Fixes https://github.com/bigcommerce/cornerstone/issues/1705

chippleh1392:
Thanks for addressing this. Does this fix also update the href of the anchor tag? Also as you're probably aware, the easy zoom functionality does not work on ie11 because of what I assume to be an issue with the third party plugin. Thanks!

bc-alexsaiannyi:
Thanks, Cristofer. This fix works with nested image only. As I see anchor href is being updated with correct URL according to selected image. The difference only is that it uses higher resolution so I decided don’t touch it. I was not sure what else I can affect with it.

PR #1755, opened 2020-07-30
Move clear link outside of Faceted Search navigation button

mattcoy-arcticleaf:
#### What?

The current HTML markup for the link to clear faceted search element's items is invalid as it is located inside a button. This update moves the clear link outside the navigation button, so it no longer violates W3C markup standards (see below for a screenshot of the error message). This update also adds padding so the title of the filter does not overlap with the clear button.

Additionally, this update componentizes the faceted search navigation. Previously, the faceted search navigation code was repeated in the exact same code across 3 different templates.

Lastly, this update cleans up some extraneous whitespace (for consistency with the rest of the codebase) and otherwise questionable code.

The only visible change to the theme is, as stated above, the padding adjustment to prevent the overlap of the faceted search filter title and the clear button.

#### Tickets / Documentation

N/A

W3C Validator error message

### DIFF
--- a/templates/components/faceted-search/facets/hierarchy.html
+++ b/templates/components/faceted-search/facets/hierarchy.html
@@ -1,40 +1,21 @@
 {{#if items}}
 <div class="accordion-block">
-    <div
-        class="accordion-navigation toggleLink {{#unless ../start_collapsed }} is-open {{/unless}}"
-        role="button"
-        data-collapsible="#facetedSearch-content--{{#replace '&' (hyphenate facet) }}{{else}}{{hyphenate facet}}{{/replace}}">
-        <h5 class="accordion-title">
-            {{ title }}
-        </h5>
+    {{> components/faceted-search/faceted-search-navigation}}
 
-        <div class="accordion-navigation-actions">
-            {{#if selected}}
-                <a href="{{ remove_url }}" class="facetedSearch-clearLink" data-faceted-search-facet>{{lang 'search.faceted.clear'}}</a>
-            {{/if}}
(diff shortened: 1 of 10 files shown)


junedkazi:
@bigcommerce/artemis-dt

templates/components/faceted-search/facets/hierarchy.html
```diff
@@ -46,7 +27,7 @@ <h5 class="accordion-title">
         </ul>
 
         {{#if show_more_toggle}}
-            <a href="#facetedSearch-navList--{{#replace '&' (hyphenate facet) }}{{else}}{{hyphenate facet}}{{/replace}}" role="button" class="toggleLink">
+            <a href="#facetedSearch-navList--{{dashcase facet}}" role="button" class="toggleLink">
```
junedkazi, line 30: @lord2800 will this break things for facets which have `&` ?

mattcoy-arcticleaf: @junedkazi , Nope! IMO, dashcase is the correct helper to use here as it strips special characters and converts everything into a nice, usable form.

junedkazi:
@mattcoy-arcticleaf this is good to go. We need to hold off on merge because we are in the process of doing a release. As soon as that is rolled out we will merge this one.

mattcoy-arcticleaf:
@junedkazi no worries. Do you want me to make a changelog entry?

templates/components/faceted-search/facets/hierarchy.html
lord2800: Yeah, I wish I'd found that when we were initially looking into that problem. 😢

PR #1758, opened 2020-08-03
feature(storefront): BCTHEME-74 As a shopper I want to have the add to cart modal immediately have keyboard focus

yurytut1993:
#### What?

@BC-tymurbiedukhin  @bc-alexsaiannyi 

Now when user open a product details modal it has automatically focused on first element (close button). When user tabbing through modal elements focusable will be only modal elements, while user click close button.

#### Tickets / Documentation

[Ticket](https://jira.bigcommerce.com/browse/BCTHEME-74)

[product_details_modal_trap.zip](https://github.com/bigcommerce/cornerstone/files/5015024/product_details_modal_trap.zip)

### DIFF
--- a/assets/js/theme/common/product-details.js
+++ b/assets/js/theme/common/product-details.js
@@ -2,7 +2,7 @@ import utils from '@bigcommerce/stencil-utils';
 import 'foundation-sites/js/foundation/foundation';
 import 'foundation-sites/js/foundation/foundation.reveal';
 import ImageGallery from '../product/image-gallery';
-import modalFactory, { showAlertModal } from '../global/modal';
+import modalFactory, { showAlertModal, modalTypes } from '../global/modal';
 import _ from 'lodash';
 import Wishlist from '../wishlist';
 import { normalizeFormData } from './utils/api';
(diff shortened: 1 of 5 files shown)


assets/js/theme/common/product-details.js
```diff
@@ -390,9 +390,10 @@ export default class ProductDetails {
 
             // Open preview modal and update content
             if (this.previewModal) {
+                console.log('open details modal');
```
BC-tymurbiedukhin, line 393: Log statement

assets/js/theme/global/modal.js
```diff
@@ -219,24 +231,33 @@ export class Modal {
...
-        const $collection = focusableElements[modalType]();
-        $collection.get(0).focus();
+        const { $modalTabbableCollection, modalSelector } = focusableElements[modalType]();
 
-        $('#modal').on('keydown', event => this.onTabbing(event, modalType));
+        this.modalSelector = modalSelector;
+
+        $modalTabbableCollection.get(0).focus();
+
+        $(this.modalSelector).on('keydown', event => this.onTabbing(event, modalType));
```
BC-tymurbiedukhin, line 240: Can we use this.$modal instead of $(this.modalSelector)? If not, it's better to set a init value in class constructor

assets/js/theme/global/modal.js
```diff
@@ -218,25 +230,33 @@ export class Modal {
 
     setupFocusableElements(modalType) {
         this.$preModalFocusedEl = $(document.activeElement);
+        const { $modalTabbableCollection, modalSelector } = focusableElements[modalType]();
 
-        const $collection = focusableElements[modalType]();
-        $collection.get(0).focus();
+        this.modalSelector = modalSelector;
```
BC-tymurbiedukhin, line 235: Thank you for changes, after that we don't need lane 235 as well

yurytut1993: thanks, working on it right now:)

PR #1761, opened 2020-08-04
feat(storefront): BCTHEME-122 Add labels to swatches

BC-tymurbiedukhin:
#### What?

Swatches on PDPs do not have appropriate labels for screen readers to announce.
Each swatch should have an appropriate label.

https://www.w3.org/TR/wai-aria-practices-1.1/#radiobutton

#### Tickets / Documentation

[Ticket](https://jira.bigcommerce.com/browse/BCTHEME-122)

Screenshot 2020-08-04 at 18 17 41


assets/js/theme/common/aria/checkableOptions.js
```diff
@@ -0,0 +1,61 @@
...
+    default: break;
+    }
+};
+
+export default ($container, itemSelector) => {
+    const $itemCollection = $container.find(itemSelector);
+
+    $itemCollection.each((idx, item) => {
+        item.addEventListener('keydown', handleItemKeyDown($itemCollection, idx));
+    });
```
yurytut1993, line 60: Smapll proposal - use delegated event handler instead of setting handler on every element
like this
$itemCollection.on('keydown', itemSelector, $itemCollection)
and after that u can figure out index of element inside handler

BC-tymurbiedukhin: Done

Issue #1372, opened 2018-10-23
JS error when adding to cart in IE: Object doesn't support property or method 'jscomp_symbol_iterator0' OR 'Symbol(Symbol.iterator)'


sacr3dc0w:
### Expected behavior

Add to cart with no JS errors in IE11.

### Actual behavior

Product is added to cart, but object error occurs. I didn't notice this until we installed tracking software noticed our IE11 traffic was throwing this error. These errors occur regardless if a product option set is used.

`Object doesn't support property or method 'jscomp_symbol_iterator0'`

**Stack**
```
TypeError: Object doesn't support property or method 'jscomp_symbol_iterator0'
at filterEmptyFilesFromForm (http://10.0.0.74:3000/stencil/00000000-0000-0000-0000-000000000001/dist/theme-bundle.main.js:1444:89)
at addProductToCart (http://10.0.0.74:3000/stencil/00000000-0000-0000-0000-000000000001/dist/theme-bundle.main.js:1659:9)
at Anonymous function (http://10.0.0.74:3000/stencil/00000000-0000-0000-0000-000000000001/dist/theme-bundle.main.js:1407:13)
at jQuery.event.dispatch (http://10.0.0.74:3000/stencil/00000000-0000-0000-0000-000000000001/dist/theme-bundle.main.js:19892:6)
at elemData.handle (http://10.0.0.74:3000/stencil/00000000-0000-0000-0000-000000000001/dist/theme-bundle.main.js:19700:5)
```

`Object doesn't support property or method 'Symbol(Symbol.iterator)'`

**Stack**
```
TypeError: Object doesn't support property or method 'Symbol(Symbol.iterator)_7.fw2y2bsv01l'
   at filterEmptyFilesFromForm (http://10.0.0.74:3000/stencil/00000000-0000-0000-0000-000000000001/dist/theme-bundle.main.js:1444:89)
   at addProductToCart (http://10.0.0.74:3000/stencil/00000000-0000-0000-0000-000000000001/dist/theme-bundle.main.js:1659:9)
   at Anonymous function (http://10.0.0.74:3000/stencil/00000000-0000-0000-0000-000000000001/dist/theme-bundle.main.js:1407:13)
   at jQuery.event.dispatch (http://10.0.0.74:3000/stencil/00000000-0000-0000-0000-000000000001/dist/theme-bundle.main.js:19892:6)
[...]

Issue #1752, opened 2020-07-29
IE11 fails to add product to cart


sacr3dc0w:
### Expected behavior

Click "Add to Cart" and the preview popup modal displays, letting me know my product has been added.

### Actual behavior

Preview popup opens, stuck spinning while no product is added to cart.

- `Unable to get property 'id' of undefined or null reference theme-bundle.main.js (1356,9)`
  - Line 1356 `_this4.updateCartContent(_this4.previewModal, response.data.cart_item.id);`

### Steps to reproduce behavior

Fresh install cornerstone, boot up IE11, go to PDP and add to cart.

sacr3dc0w:
Here's a video of the add to cart sequence with console.

https://www.creativesafetysupply.com/content/ie11-add-to-cart.mov

sacr3dc0w:
Resolved by #1762

PR #1762, opened 2020-08-04
fix(storefront)BCTHEME-154: fix issue with add product to cart on IE11

bc-alexsaiannyi:
#### What?

@BC-tymurbiedukhin @yurytut1993 

This PR fixes 2 issue that prevent a User to add product to cart on IE11:
- [HTMLInputElement.labels](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/labels) that are used is not supported in IE11 which leads to mentioned in ticket description error
-FormData Object is badly supported in IE11 as well so [polyfill](https://www.npmjs.com/package/formdata-polyfill) has been added to secure things in IE.

#### Tickets / Documentation

- [BCTHEME-154](https://jira.bigcommerce.com/browse/BCTHEME-154)

[BCT154_IE11 Add to Cart Fix](https://github.com/bigcommerce/cornerstone/files/5023338/BCT154_IE11.Add.to.Cart.fix.zip)

### DIFF
--- a/assets/js/polyfills.js
+++ b/assets/js/polyfills.js
@@ -3,6 +3,8 @@ import 'regenerator-runtime/runtime';
 import 'whatwg-fetch';
 import objectFitImages from 'object-fit-images';
 
+require('formdata-polyfill');
+
 document.addEventListener('DOMContentLoaded', () => {
     objectFitImages();
 });
--- a/assets/js/theme/common/product-details.js
+++ b/assets/js/theme/common/product-details.js
@@ -6,6 +6,7 @@ import modalFactory, { showAlertModal } from '../global/modal';
 import _ from 'lodash';
 import Wishlist from '../wishlist';
 import { normalizeFormData } from './utils/api';
+import { isBrowserIE, convertIntoArray } from './utils/ie-helpers';
 
 export default class ProductDetails {
     constructor($scope, context, productAttributesData = {}) {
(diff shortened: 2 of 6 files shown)


sacr3dc0w:
Just cherry-picked your changes to test if this resolves #1372 and #1752 on our end as support wasn't able to replicate the issue and it does successfully add products to cart! Thank you so much. I wish IE11 wasn't our largest source of revenue, but that's the B2B game.

sacr3dc0w:
Thought I jumped the gun, so I deleted my comment, but I was wrong. This does in fact address #1372 and #1752. We run a custom copy of product-details.js (just easier to maintain an original copy for cherry-picking purposes and incorporate changes) and I hadn't merged the changes into our version of product-details. Thanks again for this. It's greatly appreciated.

assets/js/theme/common/product-details.js
```diff
@@ -7,6 +7,8 @@ import _ from 'lodash';
 import Wishlist from '../wishlist';
 import { normalizeFormData } from './utils/api';
 
+require('formdata-polyfill');
```
BC-tymurbiedukhin, line 10: Could we move this import to polyfills.js file?

assets/js/theme/common/product-details.js
```diff
@@ -108,15 +110,21 @@ export default class ProductDetails {
             if (type === 'set-rectangle' || type === 'set-radio' || type === 'swatch' || type === 'input-checkbox' || type === 'product-list') {
                 const checked = value.querySelector(':checked');
                 if (checked) {
+                    const isBrowserIE = navigator.userAgent.includes('Trident');
```
BC-tymurbiedukhin, line 113: I suggest to create a separate util file with IE helpers which can be used across app

assets/js/theme/common/product-details.js
bc-alexsaiannyi: changed

assets/js/theme/common/product-details.js
bc-alexsaiannyi: done

assets/js/polyfills.js
```diff
@@ -3,6 +3,8 @@ import 'regenerator-runtime/runtime';
 import 'whatwg-fetch';
 import objectFitImages from 'object-fit-images';
 
+require('formdata-polyfill');
+
```
BC-tymurbiedukhin, line 7: Any reasons to use require instead of import?

assets/js/theme/common/product-details.js
```diff
@@ -108,15 +109,22 @@ export default class ProductDetails {
             if (type === 'set-rectangle' || type === 'set-radio' || type === 'swatch' || type === 'input-checkbox' || type === 'product-list') {
                 const checked = value.querySelector(':checked');
                 if (checked) {
+                    const getSelectedOptionLabel = () => {
+                        const productVariantslist = convertIntoArray(value.children);
+                        return productVariantslist.reduce((res, field) => {
+                            if (field.dataset.productAttributeValue === checked.value) res = field; // eslint-disable-line no-param-reassign
+                            return res;
+                        });
```
BC-tymurbiedukhin, line 117: I suggest to use .find instead of .reduce 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

assets/js/polyfills.js
bc-alexsaiannyi: it's mentioned in polyfill docs.

BC-tymurbiedukhin: Could you try import instead of require?

assets/js/theme/common/product-details.js
bc-alexsaiannyi: this code is aimed to run only in IE and use a method that requires a polyfill is not 'tasty')

PR #1764, opened 2020-08-05
fix(storefront): BCTHEME-125 Cart remove 'X' should be a button

BC-tymurbiedukhin:
#### What?

On the cart page, the remove item "x" should be a button rather than the current implementation as a link.
Additionally, this button should have an appropriate label.

#### Tickets / Documentation

[Ticket](https://jira.bigcommerce.com/browse/BCTHEME-125)

Screenshot 2020-08-05 at 09 52 49

### DIFF
--- a/templates/components/cart/content.html
+++ b/templates/components/cart/content.html
@@ -122,17 +122,15 @@
                     {{else}}
                         {{> components/common/login-for-pricing}}
                     {{/or}}
-                    {{#if can_modify}}
-                        <a class="cart-remove icon" data-cart-itemid="{{id}}" href="#" data-confirm-delete="{{lang 'cart.confirm_delete'}}">
+                    {{#or can_modify (if type '==' 'GiftCertificate')}}
+                        <button class="cart-remove icon"
+                                data-cart-itemid="{{id}}"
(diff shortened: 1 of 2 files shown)


templates/components/cart/content.html
```diff
@@ -122,17 +122,15 @@ <h4 class="cart-item-name"><a href="{{url}}">{{name}}</a></h4>
...
                     {{/or}}
-                    {{#if can_modify}}
-                        <a class="cart-remove icon" data-cart-itemid="{{id}}" href="#" data-confirm-delete="{{lang 'cart.confirm_delete'}}">
+                    {{#or can_modify (if type '==' 'GiftCertificate')}}
+                        <button class="cart-remove icon"
+                                data-cart-itemid="{{id}}"
+                                data-confirm-delete="{{lang 'cart.confirm_delete'}}"
+                                aria-label="{{lang 'cart.remove_item'}}"
+                        >
                             <svg><use xlink:href="#icon-close"></use></svg>
```
yurytut1993, line 131: this icon is decorative, maybe we should add aria-hidden="true" to hide it from screen readers?

BC-tymurbiedukhin: It's not readable by VoiceOver and ANDI now

PR #1768, opened 2020-08-06
feature(storefront): BCTHEME-142 Unnecessary heading on product cards

yurytut1993:
#### What?

We have product link wrapped in h4 tag which is unnecessary. It can be possible just to remove h4, but I decided to change h4 to span for saving style class which was setted on h4, and now it is setted on span.

Also I should mention that typography styles little bit changed after disabling h4. But we don't have heading already - so we don't need heading styles. I think it is correct - now product name link has similar typography styles with other links in cornerstone. 

#### Tickets / Documentation

[Ticket](https://jira.bigcommerce.com/browse/BCTHEME-142)

main page grid cards
grid_cards

compare page cards
compare_cards

brand page cards
brand_cards

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Draft
 
+Unnecessary heading on product cards. [#1768](https://github.com/bigcommerce/cornerstone/pull/1768)
+
 ## 4.9.0 (07-28-2020)
 - Added correct alt text on image change in product view. [#1747](https://github.com/bigcommerce/cornerstone/pull/1747)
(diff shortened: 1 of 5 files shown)


rowleyaj:
@yurytut1993 will this change have any impact on SEO due to the removal of headers from the page and no relative importance applied to the product names?

I haven't done much research but dropping headers which wrap product titles seems pretty significant

old article (2017) https://www.velsof.com/blog/search-engine-optimization/how-to-organize-the-header-tag-structure-on-an-ecommerce-site

PR #1773, opened 2020-08-07
feat(storefront): BCTHEME-135 Missing tooltip for close button on change modal

BC-tymurbiedukhin:
#### What?

Added tooltip for modal close buttons.

#### Tickets / Documentation

[Jira Ticket](https://jira.bigcommerce.com/browse/BCTHEME-135)

Screenshot 2020-08-07 at 14 27 00


mattcoy-arcticleaf:
@BC-tymurbiedukhin ,

A few notes:

1. The close button could have been made into a component since it's repeated so many times.
2. aria-labels should not be used when you can use text. You should place invisible text inside the button and remove the aria-label.
3. The label is hardcoded as "close" in some places instead of using the variable.
4. Button tags should use type="button" when they are just a button.
5. The text for the tooltip should be "Close" or "Close modal", not "Close button".

BC-tymurbiedukhin:
@mattcoy-arcticleaf Thanks for taking part in our code review process
1. Agree
2. Could you share a doc where I can confirm it?
3. Agree
4. Could you share a doc where I can confirm it?
5. Agree

mattcoy-arcticleaf:
@BC-tymurbiedukhin ,

2. https://www.accessibility-developer-guide.com/knowledge/aria/bad-practices/#approach-using-aria (and the section below it). aria-label is a good step in the right direction, but it isn't the best possible option. aria-label also isn't read by all screen readers, but the invisible text would be.
4. https://stackoverflow.com/questions/41904199/whats-the-point-of-button-type-button It's really just semantics, but if for some reason, the button wound up inside of a form, it would break. There is no reason to really not do it.

bc-alexsaiannyi:
@mattcoy-arcticleaf,  regarding to role button W3C points out on [Example#2](https://www.w3.org/TR/html-aria/#example-2) that button implicitly has role="button".

mattcoy-arcticleaf:
@bc-alexsaiannyi , Yes, but I am not talking about role here. I am talking about type. By default, a button has type="submit". It is good practice to add type="button" to all buttons that are not submitting/resetting a form.

bc-alexsaiannyi:
@mattcoy-arcticleaf, my bad. I misread it. Agree, good point.

BC-tymurbiedukhin:
@mattcoy-arcticleaf Your suggestions were implemented, thank you

PR #1789, opened 2020-08-19
feature(storefront): BCTHEME-130 Carousel links should have discernible text

yurytut1993:
#### What?

Carousel slides had no information about what is it. In carousel bilder we can add alt text to slides to describe them. I used this text to describe slide. And in case when there is no alt text we can add slide number for user to understand what is focused element about

#### Tickets / Documentation

[Ticket](https://jira.bigcommerce.com/browse/BCTHEME-130)

if slide has alt text it will be spelled by screen reader
Screenshot 2020-08-19 at 12 17 47

if no alt text screeen reader will tell that it is slide and its number
Screenshot 2020-08-19 at 12 19 05


templates/components/carousel.html
```diff
@@ -10,7 +10,7 @@
         "accessibility": false
     }'>
     {{#each carousel.slides}}
-    <a href="{{url}}">
+    <a href="{{url}}" aria-label="{{#if this.alt_text.length '!=' 0}}{{this.alt_text}}{{else}}{{lang 'common.carousel_slide_number'}} {{ add @index 1 }}{{/if}}">
```
BC-tymurbiedukhin, line 13: Better to pass slide number into string template as param

BC-tymurbiedukhin (changes requested):

templates/components/carousel.html
yurytut1993: done

BC-tymurbiedukhin (approved):

bc-alexsaiannyi (approved):

PR #1790, opened 2020-08-21
fix(storefront): BCTHEME-161 ARIA attributes on Write Review modal ne…

BC-tymurbiedukhin:
…ed valid values

#### What?

ARIA attributes (i.e. starting with aria-) must contain valid values. These values must be spelled correctly and correspond to values that make sense for a particular attribute to perform the intended accessibility function.

Provide aria-required="true" to the mandatory <input> having id="revfromname_input".

#### Tickets / Documentation

[Jira Ticket](https://jira.bigcommerce.com/browse/BCTHEME-161)


templates/components/products/modals/writeReview.html
```diff
@@ -44,17 +44,17 @@ <h5 class="product-title">{{ product.title }}</h5>
 
                 <!-- Name -->
                 {{#if product.reviews.author}}
-                    {{> components/common/forms/text id="revfromname" name="revfromname" label=(lang 'products.reviews.form_write.name') value=product.reviews.author}}
+                    {{> components/common/forms/text id="revfromname" name="revfromname" required="true" label=(lang 'products.reviews.form_write.name') value=product.reviews.author}}
```
yurytut1993, line 47: Maybe here we also need to use aria-required instead of simple required?

BC-tymurbiedukhin: I'm passing required='true' to partial element which will be converted to aria-required='true'

yurytut1993 (approved):

bc-alexsaiannyi (approved):

PR #1792, opened 2020-08-25
fix(storefront): BCTHEME-144 Button inputs with the same name attribu…

BC-tymurbiedukhin:
…te should be part of a group

#### What?

Radio buttons (and others) in product options are not grouped together.  When this happens, buttons may be misidentified or be grouped with other irrelevant buttons.

Fixed for Radio, Rectangle and Product list product variants. (Was fixed for swatches in https://jira.bigcommerce.com/browse/BCTHEME-122)

See here for more details - https://www.w3.org/TR/2016/WD-wai-aria-practices-1.1-20160317/examples/radio/radio.html
#### Tickets / Documentation

[Jira Ticket](http://example.com)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Draft
+- Button inputs with the same name attribute should be part of a group. [#1792](https://github.com/bigcommerce/cornerstone/pull/1792)
 
 ## 4.9.0 (08-05-2020)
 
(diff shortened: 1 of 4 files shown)


mattcoy-arcticleaf:
@BC-tymurbiedukhin This will cause huge issues if a product has multiple radio options, since you will now have multiple elements with the same ID. Perhaps you could make each one unique by adding the display name to the end?

BC-tymurbiedukhin:
@mattcoy-arcticleaf Thank you for participating, Jira ticket was logged and it will be updated in the nearest future

PR #1805, opened 2020-08-28
Bump stencil-utils to 6.0.0

bookernath:
#### What?

Bump stencil-utils to 6.0.0.

This version of utils gets rid of jquery, and as a result contains significant changes to the event system. It'll need thorough testing, and we might even consider it a major release of Cornerstone as a result (CC @bc-as )


bookernath:
FYI @jairo-bc - it looks like Cornerstone will need some updates to consume the events properly now that they're bound differently.

PR #1820, opened 2020-09-04
fix(storefront): BCTHEME-213 Add Info and Add Coupon forms on Cart Page tabbable when hidden

yurytut1993:
#### What?

In this pr state class of Add Info form from cart page changed from u-hiddenVisually to u-hidden. Because form need to hidden when inactive not only visually, but at all. Add Coupon form was working correctly - it has display: none when hidden

#### Tickets / Documentation

[Ticket](https://jira.bigcommerce.com/browse/BCTHEME-213)

[cart_forms_now_not_tabbable_when_hidden.zip](https://github.com/bigcommerce/cornerstone/files/5174307/cart_forms_now_not_tabbable_when_hidden.zip)


bc-alexsaiannyi (approved):

BC-tymurbiedukhin (approved):

BC-tymurbiedukhin:
The issue has the same code changes with https://github.com/bigcommerce/cornerstone/pull/1844, so only changelog item is visible.

PR #1821, opened 2020-09-04
fix(storefront): STRF-8599 Drop Jquery: Replaced event's current target to the element passed from utils(on hook)

jairo-bc:
…

#### What?

Changed the behaviour of the stencil utils event handlers. Target element is now taken not from event.currentTarget, but passed from the stencil-utils eventemitter.

#### Tickets / Documentation

https://jira.bigcommerce.com/browse/STRF-8599

Example Image

### DIFF
--- a/assets/js/theme/cart.js
+++ b/assets/js/theme/cart.js
@@ -139,8 +139,8 @@ export default class Cart extends PageManager {
             this.bindGiftWrappingForm();
         });
 
-        utils.hooks.on('product-option-change', (event, option) => {
-            const $changedOption = $(option);
+        utils.hooks.on('product-option-change', (event, currentTarget) => {
+            const $changedOption = $(currentTarget);
             const $form = $changedOption.parents('form');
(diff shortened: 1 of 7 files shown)


assets/js/theme/cart.js
```diff
@@ -139,8 +139,8 @@ export default class Cart extends PageManager {
             this.bindGiftWrappingForm();
         });
 
-        utils.hooks.on('product-option-change', (event, option) => {
-            const $changedOption = $(option);
+        utils.hooks.on('product-option-change', (event, currentTarget) => {
```
junedkazi, line 142: Don't we need to change places where we do the `bind` to pass `currentTarget` for all of them ?

jairo-bc: I've already changed it in stencil-utils:
https://github.com/bigcommerce/stencil-utils/blob/master/src/hooks/product.js

junedkazi (approved):

PR #1825, opened 2020-09-08
fix(storefront): BCTHEME-214 Wishlist first option has no bottom outline when in focus

yurytut1993:
#### What?

In this pr focus state of wishlist items was updated. First option now has correct outline at the bottom. Also wishlist main link has correct outline when its focused and option items are opened 

#### Tickets / Documentation

[Ticket](https://jira.bigcommerce.com/browse/BCTHEME-214)

wishlist_option_in_focus


bc-alexsaiannyi (approved):

assets/scss/components/foundation/dropdown/_dropdown.scss
```diff
@@ -10,6 +10,8 @@
 }
 
 .dropdown-menu {
+    z-index: 1;
```
BC-tymurbiedukhin, line 13: Could we use zIndex tool map from _z-index.scss file or do we need exactly 1 and 2 values for the fix?

yurytut1993: We can use $zIndex-low and $zIndex-high from zIndex tool but we do not need so big values. Should we add more variables to the tool? z-index value of .dropdown-menu-button must be higher then .dropdown-menu

BC-tymurbiedukhin: I believe, if we can use current variables and this won't affect any new issue, we should do it)

yurytut1993: done

BC-tymurbiedukhin (approved):

PR #1827, opened 2020-09-08
Reduce lodash usage in compare-products.js and image-gallery.js

sacr3dc0w:
#### What?

Reduce lodash usage (and bundle size) by utilizing the object spread operator to clone imgObj in image-gallery.js when setMainImage() runs.

**BEFORE**
01 before

**AFTER**
02 after

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Draft
+- Reduce lodash usage in compare-products.js and image-gallery.js [#1827](https://github.com/bigcommerce/cornerstone/pull/1827)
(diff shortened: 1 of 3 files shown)


junedkazi (approved):
If we can drop it completely it will be huge. One small step in the right direction.

bookernath (approved):

BC-tymurbiedukhin:
Should themes team take attention on replacing lodash utils with custom helpers?

sacr3dc0w:
I was removing `_.map()` in an internal module we use and figured I'd replace it here as well with 882d01e.

junedkazi (approved):

PR #1828, opened 2020-09-09
fix(storefront): BCTHEME-229 Product carousel with related products - arrows not working

yurytut1993:
#### What?

This PR has fix for unclickable arrow of related products carousel on PDP when related products carousel enabled both with recently viewed carousel.
The problem was because both carousels are made by the same component and they have similar arrow classes. Because of that slick setting on change events incorrectly.
Now in html file of product carousel we define is it recently viewed carousel and setting separate classes for its arrows

#### Tickets / Documentation

[Ticket](https://jira.bigcommerce.com/browse/BCTHEME-229)

related_carousel_arrows


templates/components/products/carousel.html
```diff
@@ -6,8 +6,8 @@
         "slidesToShow": 2,
         "slidesToScroll": 1,
         "slide": ".js-product-slide",
-        "prevArrow": ".js-product-prev-arrow",
-        "nextArrow": ".js-product-next-arrow",
+        "prevArrow": ".js-product-prev-arrow{{#if list '===' "Customers Also Viewed"}}-viewedProducts{{/if}}",
+        "nextArrow": ".js-product-next-arrow{{#if list '===' "Customers Also Viewed"}}-viewedProducts{{/if}}",
```
BC-tymurbiedukhin, line 10: I believe we shouldn't use hardcode like this in reusable component. Could we find any better solution?

bc-alexsaiannyi (approved):

templates/components/products/carousel.html
yurytut1993: we have this text as a param in template
tab_paremeter_for_carousel
any ideas how to save it to variable?

BC-tymurbiedukhin: Could we use more specific selector for arrows with carousel specifying?

yurytut1993: Can you please add more details, can not understand the context. We have two instances of templates/component/product/carousel.html, and we need somehow to specify different arrow classes for them

BC-tymurbiedukhin: We could use something like additional field 'customArrowClass' or data-attribute but not create dependency with title value @yurytut1993

yurytut1993: I added new param customArrowClassAddition to update carousel arrow class and to make in unique

BC-tymurbiedukhin: From my view, better way to implement it - to add additional class, not addition. In case of additional css/js logic we need to rewrite current implementation

yurytut1993: I added separate param for arrow class, when this param is not setted, product carousel use its base class for arrows

BC-tymurbiedukhin (approved):

PR #1830, opened 2020-09-09
fix(storefront): BCTHEME-217 Product link has incorrect voice over focus on Cart Page

yurytut1993:
#### What?

In this PR updated styles for cart link for it to have correct voice over focus outline

#### Tickets / Documentation

[Ticket](https://jira.bigcommerce.com/browse/BCTHEME-217)

product_name_link_in_cart


templates/components/cart/content.html
```diff
@@ -27,7 +27,9 @@
                     {{#if brand.name}}
                         <p class="cart-item-brand">{{brand.name}}</p>
                     {{/if}}
-                    <h2 class="cart-item-name"><a href="{{url}}">{{name}}</a></h2>
+                    <h2 class="cart-item-name">
+                        <a class="cart-item-name-label" href="{{url}}">{{name}}</a>
```
BC-tymurbiedukhin, line 31: As I know according to BEM cart-item-name is block here and label is element, so we should use cart-item-name__label syntax, agree?

yurytut1993: to be honest I think that .cart-item-block is block here:)
cart_table_block

bc-alexsaiannyi (approved):

templates/components/cart/content.html
yurytut1993: done

BC-tymurbiedukhin (approved):

PR #1834, opened 2020-09-10
feat(storefront): STRF-8630 update cornerstone to be compatible with handlebars v4

jairo-bc:
#### What?

Updated cornerstone theme to be compatible with the handlebars v4. Most changes are applied due to changed logic in the calculating `../` and `=` character is now HTML escaped. 
https://github.com/handlebars-lang/handlebars.js/blob/master/release-notes.md#v400---september-1st-2015

#### Tickets / Documentation

https://jira.bigcommerce.com/browse/STRF-8626
https://jira.bigcommerce.com/browse/STRF-8630

#### How it was tested? 
Several times tried to do that using functional tests. Couldn't get a case, when the tests weren't failing. I've got, that latest master(e1270850b1b083d762b339e57096b2c06be53ef3) and these changes have same amount errors(except 1 case on master). Links I'm able to provide in private

### DIFF
--- a/templates/components/account/order-contents.html
+++ b/templates/components/account/order-contents.html
@@ -21,14 +21,18 @@
             </div>
             <figure class="account-product-figure">
                 {{#if type '===' 'giftcertificate'}}
-                    <img src="{{cdn ../../theme_settings.default_image_gift_certificate}}" alt="Gift Certificate" title="Gift Certificate">
+                    <img
+                        src="{{cdn ../theme_settings.default_image_gift_certificate}}"
+                        alt="Gift Certificate"
+                        title="Gift Certificate"
+                    >
                 {{else}}
                     {{> components/common/responsive-img
                         image=image
                         class="account-product-image"
-                        fallback_size=../../theme_settings.productthumb_size
-                        lazyload=../../theme_settings.lazyload_mode
-                        default_image=../../theme_settings.default_image_product
+                        fallback_size=../theme_settings.productthumb_size
+                        lazyload=../theme_settings.lazyload_mode
+                        default_image=../theme_settings.default_image_product
                     }}
                 {{/if}}
                 {{#unless refunded}}
--- a/templates/components/common/quick-search.html
+++ b/templates/components/common/quick-search.html
@@ -2,11 +2,11 @@
     <form class="form" onsubmit="return false" data-url="{{urls.search}}" data-quick-search-form>
         <fieldset class="form-fieldset">
             <div class="form-field">
-                <label class="is-srOnly" for={{name}}>{{lang "search.quick_search.input_label"}}</label>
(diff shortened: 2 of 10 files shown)


templates/components/common/quick-search.html
```diff
@@ -2,11 +2,11 @@
     <form class="form" action="{{urls.search}}">
         <fieldset class="form-fieldset">
             <div class="form-field">
-                <label class="is-srOnly" for={{name}}>{{lang "search.quick_search.input_label"}}</label>
+                <label class="is-srOnly" for="{{name}}>{{lang 'search.quick_search.input_label'}}"</label>
```
BC-tymurbiedukhin, line 5: Could we double check escaping here?

BC-tymurbiedukhin (changes requested):

templates/components/common/quick-search.html
jairo-bc: You are right, closing quote should be after `name`

BC-tymurbiedukhin (approved):

templates/components/account/order-contents.html
```diff
@@ -21,14 +21,18 @@ <h5 class="account-listShipping-title">{{lang 'account.orders.details.ship_to_mu
...
                 {{else}}
                     {{> components/common/responsive-img
                         image=image
                         class="account-product-image"
-                        fallback_size=../../theme_settings.productthumb_size
-                        lazyload=../../theme_settings.lazyload_mode
-                        default_image=../../theme_settings.default_image_product
+                        fallback_size=../theme_settings.productthumb_size
+                        lazyload=../theme_settings.lazyload_mode
+                        default_image=../theme_settings.default_image_product
```
bookernath, line 35: Does `@root` work in v4? It might make for cleaner syntax if we just did `@root.theme_settings.default_image_product` for example.

https://handlebarsjs.com/api-reference/data-variables.html

jairo-bc:
> How have you tested this?

Mostly manually. Took a look at all occurrences of the `../` in the codebase and checked if the are inside #if or other block helpers, that have changed the context behaviour. 

I think after stencil-cli is released with newer version I will be able to run functional tests on integration. Does it take specific version or the latest master? @junedkazi

templates/components/account/order-contents.html
jairo-bc: Didn't check, will take a look.

jairo-bc: Just checked this, for some reason @root is undefined for v3 and v4.

jairo-bc:
@junedkazi @mattolson Added to the description findings from the testings

junedkazi:
@jairo-bc this need a rebase

junedkazi (approved):

PR #1836, opened 2020-09-10
fix(storefront): STRF-8607 Unified browsers list that we support

jairo-bc:
#### What?

Cornerstone had webpack config that overrides babel for the browsers list support. That list was updated to target the browsers, that are in the documentation 

#### Tickets / Documentation

https://support.bigcommerce.com/s/article/Themes-Supported-Browsers

Screenshot 2020-09-10 at 19 30 52

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,0 +1,1 @@
+>2%, last 2 versions, Firefox ESR
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,17 +1,18 @@
-module.exports = {
-    plugins: [
-        '@babel/plugin-syntax-dynamic-import', // add support for dynamic imports (used in app.js)
-        'lodash', // Tree-shake lodash
-    ],
-    presets: [
-        ['@babel/preset-env', {
-            //debug: true,
-            loose: true, // Enable "loose" transformations for any plugins in this preset that allow them
-            modules: 'auto',
-            useBuiltIns: 'usage', // Tree-shake babel-polyfill
-            targets: {
-                node: 'current',
-            },
-        }],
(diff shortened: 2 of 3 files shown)


BC-tymurbiedukhin (approved):
Changes approved, but EOLs needed in .browserslistrc and babel.config.js

.browserslistrc
```diff
@@ -0,0 +1 @@
+last 1 version, >2%
```
junedkazi, line 1: Is there a reason why we are dropping firefox ESR ?

jairo-bc: I've taken the list of supported browsers from the documentation: https://support.bigcommerce.com/s/article/Themes-Supported-Browsers. Firefox ESR stands for _firefox 52_
Screenshot 2020-09-14 at 14 28 13

junedkazi: Changing this to `>2%` is a big change. We should check with @bookernath @bc-as before we make this change. Also we should stick with the last 2 versions.

jairo-bc: last 2 versions > will mean, that we will include IE 10 support and polyfills for it. The doc https://support.bigcommerce.com/s/article/Themes-Supported-Browsers also says, that we are targeting more then 2% browsers

junedkazi: Agreed and I just want a confirmation from product that the doc is right or if it needs updating

bc-as: Lets go with: >2%, last 1 version, Firefox ESR

junedkazi (approved):

junedkazi (approved):

PR #1840, opened 2020-09-15
fix(storefront): BCTHEME-164 Fix image in modal for IE11

bc-alexsaiannyi:
@BC-tymurbiedukhin   @yurytut1993 

#### What?

This PR fixes image absence in modal for IE11. 
The issue relates to using object-fit-images package for resizing images (to fit its container) on legacy browsers like IE11. Globally, we've switched on using polyfill approach by importing all needed polyfills into `/assets/js/polyfills.js`. According to this we use Event Listener with calling **objectFitImages** function in callback if browser doesn't support object-fit/object-position style properties (see templates/components/common/polyfill-script.html). At the same time, function objectFitImages is being invoked second time in Global.js that leads to beaking a product image in modal. 
As a fix the last call has been removed from Global.js alongside object-fit-polyfill.js. objectFitImages has been imported directly in polyfills.js

#### Tickets / Documentation

- [BCTHEME-164](https://jira.bigcommerce.com/browse/BCTHEME-164)

IE11 Product Image in Modal  Fixed


assets/js/theme/global.js
```diff
@@ -36,6 +35,5 @@ export default class Global extends PageManager {
         }
         loadingProgressBar();
         svgInjector();
-        objectFitImages();
```
BC-tymurbiedukhin, line 39: Could we be sure that it wont affect any new issues, I believe in some time it was fix for another problem

bc-alexsaiannyi:
@bookernath, can you please review if removing objectFitImages call from Global won't break anything?

bookernath:
This should be fine. As long as the code is invoked globally (as part of polyfill loading) on the appropriate browsers, no issues.

bookernath (approved):

yurytut1993 (approved):

BC-tymurbiedukhin (approved):

PR #1856, opened 2020-09-29
Upload bundled .zip file on PRs for easy testing

bookernath:
After the bundle test, upload the bundled .zip for easy testing on real stores if needed.

CC @bc-as @bigcommerce/themes-team


junedkazi (approved):
LGTM but we should explore how we can directly attach it to the PR so it is easier to navigate.

bookernath:
Agreed. Let's do it in a future iteration.

PR #1857, opened 2020-09-30
fix(storefront):  BCTHEME-44 Cornerstone quick view

yurytut1993:
#### What?

This PR has fix for variant option in change modal. When there are a lot of options (to scroll appear) selection of options at the bottom cause content block shift to top on Chrome.

#### Tickets / Documentation

[Ticket](https://jira.bigcommerce.com/browse/BCTHEME-44) 

content shift bug
content_jump_bug

content shift fix
content_jump_fix


BC-tymurbiedukhin (approved):

assets/scss/components/foundation/modal/_modal.scss
```diff
@@ -1,6 +1,7 @@
 // =============================================================================
 // MODAL (CSS)
 // =============================================================================
+// Fix for content shifted to top in modal window when bottom variant option selected
```
BC-tymurbiedukhin, line 4: Could we add comment above the fix but not at the top of file?

BC-tymurbiedukhin (approved):

assets/scss/components/foundation/modal/_modal.scss
yurytut1993: Yes we can. I add the comment in such way trying to inherit style of css comments we already have? So should we?
Screenshot 2020-10-01 at 10 40 51

bc-alexsaiannyi (approved):

PR #1860, opened 2020-10-05
feat(storefront): BCTHEME-208 Add to cart notification

bc-alexsaiannyi:
#### What?

This PR adds notifications by screen reader to 'Add to Cart' buttons on PDP & PLPs. In case of PLP Status Message's text is added and can be modified via `en.json` for better UX. 

#### Tickets / Documentation

- [BCTHEME-208](https://jira.bigcommerce.com/browse/BCTHEME-208)

Add to Cart Announcement#1
Add to Cart Announcement#2
Add to Cart Announcement#4


templates/components/products/add-to-cart.html
```diff
@@ -40,8 +40,8 @@
     <p class="alertBox-column alertBox-message"></p>
 </div>
 {{#or customer (if theme_settings.restrict_to_login '!==' true)}}
-    <div class="form-action">
-        <input id="form-action-addToCart" data-wait-message="{{lang 'products.adding_to_cart'}}" class="button button--primary" type="submit"
+    <div class="form-action" role="status" aria-live="polite">
+        <input id="form-action-addToCart"   data-wait-message="{{lang 'products.adding_to_cart'}}" class="button button--primary" type="submit"
```
BC-tymurbiedukhin, line 44: Unnecessary space was added

templates/components/products/card.html
```diff
@@ -53,10 +53,13 @@
                             <a href="{{pre_order_add_to_cart_url}}" data-event-type="product-click" class="button button--small card-figcaption-button">{{lang 'products.pre_order'}}</a>
                         {{/if}}
                         {{#if add_to_cart_url }}
-                            <a href="{{add_to_cart_url}}" data-event-type="product-click" class="button button--small card-figcaption-button">{{lang 'products.add_to_cart'}}</a>
+                            <span class="addCart-wrapper">
```
BC-tymurbiedukhin, line 56: Why do we need additional span with addCart-wrapper css class?

templates/components/products/card.html
```diff
@@ -53,10 +53,13 @@
...
+                                <a href="{{add_to_cart_url}}" data-event-type="product-click" class="button button--small card-figcaption-button addCart">{{lang 'products.add_to_cart'}}</a>
```
BC-tymurbiedukhin, line 57: I believe we should use data attribute instead of css class here

templates/components/products/card.html
```diff
@@ -53,10 +53,13 @@
...
+                                <span class="product-status-message aria-description--hidden">{{lang 'category.add_cart_announcement'}}</span>
+                            </span>
                         {{/if}}
                         {{#if out_of_stock_message }}
-                            <a href="{{url}}" data-event-type="product-click" class="button button--small card-figcaption-button" data-product-id="{{id}}">{{out_of_stock_message}}</a>
+                            <a href="{{url}}" id="test-message" data-event-type="product-click" class="button button--small card-figcaption-button" data-product-id="{{id}}">{{out_of_stock_message}}</a>
```
BC-tymurbiedukhin, line 62: Why do we need to add id for this link?

templates/components/products/card.html
bc-alexsaiannyi: data-event-type="product-click" as other class names are not specifically unique for "Add to Cart" Button. As I see it by using this data attribute we will add extra event listeners that won't be in use.

templates/components/products/card.html
bc-alexsaiannyi: a glitch, will remove

templates/components/products/card.html
BC-tymurbiedukhin: I mean new data-attribute, not event-type

templates/components/products/add-to-cart.html
bc-alexsaiannyi: thanks, removed

templates/components/products/card.html
bc-alexsaiannyi: done

BC-tymurbiedukhin:
LGTM but Changelog item is needed

BC-tymurbiedukhin:
Good job, approved

BC-tymurbiedukhin (approved):

yurytut1993 (approved):

PR #1867, opened 2020-10-07
fix(storefront): BCTHEME-209 Search result not notified by screen reader

yurytut1993:
#### What?

This PR has adding status notifications for search. While user typing in search field live search is working and screen reader announce status of the search results on mobile and on desktop. Also when user getting on search page he has announcement of the search result status too.

#### Tickets / Documentation

[Ticket](https://jira.bigcommerce.com/browse/BCTHEME-209)

mobile search
mobile-search

desktop search
desktop-search

no results found
search-no-results

search page
search-page


assets/js/theme/global/quick-search.js
```diff
@@ -40,6 +40,29 @@ export default function () {
...
+                    'aria-live': 'polite',
+                });
+            } else {
+                const $quickSearchAriaMessage = $quickSearchResultsCurrent.closest('nav').find('[data-search-aria-message]');
+                $quickSearchAriaMessage.addClass('u-hidden');
+
+                const predefinedText = $quickSearchAriaMessage.data('search-aria-message-predefined-text');
+                const itemsFoundCount = $quickSearchResultsCurrent.find('.product').length;
+
+                $quickSearchAriaMessage.html('').html(`
```
bc-alexsaiannyi, line 58: what is the purpose of using html method twice?

yurytut1993: I need to clear previous search result message before adding the new one

bc-alexsaiannyi: Does  previous content won't be [completely replaced](https://api.jquery.com/html/) by the new one from `.html(${itemsFoundCount} ${predefinedText} ${searchQuery})` ?

yurytut1993: yes it is correct html() deleting previous content by default. removed. thanks!

bc-alexsaiannyi (approved):

templates/components/common/navigation.html
```diff
@@ -105,4 +105,5 @@
     <div class="dropdown dropdown--quickSearch" id="quickSearch" aria-hidden="true" tabindex="-1" data-prevent-quick-search-close>
         {{> components/common/quick-search name='nav-quick-search'}}
     </div>
+    {{> components/search/quick-results-aria-message }}
```
BC-tymurbiedukhin, line 108: Could we move this to quick-search file?

templates/components/common/navigation-menu.html
```diff
@@ -210,4 +210,5 @@
             </li>
         {{/if}}
     </ul>
+    {{> components/search/quick-results-aria-message }}
```
BC-tymurbiedukhin, line 213: Could we move this to quick-search file?

assets/js/theme/search.js
```diff
@@ -201,6 +201,10 @@ export default class Search extends CatalogPage {
                 $searchForm.append(input);
             }
         });
+
+        setTimeout(() => {
+            $('[data-search-aria-message]').removeClass('u-hidden');
+        }, 100);
```
BC-tymurbiedukhin, line 207: Why do we need to remove u-hidden class after timeout? Could we just remove u-hidden class from quickSearchResultsAriaMessage element?

templates/components/search/quick-results-aria-message.html
```diff
@@ -0,0 +1,6 @@
+<p role="status"
+   aria-live="polite"
+   class="aria-description--hidden"
+   data-search-aria-message
+   data-search-aria-message-predefined-text="{{{lang 'search.results.quick_count_live'}}}"
+></p>
```
BC-tymurbiedukhin, line 6: If we move this partial to quick-search we don't need to create separate file for it

assets/js/theme/global/quick-search.js
```diff
@@ -40,6 +40,27 @@ export default function () {
...
+                    'aria-live': 'polite',
+                });
+            } else {
+                const $quickSearchAriaMessage = $quickSearchResultsCurrent.closest('nav').find('[data-search-aria-message]');
+                $quickSearchAriaMessage.addClass('u-hidden');
+
+                const predefinedText = $quickSearchAriaMessage.data('search-aria-message-predefined-text');
+                const itemsFoundCount = $quickSearchResultsCurrent.find('.product').length;
+
+                $quickSearchAriaMessage.html(`${itemsFoundCount} ${predefinedText} ${searchQuery}`);
```
BC-tymurbiedukhin, line 58: Could we use text here instead of html?

yurytut1993: changed

templates/components/common/navigation.html
yurytut1993: When input in focus, browser determines its wrapper as group, and reading all its content including search notification text

templates/components/common/navigation-menu.html
yurytut1993: When input in focus, browser determines its wrapper as group, and reading all its content including search notification text

assets/js/theme/search.js
yurytut1993: We need to update this block after every search request for notification to be read by screen reader. . So firstly we need to hide block and then show it. setTimeout will help to reinit notification node and it will be announced. Timeout in debounce, so it will not affect performance

templates/components/search/quick-results-aria-message.html
yurytut1993: When input in focus, browser determines its wrapper as group, and reading all its content including search notification text. So I created this file to import it in two separate navigations

BC-tymurbiedukhin:
A way better, thank you

BC-tymurbiedukhin (approved):

yurytut1993:
> A way better, thank you

As we figured out, tabindex=-1 defined element as a group and because of that screen reader was announcing all content in search input block when input filed is in focus, which is incorrect. So in that ticket we not only added search result notification but also updated search field announcement then its in focus
Thank you for great review

PR #1910, opened 2020-11-20
fix(storefront): BCTHEME-85 “Sort by” dropdown selection not reflected on search results page for News and Information tab

yurytut1993:
#### What?

This PR has fix for search page sorting. It has fix for both product sorting and content sorting.
Two hidden inputs were added to control search query and active tab of the search page (product or content)
Form submits from stencil utils.
sort-by-submit

#### Tickets / Documentation

[BCTHEME-85](https://jira.bigcommerce.com/browse/BCTHEME-85)

bug on products or content request after submitting sort form looks the same, because sending request from content tab will reload page on content tab
bug-after-sortby-product-and-content
fix product
fix-sortby-product
fix content
fix-sortby-content

### DIFF
--- a/assets/js/theme/search.js
+++ b/assets/js/theme/search.js
@@ -132,8 +132,37 @@ export default class Search extends CatalogPage {
         $($tabsCollection.get(nextTabIdx)).focus().trigger('click');
     }
 
+    getUrlParameter(queryParam) {
+        const regex = new RegExp(`[\\?&]${queryParam}=([^&#]*)`);
+        const results = regex.exec(window.location.search);
+        return results === null ? '' : decodeURIComponent(results[1].replace(/\+/g, ' '));
+    }
+
+    setupSortByQuerySearchParam() {
+        const searchQuery = this.getUrlParameter('search_query');
+
+        if (searchQuery.length === 0) return;
+
+        const $baseInput = $('<input/>').attr('type', 'hidden');
+
+        $('[data-sort-by]').each((idx, form) => {
+            const $form = $(form);
+            $form.append(
+                $baseInput.clone().attr({
+                    name: 'search_query',
+                    value: searchQuery,
+                }),
+                $baseInput.clone().attr({
+                    name: 'section',
+                    value: $form.data('sort-by'),
+                }),
+            );
+        });
+    }
+
     onReady() {
         compareProducts(this.context.urls);
+        this.setupSortByQuerySearchParam();
 
         const $searchForm = $('[data-advanced-search-form]');
         const $categoryTreeContainer = $searchForm.find('[data-search-category-tree]');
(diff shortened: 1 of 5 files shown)


assets/js/theme/search.js
```diff
@@ -132,8 +132,43 @@ export default class Search extends CatalogPage {
...
+
+    setupSortByQuerySearchParam() {
+        const searchQuery = this.getUrlParameter('search_query');
+
+        if (searchQuery.length === 0) return;
+
+        const $baseInput = $('<input/>')
+            .attr({
+                type: 'text',
+                class: 'aria-description--hidden',
```
BC-tymurbiedukhin, line 150: I believe it's not semantic to use aria-description--hidden class here

assets/js/theme/search.js
```diff
@@ -132,8 +132,43 @@ export default class Search extends CatalogPage {
         $($tabsCollection.get(nextTabIdx)).focus().trigger('click');
     }
 
+    getUrlParameter(queryParam) {
+        const queryParamKey = queryParam.replace(/[\[]/, '\\[').replace(/[\]]/, '\\]');
```
BC-tymurbiedukhin, line 136: Why do we need to do this if you are passing 'search_query' string?

BC-tymurbiedukhin (commented):
Did you check your fix for mobile?

yurytut1993:
> Did you check your fix for mobile?

yes, it works. Do we have any reasons it can work on desktop but do not work on mobile?

assets/js/theme/search.js
yurytut1993: Maybe we can try use input with type hidden here instead

assets/js/theme/search.js
yurytut1993: This non absolutely necessary in this case. It was made for sterling realisation of such type of function

BC-tymurbiedukhin: For now I believe we should not add it, in case of reusing this function, is should be moved to utils and updated

assets/js/theme/search.js
```diff
@@ -132,8 +132,41 @@ export default class Search extends CatalogPage {
...
+
+    setupSortByQuerySearchParam() {
+        const searchQuery = this.getUrlParameter('search_query');
+
+        if (searchQuery.length === 0) return;
+
+        const $baseInput = $('<input/>')
+            .attr({
+                type: 'hidden',
+            });
```
BC-tymurbiedukhin, line 150: Could we use string literal instead of obj here?

yurytut1993: yeap, now it is possible

BC-tymurbiedukhin (approved):
LGTM

bc-alexsaiannyi (approved):

PR #1911, opened 2020-11-23
fix(storefront): BCTHEME-11 hide product options that are out of stock on cart

bc-alexsaiannyi:
#### What?

This PR fixes an issue with showing product options that are out of stock on Cart Page while hiding them on Product Page. The issue was caused that on Cart Product attributes weren't updated. It was addressed by initiating **ProductDetails Class** on  editing Cart Options via `cartEditOptions` method. It using `updateProductAttributes` method for handle this case.

`productAttributes.optionChange` method from stencil-utils requires productId as its first param. Using itemId instead leads to sending request with undefined and received empty response (the scree's attached). Therefore it was also changed on productId. 

And selector  for `$changedOption`  has been replaced from currentTarget on event.target.Since first one returns us modal instead of option and leads to missing elements afterwards. 

#### Tickets / Documentation

- [BCTHEME-11](https://jira.bigcommerce.com/browse/BCTHEME-11)

#### Screenshots are attached, video is zipped(if appropriate)

[BCTHEME-11.Video4.mov.zip](https://github.com/bigcommerce/cornerstone/files/5581783/BCTHEME-11.Video4.mov.zip)

BCTHEME-11 Control Panel Product
BCTHEME-11  Product options  PDP

BCTHEME-11 Product Options  Cart

### DIFF
--- a/assets/js/theme/cart.js
+++ b/assets/js/theme/cart.js
@@ -1,13 +1,15 @@
 import PageManager from './page-manager';
-import _ from 'lodash';
+import { bind, debounce } from 'lodash';
 import giftCertCheck from './common/gift-certificate-validator';
 import utils from '@bigcommerce/stencil-utils';
 import ShippingEstimator from './cart/shipping-estimator';
 import { defaultModal, modalTypes } from './global/modal';
 import swal from './global/sweet-alert';
+import CartItemDetails from './common/cart-item-details';
 
 export default class Cart extends PageManager {
     onReady() {
+        this.$modal = null;
         this.$cartContent = $('[data-cart-content]');
         this.$cartMessages = $('[data-cart-status]');
         this.$cartTotals = $('[data-cart-totals]');
@@ -125,16 +127,28 @@ export default class Cart extends PageManager {
         });
     }
 
-    cartEditOptions(itemId) {
+    cartEditOptions(itemId, productId) {
+        const context = { productForChangeId: productId, ...this.context };
         const modal = defaultModal();
+
+        if (this.$modal === null) {
+            this.$modal = $('#modal');
+        }
+
         const options = {
             template: 'cart/modals/configure-product',
         };
 
         modal.open();
+        this.$modal.find('.modal-content').addClass('hide-content');
 
         utils.api.productAttributes.configureInCart(itemId, options, (err, response) => {
             modal.updateContent(response.content);
+            const $productOptionsContainer = $('[data-product-attributes-wrapper]', this.$modal);
+            const modalBodyReservedHeight = $productOptionsContainer.outerHeight();
+            $productOptionsContainer.css('height', modalBodyReservedHeight);
+
+            this.productDetails = new CartItemDetails(this.$modal, context);
 
             this.bindGiftWrappingForm();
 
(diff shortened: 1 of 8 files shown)


assets/js/theme/common/product-page-details.js
```diff
@@ -1,51 +1,27 @@
...
-    DATE: 'date',
-    SET_SELECT: 'set-select',
-    SET_RECTANGLE: 'set-rectangle',
-    SET_RADIO: 'set-radio',
-    SWATCH: 'swatch',
-    PRODUCT_LIST: 'product-list',
-};
-
-export default class ProductDetails {
+export default class ProductPageDetails extends ProductDetailsBase {
```
yurytut1993, line 11: I think we should not attach class name to the Page, because generally product details can appear on any page

assets/js/theme/cart.js
```diff
@@ -125,30 +131,35 @@ export default class Cart extends PageManager {
...
         modal.open();
+        $('#modal').find('.modal-content').addClass('hide-content');
 
         utils.api.productAttributes.configureInCart(itemId, options, (err, response) => {
             modal.updateContent(response.content);
+            const $productOptionsContainer = $('[data-product-attributes-wrapper]', $('#modal'));
+            const modalBodyReservedHeight = $productOptionsContainer.outerHeight();
+            $productOptionsContainer.css('height', modalBodyReservedHeight);
+
+            this.productDetails = new CartItemDetails($('#modal'), context);
```
yurytut1993, line 150: in cartEditOptions we are catching modal several times by selector. Can we catch it once and then reuse?

bc-alexsaiannyi: good point. thanks

assets/js/theme/common/product-page-details.js
bc-alexsaiannyi: agree

assets/js/theme/cart.js
```diff
@@ -1,20 +1,27 @@
...
         this.$cartTotals = $('[data-cart-totals]');
+        this.timer = null;
         this.$overlay = $('[data-cart] .loadingOverlay')
             .hide(); // TODO: temporary until roper pulls in his cart components
 
         this.bindEvents();
+
+        $(window).on('beforeunload', () => {
+            clearTimeout(this.timer);
+        });
```
BC-tymurbiedukhin, line 24: Do we need this timer now?

assets/js/theme/common/product-details-base.js
```diff
@@ -0,0 +1,396 @@
...
+    TEXTAREA: 'textarea',
+    DATE: 'date',
+    SET_SELECT: 'set-select',
+    SET_RECTANGLE: 'set-rectangle',
+    SET_RADIO: 'set-radio',
+    SWATCH: 'swatch',
+    PRODUCT_LIST: 'product-list',
+};
+
+export function optionChangeDecorator(condition) {
```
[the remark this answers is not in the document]
bc-alexsaiannyi: done

assets/js/theme/common/cart-item-details.js
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    /**
+     * Hide or mark as unavailable out of stock attributes if enabled
+     * @param  {Object} data Product attribute data
+     */
+    updateProductAttributes(data) {
+        super.updateProductAttributes(data);
+
+        this.$scope.find('.modal-content').removeClass('hide-content');
```
BC-tymurbiedukhin, line 141: Could we add it after line 29 in view like:
`if (this.$scope.find('.modal-content').contains('hide-content')) { 
    this.$scope.find('.modal-content').removeClass('hide-content')
}`

assets/js/theme/cart.js
bc-alexsaiannyi: sorry for that, removed.

assets/js/theme/common/cart-item-details.js
bc-alexsaiannyi: Not exactly. `updateProductAttributes` can be done asynchronously via optionChange,  optionChangeCallback also uses this method inside. If we place your this code below line 29 Product Attributes can be not updated yet and content  will have unwanted attributes. Therefore it's better to keep this line within `updateProductAttributes`.

BC-tymurbiedukhin (approved):
LGTM

yurytut1993 (approved):

PR #1917, opened 2020-11-30
fix(storefront): BCTHEME-313 Parse HTML entities in jsContext

bdav87:
#### What?
This change introduces a utility function to Cornerstone that parses strings with HTML entities safely.

When https://github.com/bigcommerce/paper-handlebars/pull/110 and https://github.com/bigcommerce/paper/pull/215 are deployed, all strings injected into jsContext will have unsafe characters escaped to prevent XSS. This means any string pulled directly from jsContext and surfaced to the storefront user will display HTML entities if it has escaped characters. 

There are values that can be controlled exclusively by the merchant/developer that should be parsed so that they're rendered correctly.

The product review modal is one example; it uses injected strings that are managed from the lang files. Since the strings include escaped single quotes, HTML entities will be displayed to the storefront user unless you parse the string.

#### Tickets / Documentation

- [BCTHEME-313](https://jira.bigcommerce.com/browse/BCTHEME-313)
- [List of escaped characters](https://handlebarsjs.com/api-reference/utilities.html#handlebars-escapeexpression-string)

Before:
image

After:
image

### DIFF
--- a/assets/js/theme/common/utils/safe-string.js
+++ b/assets/js/theme/common/utils/safe-string.js
@@ -1,0 +1,9 @@
+/**
+ * This function parses HTML entities in strings
+ * @param str: String
+ * @returns String
+*/
+export const safeString = (str) => {
+    const d = new DOMParser();
+    return d.parseFromString(str, 'text/html').body.textContent;
+};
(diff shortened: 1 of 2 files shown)


BC-tymurbiedukhin (approved):
LGTM

PR #1923, opened 2020-12-08
fix(storefront): BCTHEME-37 Cornerstone - Image Zoom Does Not Work on Internet Explorer

yurytut1993:
#### What?

This PR has fix for product image zoom on PDP in IE11. The problem was because zoom image height and width were calculated as 0 in IE, so if statement with hiding zoom was truthful. Changing the method of calculating width and height fixed the problem

#### Tickets / Documentation

[BCTHEME-37](https://jira.bigcommerce.com/browse/BCTHEME-37)

in chrome calculations of image sizes was correct
chrome
but in ie11 old method of calculation returned 0
ie

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Draft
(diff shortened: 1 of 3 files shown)


BC-tymurbiedukhin (approved):
LGTM

bc-alexsaiannyi (approved):

PR #1925, opened 2020-12-11
fix(storefront): BCTHEME-16 Cornerstone - loading of thumbnail image delayed on cart page 

yurytut1993:
#### What?

This PR has fix for long loading of product images on Cart Page. The reason of this issue is using wrong path to images from cdn - they are full size instead of being optimized for small size of cart item. Improvement is that for common resolution displays used 100x100 images and for high resolution displays used 200x200 images

#### Tickets / Documentation

[BCTHEME-16](https://jira.bigcommerce.com/browse/BCTHEME-16)

images were taking from the wrong srcset url because of no size attribute it it
retina-images-old

after changing lazyload field sent to responsive-img component, lazysizes library sets sizes attributes to the images and correct url from src set is taken 
Screenshot 2020-12-17 at 16 18 49

### DIFF
--- a/templates/components/cart/content.html
+++ b/templates/components/cart/content.html
@@ -23,7 +23,7 @@
                             image=image
                             class="cart-item-image"
                             fallback_size=../theme_settings.productthumb_size
-                            lazyload='disabled'
+                            lazyload=../theme_settings.lazyload_mode
                             default_image=../theme_settings.default_image_product
                         }}
                     {{/if}}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Draft
+- Cornerstone - loading of thumbnail image delayed on cart page . [#1925](https://github.com/bigcommerce/cornerstone/pull/1925)
 - Parse HTML entities in jsContext. [#1917](https://github.com/bigcommerce/cornerstone/pull/1917)
 - Product images squashed in Category view in AMP. [#1921](https://github.com/bigcommerce/cornerstone/pull/1921)
 - Fixed misaligned tooltip for required product option. [#1915](https://github.com/bigcommerce/cornerstone/pull/1915)


templates/components/cart/content.html
```diff
@@ -19,13 +19,13 @@
                             title="{{lang 'cart.gift_certificate'}}"
                         >
                     {{else}}
-                        {{> components/common/responsive-img
-                            image=image
+                        <img 
+                            src="{{getImageSrcset image 1x=(default fallback_size ../theme_settings.productthumb_size)}}" 
+                            srcset="{{getImageSrcset image 1x=(default fallback_size '200x200')}} 2x"
```
yurytut1993, line 24: I think this value should be get from some new property

templates/components/cart/content.html
```diff
@@ -19,13 +19,17 @@
                             title="{{lang 'cart.gift_certificate'}}"
                         >
                     {{else}}
-                        {{> components/common/responsive-img
-                            image=image
+                        <img
+                            {{#if image}}
+                                src="{{getImageSrcset image 1x=(default fallback_size ../theme_settings.productthumb_size)}}" 
+                                srcset="{{getImageSrcset image 1x=(default fallback_size '200x200')}} 2x"
```
yurytut1993, line 25: I think this value should be get from some new property

templates/components/cart/content.html
```diff
@@ -19,13 +19,17 @@
                             title="{{lang 'cart.gift_certificate'}}"
                         >
                     {{else}}
-                        {{> components/common/responsive-img
```
BC-tymurbiedukhin, line 22: Do we have any ways to not remove usage of responsive-img component?

yurytut1993: To save usage of responsive-img component we need to write additional if statements inside src and arcset properties to use ../theme_settings.productthumb_size as fallback_size. Also we don't need to have so many srcset values here, because the size of cart item image depends on its wrapper size, not the device screen width. Also this responsive-img component is used in a lot of places, that is one more reason not to make its logic more difficult

templates/components/cart/content.html
```diff
@@ -19,13 +19,17 @@
...
-                            image=image
+                        <img
+                            {{#if image}}
+                                src="{{getImageSrcset image 1x=(default fallback_size ../theme_settings.productthumb_size)}}" 
+                                srcset="{{getImageSrcset image 1x=(default fallback_size '200x200')}} 2x"
+                                title="{{image.alt}}"
+                                alt="{{image.alt}}"    
+                            {{else if default_image}}
+                                src="{{cdn default_image}}
+                            {{/if}}
```
bc-alexsaiannyi, line 30: Can you check this please: 
When _responsive-img component_ has been used we explicitly provide _default_image_ path `../theme_settings.default_image_product` there (in old version Line27) since it’s not within _cart.items_ array. If we remove the component then mentioned condition never works.

Am I wrong or we need rewrite/remove condition?

yurytut1993: Yes that is correct - ../theme_settings.default_image_product value should be instead of variable here. I just forgot to push this update - its already done. But thanks for attention:)

templates/components/cart/content.html
```diff
@@ -23,7 +23,7 @@
                             image=image
                             class="cart-item-image"
                             fallback_size=../theme_settings.productthumb_size
-                            lazyload='disabled'
```
yurytut1993, line 26: Hi @bookernath! Need your review. We use disabled value for lazyload and because of that lazyload class is not sertting to responsive-img component. Because of that lazysizes library ignores images from cart list. This lib is responsive to set size attribute to the image to help browser decide which url he should use to load the images. As we have no sizes here, browser chose url from srcset due to viewport (screen size in this case) and loading to big images. Because of loading full size images we have delay on its loading. Any major reasons to use disabled here or we can just use ../theme_settings.lazyload_mode, as we using it for such type of images in similar entities (cart preview list and account order list)?

bc-alexsaiannyi (approved):

templates/components/cart/content.html
bookernath: My reason for disabling here was mostly because the Cart page does not have much other content in terms of images, and the cart images are above-the-fold high priority content. So this was a small optimization for user experience, to avoid delaying the loading of these key images.

However, I am fine with this change if it fixes a problem. 👍

BC-tymurbiedukhin (approved):
LGTM

mattcoy-arcticleaf:
@yurytut1993 @BC-tymurbiedukhin @bc-alexsaiannyi , this change doesn't even work. It will always use the fallback here, regardless of the Lazyload mode set in the theme settings, since `../theme_settings.lazyload_mode` yields nothing in the context you have specified it in here. This is because you have gone a level deeper inside the if statement.

To resolve, replace with `../../theme_settings.lazyload_mode`

/cc @bc-as @vasiliy-seleznev

yurytut1993:
> @yurytut1993 @BC-tymurbiedukhin @bc-alexsaiannyi , this change doesn't even work. It will always use the fallback here, regardless of the Lazyload mode set in the theme settings, since `../theme_settings.lazyload_mode` yields nothing in the context you have specified it in here. This is because you have gone a level deeper inside the if statement.
> 
> To resolve, replace with `../../theme_settings.lazyload_mode`
> 
> /cc @bc-as @vasiliy-seleznev
 
Hello @mattcoy-arcticleaf
In my attempts to reproduce it I found out that current implementation has correct value but your proposal has not.
code
result
Do you have any customisations, maybe in your config.json file?

mattcoy-arcticleaf:
@yurytut1993 , Nope. Pristine Cornerstone 5.3. I also tested directly on a published store by making a copy of Cornerstone and editing the theme files right in the admin. Unless there was a change made to the way Handlebars passes context, the second ../ is required to escape the if statement. 

@bookernath , can you comment on if a change was made to the way Handlebars passes context?

cart-content-code
cart-page

yurytut1993:
@mattcoy-arcticleaf thanks for your help in detection the issue. I have investigated it with @bc-alexsaiannyi and @jairo-bc -   we confirming that it exists but not related to Cornerstone repo. The problem appears on backend side. Cornerstone now using Handlebars v4, but some files mistakenly rendering with Handlebars v3. It will be fixed soon.

PR #1929, opened 2020-12-21
fix(storefront): BCTHEME-345 define a main region on pages

bc-alexsaiannyi:
#### What?

This PR fixes an issue with not defining main region on pages by using HTML5 semantic tag `main`; for supporting this on IE11 `role='main' `has been [explicitly](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main) added  to ensure it is accessible in screen readers like JAWS. 

#### Tickets / Documentation

- [BCTHEME-345](https://jira.bigcommerce.com/browse/BCTHEME-345)

main region on category page
main region on Homepage
main region on PDP

### DIFF
--- a/templates/components/common/body.html
+++ b/templates/components/common/body.html
@@ -1,8 +1,8 @@
-<div class="body" id='main-content' data-currency-code="{{currency_selector.active_currency_code}}">
+<main class="body" id='main-content' role='main' data-currency-code="{{currency_selector.active_currency_code}}">
     {{#block "hero"}} {{/block}}
     <div class="container">
         {{#block "page"}} {{/block}}
     </div>
     {{> components/common/modal/modal}}
     {{> components/common/alert/alert-modal}}
-</div>
+</main>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## Draft
-
+- Add main tag on pages like Homepage, Category, Product etc to define the dominant content. [#1929](https://github.com/bigcommerce/cornerstone/pull/1929)
 ## 5.0.0 (12-14-2020)
 - Parse HTML entities in jsContext. [#1917](https://github.com/bigcommerce/cornerstone/pull/1917)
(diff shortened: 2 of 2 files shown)


yurytut1993 (approved):

templates/components/common/body.html
```diff
@@ -1,8 +1,8 @@
-<div class="body" id='main-content' data-currency-code="{{currency_selector.active_currency_code}}">
+<main class="body" id='main-content' role='main' data-currency-code="{{currency_selector.active_currency_code}}">
```
BC-tymurbiedukhin, line 1: Do we need role='main' if we are using main tag?

bc-alexsaiannyi: it's recommended on MDN to add it for supporting old browsers like IE11

BC-tymurbiedukhin (approved):
LGTM

mattcoy-arcticleaf:
@bc-alexsaiannyi  @BC-tymurbiedukhin @yurytut1993 , won't this cause issues with having 2 main elements on all the pages that already have a main element (page, contact-us, category, etc.)? https://html.spec.whatwg.org/multipage/grouping-content.html#the-main-element 
`A document must not have more than one main element that does not have the hidden attribute specified.`

On an unrelated note, is there a reason single quotes are used instead of double quotes?

/cc @bc-as @vasiliy-seleznev

bc-alexsaiannyi:
@mattcoy-arcticleaf, you’re right. We’ve already created a [ticket](https://jira.bigcommerce.com/browse/BCTHEME-449) to address this issue. 

Regarding quotes - it was a tendency in writing without any specific reason behind and for consistency it will be overwritten in the same ticket. All changes will be delivered in the next release. Thanks for noting this to us.

mattcoy-arcticleaf:
@bc-alexsaiannyi How would I log in to Jira?

bc-alexsaiannyi:
@mattcoy-arcticleaf, you won't. it's for internal use.

PR #1932, opened 2020-12-23
fix(storefront): BCTHEME-346 add announcement for shipping estimator error messages

bc-alexsaiannyi:
#### What?

This PR fixes an issue when shipping estimator error message are not announced by screen reader. aria-live="assertive" has been added to maximize compatibility based on MDN.

#### Tickets / Documentation

- [BCTHEME-346](https://jira.bigcommerce.com/browse/BCTHEME-346)

https://user-images.githubusercontent.com/67792608/102997339-86106b80-452d-11eb-9f4e-32adbe703346.mov

### DIFF
--- a/assets/js/theme/cart/shipping-estimator.js
+++ b/assets/js/theme/cart/shipping-estimator.js
@@ -17,12 +17,22 @@ export default class ShippingEstimator {
     }
 
     initFormValidation() {
+        const shippingEstimatorAlert = $('.shipping-quotes');
+
         this.shippingEstimator = 'form[data-shipping-estimator]';
         this.shippingValidator = nod({
             submit: `${this.shippingEstimator} .shipping-estimate-submit`,
         });
 
         $('.shipping-estimate-submit', this.$element).on('click', event => {
+            // estimator error messages are being injected in html as a result
+            // of user submit; clearing and adding role on submit provides
+            // regular announcement of these error messages
+            if (shippingEstimatorAlert.attr('role')) {
+                shippingEstimatorAlert.removeAttr('role');
+            }
+
+            shippingEstimatorAlert.attr('role', 'alert');
             // When switching between countries, the state/region is dynamic
             // Only perform a check for all fields when country has a value
             // Otherwise areAll('valid') will check country for validity
--- a/templates/components/cart/shipping-estimator.html
+++ b/templates/components/cart/shipping-estimator.html
@@ -67,5 +67,5 @@
             <button class="button button--primary button--small shipping-estimate-submit">{{lang 'cart.shipping_estimator.estimate_shipping'}}</button>
         </dl>
     </form>
-    <div class="shipping-quotes"></div>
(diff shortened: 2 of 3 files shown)


assets/js/theme/cart/shipping-estimator.js
```diff
@@ -17,12 +17,19 @@ export default class ShippingEstimator {
...
+
         this.shippingEstimator = 'form[data-shipping-estimator]';
         this.shippingValidator = nod({
             submit: `${this.shippingEstimator} .shipping-estimate-submit`,
         });
 
         $('.shipping-estimate-submit', this.$element).on('click', event => {
+            if (shippingEstimatorAlert.length && shippingEstimatorAlert.attr('role')) {
+                shippingEstimatorAlert.removeAttr('role');
+            }
```
yurytut1993, line 30: can we avoid double check here? 
Because we are using jQuery and shippingEstimatorAlert its node, we can just check hippingEstimatorAlert.attr('role'). If this attribute will not be present, undefined will be returned value and condition will work

templates/components/cart/shipping-estimator.html
```diff
@@ -67,5 +67,5 @@
             <button class="button button--primary button--small shipping-estimate-submit">{{lang 'cart.shipping_estimator.estimate_shipping'}}</button>
         </dl>
     </form>
-    <div class="shipping-quotes"></div>
+    <div class="shipping-quotes" role="alert" aria-atomic=true aria-live="assertive"></div>
```
yurytut1993, line 70: I believe we should use aria-atomic="true" instead of aria-atomic=true. Attribute value should be in quotes

assets/js/theme/cart/shipping-estimator.js
bc-alexsaiannyi: agree. good point

templates/components/cart/shipping-estimator.html
bc-alexsaiannyi: done

yurytut1993 (approved):
LGTM

assets/js/theme/cart/shipping-estimator.js
```diff
@@ -17,12 +17,19 @@ export default class ShippingEstimator {
...
         this.shippingValidator = nod({
             submit: `${this.shippingEstimator} .shipping-estimate-submit`,
         });
 
         $('.shipping-estimate-submit', this.$element).on('click', event => {
+            if (shippingEstimatorAlert.attr('role')) {
+                shippingEstimatorAlert.removeAttr('role');
+            }
+
+            shippingEstimatorAlert.attr('role', 'alert');
```
BC-tymurbiedukhin, line 32: Could you add some code comments to clarify why it needs to be done

bc-alexsaiannyi: done

BC-tymurbiedukhin (approved):

PR #1933, opened 2020-12-23
fix(storefront): BCTHEME-341 fix placeholder contrast ratio according to AA standard

bc-alexsaiannyi:
#### What?

This PR fixes input placeholder color contrast in compliance with AA standard. It's using ` input-font-color`  from `config.json` for changing color through 3 Theme's variants:
`#666666` for Light, 
`#cccccc` for Bold,
`#4f3f2f` for Warm

#### Tickets / Documentation

- [BCTHEME-341](https://jira.bigcommerce.com/browse/BCTHEME-341)

Placeholder Color
Contrast Ratio Placeholder

### DIFF
--- a/assets/scss/components/citadel/forms/_forms.scss
+++ b/assets/scss/components/citadel/forms/_forms.scss
@@ -137,6 +137,11 @@
 //    appear before the button.
 //
 // -----------------------------------------------------------------------------
+@mixin placeholder {
+    &::-webkit-input-placeholder {@content}
+    &::-moz-placeholder          {@content}
+    &:-ms-input-placeholder      {@content}
+}
 
 .form-label--alternate {
     font-family: fontFamily("headingSans");
@@ -208,6 +213,10 @@
     .form-input {
         @include breakpoint("large") {
             width: auto;
+        }
+
+        @include placeholder {
+            color: $formInput-placeholder-color;
         }
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Draft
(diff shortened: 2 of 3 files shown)


assets/scss/components/citadel/forms/_forms.scss
```diff
@@ -209,6 +209,10 @@
         @include breakpoint("large") {
             width: auto;
         }
+
+        &::placeholder {
+            color: $formInput-placeholder-color;
+        }
```
yurytut1993, line 215: As I can see with [can i use](https://caniuse.com/?search=%3A%3Aplaceholder) ::placeholder won't be working in ie. Should we do something with that?

assets/scss/components/citadel/forms/_forms.scss
```diff
@@ -137,6 +137,12 @@
 //    appear before the button.
 //
 // -----------------------------------------------------------------------------
+@mixin placeholder {
+    &::-webkit-input-placeholder {@content}
+    &::-moz-placeholder          {@content}
+    &:-moz-placeholder          {@content}
```
BC-tymurbiedukhin, line 143: I believe this rule is for Firefox 4-18 and we are not supporting it. Also could you please verify that this solution works in Edge?

assets/scss/components/citadel/forms/_forms.scss
bc-alexsaiannyi: done

assets/scss/components/citadel/forms/_forms.scss
bc-alexsaiannyi: it's true. already removed. edge is supported
edge support

BC-tymurbiedukhin: 👍

BC-tymurbiedukhin (approved):

Issue #1919, opened 2020-12-01
invalid itemCondition in product schema


wwwdepot:
itemCondition is not rendering probably if  **Show condition on storefront** is unchecked

https://github.com/bigcommerce/cornerstone/blob/master/templates/components/products/price.html#L34

### Expected behavior

Product's Condition set to **New** and **Show condition on storefront** is not checked, storefront renders 

```html
<meta itemprop="itemCondition" itemtype="http://schema.org/OfferItemCondition" content="http://schema.org/NewCondition">
```

### Actual behavior

Product's Condition set to **New** and **Show condition on storefront** is not checked, storefront renders 

```html
<!-- notice the "New" is missing -->
<meta itemprop="itemCondition" itemtype="http://schema.org/OfferItemCondition" content="http://schema.org/Condition">
```

### Steps to reproduce behavior
Create a product, with `New condition` & `Show condition on storefront` unchecked

Cornerstone Version: 4.12.1

PR #1936, opened 2020-12-29
fix(storefront): #1919 fix item condition showing if option disabled

fthobe:
#### What?
Hides item condition structured data if item condition not set as active on product page in backend

#### Tickets / Documentation
#1919


BC-tymurbiedukhin:
@fthobe Could you please changelog item with description of the staff that have been done? Also I have a question about condition property, do we have it right now or it will be introduced in future release?

fthobe:
Hey @BC-tymurbiedukhin ,
**Could you please changelog item with description of the staff that have been done?**
 I am not a BC employee, so I don't have that information

**Also I have a question about condition property, do we have it right now or it will be introduced in future release?**
The feature is included since very early. It's needed for proper google shopping campaigns and part of the structured data.

BC-tymurbiedukhin:
> Hey @BC-tymurbiedukhin ,
> **Could you please changelog item with description of the staff that have been done?**
> I am not a BC employee, so I don't have that information
> 
> **Also I have a question about condition property, do we have it right now or it will be introduced in future release?**
> The feature is included since very early. It's needed for proper google shopping campaigns and part of the structured data.

Just add a short description with PR link to draft section of CHANGELOG.md as other list items

PR #1952, opened 2021-01-15
fix(storefront): BCTHEME-361 fix subscription message announcement

bc-alexsaiannyi:
#### What?

This PR fixes an issue when User clicks on subscribe button he has been redirected to a new page with Subscription status that is not announced by Screen Reader. The solution is to add role="status" and aria-live="assertive" on element with message and move focus to that element after loading the page.

#### Tickets / Documentation

- [BCTHEME-361](https://jira.bigcommerce.com/browse/BCTHEME-361)

https://user-images.githubusercontent.com/67792608/104756907-13c62c00-5765-11eb-97b8-1822953434d3.mov


BC-tymurbiedukhin (commented):
could we access this page manually?

bc-alexsaiannyi:
> could we access this page manually?

No, I didn't find the way.

BC-tymurbiedukhin (approved):
LGTM

yurytut1993 (approved):

mattcoy-arcticleaf:
@bc-alexsaiannyi ,

Wouldn't we want to avoid using script tags in our HTML? How about we just make a global JS file that focuses the first element with a certain data attribute, and then we can just add that data attribute to any item that needs to be focused on page load?

PR #1961, opened 2021-01-20
Update _navUser.scss

jake1093:
top: auto !important; is overwriting position set inline by javascript.

#### What?

Adding a top: auto value to minicart dropdown overrides inline styling set by javascript function.


BC-tymurbiedukhin:
Could you please attach before and after screenshots why this change was made?

PR #1964, opened 2021-01-21
fix(storefront): BCTHEME-207 fix focus on sort by

bc-alexsaiannyi:
#### What?

This PR addresses focus issue on SortBy element after reloading page. It can be seen as a solution before implementing partly page update mechanism. At the moment, new method `arrangeFocusOnSortBy` has been added to CatalogPage Class and can be used on pages where sortBy select is presented (such as category or search). CatalogPage constructor helps to set up event listener that tracks if sortBy is active and keeps it on LocalStorage  in that case.

#### Tickets / Documentation
- [BCTHEME-207](https://jira.bigcommerce.com/browse/BCTHEME-207)

https://user-images.githubusercontent.com/67792608/105361889-dbaa6780-5c02-11eb-8d58-c2bdc5bdf070.mov

### DIFF
--- a/assets/js/theme/catalog.js
+++ b/assets/js/theme/catalog.js
@@ -3,6 +3,25 @@ import urlUtils from './common/utils/url-utils';
 import Url from 'url';
 
 export default class CatalogPage extends PageManager {
+    constructor(context) {
+        super(context);
+
+        window.addEventListener('beforeunload', () => {
+            if (document.activeElement.id === 'sort') {
+                window.localStorage.setItem('sortByStatus', 'selected');
+            }
+        });
+    }
+
+    arrangeFocusOnSortBy() {
+        const $sortBySelector = $('[data-sort-by="product"] #sort');
+
+        if (window.localStorage.getItem('sortByStatus')) {
+            $sortBySelector.focus();
+            window.localStorage.removeItem('sortByStatus');
+        }
+    }
+
     onSortBySubmit(event, currentTarget) {
         const url = Url.parse(window.location.href, true);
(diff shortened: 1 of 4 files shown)


assets/js/theme/category.js
```diff
@@ -8,16 +8,33 @@ export default class Category extends CatalogPage {
...
+
+    arrangeFocusOnElement($element) {
+        if (this.isSortBySelected) {
+            $element.focus();
+            window.localStorage.removeItem('sortByStatus');
+        }
     }
 
     onReady() {
+        const sortBySelector = $('[data-sort-by="product"] #sort');
```
BC-tymurbiedukhin, line 26: Could we use only #sort id here?

yurytut1993 (approved):

assets/js/theme/catalog.js
```diff
@@ -3,6 +3,23 @@ import urlUtils from './common/utils/url-utils';
...
+        super(context);
+
+        window.addEventListener('beforeunload', () => {
+            if (document.activeElement.id === 'sort') {
+                window.localStorage.setItem('sortByStatus', 'selected');
+            }
+        });
+    }
+
+    arrangeFocusOnSortBy($elementSelector) {
```
BC-tymurbiedukhin, line 16: I believe we could skip passing selector here

bc-alexsaiannyi: I've added it in case the sortBy Selector will be change later for some reason. Removed it.

BC-tymurbiedukhin (approved):
LGTM

PR #1977, opened 2021-02-01
feat(storefront): BCTHEME-391 Move focus on filter items Modal after show more button was clicked and accessibility refactoring

yurytut1993:
#### What?

This PR adds focus trap on show more modal with filter options

#### Tickets / Documentation

[BCTHEME-391
](https://jira.bigcommerce.com/browse/BCTHEME-391)

This flow can be present on PLP

https://user-images.githubusercontent.com/66325265/106426431-42831880-646e-11eb-8d58-ab0ad35df03b.mov

search page

https://user-images.githubusercontent.com/66325265/106426472-5169cb00-646e-11eb-9d3c-e79a94d60749.mov

and brands

https://user-images.githubusercontent.com/66325265/106426487-5af33300-646e-11eb-8221-b6dfdd0de070.mov

### DIFF
--- a/assets/js/theme/common/faceted-search.js
+++ b/assets/js/theme/common/faceted-search.js
@@ -2,10 +2,13 @@ import { hooks, api } from '@bigcommerce/stencil-utils';
 import _ from 'lodash';
 import Url from 'url';
 import urlUtils from './utils/url-utils';
-import modalFactory from '../global/modal';
+import modalFactory, { modalTypes, ModalEvents } from '../global/modal';
 import collapsibleFactory from './collapsible';
 import { Validators } from './utils/form-utils';
 import nod from './nod';
+
+const { SHOW_MORE_OPTIONS } = modalTypes;
+const { opened } = ModalEvents;
 
 const defaultOptions = {
     accordionToggleSelector: '#facetedSearch .accordion-navigation, #facetedSearch .facetedSearch-toggle',
@@ -55,6 +58,15 @@ class FacetedSearch {
         this.options = _.extend({}, defaultOptions, options);
         this.collapsedFacets = [];
         this.collapsedFacetItems = [];
+
+        if (this.options.modal) {
+            this.options.modal.$modal.on(opened, event => {
+                const $filterItems = $(event.target).find('#facetedSearch-filterItems');
+                if ($filterItems.length) {
+                    this.options.modal.setupFocusableElements(SHOW_MORE_OPTIONS);
+                }
+            });
+        }
 
         // Init collapsibles
         collapsibleFactory();
--- a/assets/js/theme/global/modal.js
+++ b/assets/js/theme/global/modal.js
@@ -22,24 +22,26 @@ export const modalTypes = {
     PRODUCT_DETAILS: 'forProductDetails',
     CART_CHANGE_PRODUCT: 'forCartChangeProduct',
     WRITE_REVIEW: 'forWriteReview',
+    SHOW_MORE_OPTIONS: 'forShowMore',
 };
 
+const findRootModalTabbableElements = () => (
+    $('#modal.open')
+        .find(allTabbableElementsSelector)
+        .not('#modal-review-form *')
+        .not('#previewModal *')
+);
+
 const focusableElements = {
-    [modalTypes.QUICK_VIEW]: () => (
-        $('#modal')
-            .find(allTabbableElementsSelector)
-            .not('#modal-review-form *')
-            .not('#previewModal *')
-    ),
+    [modalTypes.QUICK_VIEW]: findRootModalTabbableElements,
     [modalTypes.PRODUCT_DETAILS]: () => (
-        $('#previewModal').find(allTabbableElementsSelector)
+        $('#previewModal.open').find(allTabbableElementsSelector)
     ),
-    [modalTypes.CART_CHANGE_PRODUCT]: () => (
-        $('#modal').find(allTabbableElementsSelector)
-    ),
+    [modalTypes.CART_CHANGE_PRODUCT]: findRootModalTabbableElements,
     [modalTypes.WRITE_REVIEW]: () => (
-        $('#modal-review-form').find(allTabbableElementsSelector)
+        $('#modal-review-form.open').find(allTabbableElementsSelector)
     ),
+    [modalTypes.SHOW_MORE_OPTIONS]: findRootModalTabbableElements,
 };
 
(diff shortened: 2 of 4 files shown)


assets/js/theme/global/modal.js
```diff
@@ -22,24 +22,28 @@ export const modalTypes = {
...
     WRITE_REVIEW: 'forWriteReview',
+    SHOW_MORE_OPTIONS: 'forShowMore',
 };
 
+const findRootModal = () => (
+    $('#modal.open')
+        .find(allTabbableElementsSelector)
+        .not('#modal-review-form *')
+        .not('#previewModal *')
+);
```
BC-tymurbiedukhin, line 33: Could we make this name more descriptive?

assets/js/theme/global/modal.js
```diff
@@ -138,6 +147,7 @@ export class Modal {
         this.onModalClose = this.onModalClose.bind(this);
         this.onModalClosed = this.onModalClosed.bind(this);
 
+        this.unbindEvents();
```
BC-tymurbiedukhin, line 150: Why do we need this unbind?

assets/js/theme/common/faceted-search.js
```diff
@@ -56,6 +59,15 @@ class FacetedSearch {
...
         this.collapsedFacetItems = [];
 
+        this.onMoreOptionsOpened = (event) => {
+            const $filterItems = $(event.target).find('#facetedSearch-filterItems');
+            if ($filterItems.length) {
+                this.options.modal.setupFocusableElements(SHOW_MORE_OPTIONS);
+            }
+        };
+
+        $('body').on(opened, this.onMoreOptionsOpened);
```
BC-tymurbiedukhin, line 69: Do we need to bind onMoreOptionsOpened on every modal which is open in body at pages with faceted search?

assets/js/theme/global/modal.js
yurytut1993: When page loaded, modal class create instances for all modals and binding events on constructor. After that this instances cashing and using until page will be reloaded. But I am thinking about if in some cases instances can be recreated and set up event listeners again and double them. I did not find scenario when it can happen. But decided to turn off old listeners before setting new, just in case

assets/js/theme/common/faceted-search.js
yurytut1993: Do you mean to replace onMoreOptionsOpened just on modal where options can appear?

assets/js/theme/global/modal.js
yurytut1993: findRootModalTabbableElements?

assets/js/theme/common/faceted-search.js
BC-tymurbiedukhin: Could we do it?

assets/js/theme/global/modal.js
BC-tymurbiedukhin: I believe we could skip this for now, modals should not be initialised separately

assets/js/theme/common/faceted-search.js
yurytut1993: yes -its done

assets/js/theme/global/modal.js
yurytut1993: To be honest I also think that unbind should not be useful - removed

assets/js/theme/global/modal.js
```diff
@@ -181,13 +188,6 @@ export class Modal {
         this.$modal.on(ModalEvents.opened, this.onModalOpened);
     }
 
-    unbindEvents() {
-        this.$modal.off(ModalEvents.close, this.onModalClose);
-        this.$modal.off(ModalEvents.closed, this.onModalClosed);
-        this.$modal.off(ModalEvents.open, this.onModalOpen);
-        this.$modal.off(ModalEvents.opened, this.onModalOpened);
-    }
```
BC-tymurbiedukhin, line 189: Why do we need to removed method which is in use?

yurytut1993: Unbind is not in use now. It was using firstly in onmodalclosed function then I replaced it to constructor. After we decided to remove it from constructor - it is not in use anymore

BC-tymurbiedukhin: Why did we move it from onmodalclosed?

yurytut1993: Because after we closed modal once, we still need this events to exist. For them to work when modal will be opening and closing next times

BC-tymurbiedukhin: Could we put event binding on modal opened and unbinding on modal closed?

assets/js/theme/common/faceted-search.js
```diff
@@ -56,6 +59,17 @@ class FacetedSearch {
         this.collapsedFacets = [];
         this.collapsedFacetItems = [];
 
+        this.onMoreOptionsOpened = (event) => {
+            const $filterItems = $(event.target).find('#facetedSearch-filterItems');
+            if ($filterItems.length) {
+                this.options.modal.setupFocusableElements(SHOW_MORE_OPTIONS);
+            }
+        };
```
BC-tymurbiedukhin, line 67: Do we really need this method?

assets/js/theme/global/modal.js
```diff
@@ -22,24 +22,26 @@ export const modalTypes = {
...
-        $('#modal').find(allTabbableElementsSelector)
+        $('#previewModal.open').find(allTabbableElementsSelector)
     ),
+    [modalTypes.CART_CHANGE_PRODUCT]: findRootModalTabbableElements,
     [modalTypes.WRITE_REVIEW]: () => (
-        $('#modal-review-form').find(allTabbableElementsSelector)
+        $('#modal-review-form.open').find(allTabbableElementsSelector)
     ),
+    [modalTypes.SHOW_MORE_OPTIONS]: findRootModalTabbableElements,
 };
```
BC-tymurbiedukhin, line 45: Could we unify approach? For me it looks like adding custom attributes for show more and quick view modals looks good

assets/js/theme/common/faceted-search.js
yurytut1993: updated us just event method

assets/js/theme/global/modal.js
yurytut1993: discussed in dm - focusableElements object with functions stays at it was

PR #1978, opened 2021-02-02
Replaced scss-lint with stylelint and dropped out Gem files

BC-tymurbiedukhin:
#### What?

Scss lint is no longer supported https://github.com/sds/scss-lint
Also it is the only reason why as part of CI/CD test perspective we need to install ruby (See the Gemfile and the Gem lock file)
So we should switch over to something better.

#### Tickets / Documentation

[Jira ticket](https://jira.bigcommerce.com/browse/BCTHEME-152)


.scss-lint.yml
```diff
@@ -1,60 +0,0 @@
...
-    style: new_line # or 'same_line'
-
-  EmptyLineBetweenBlocks:
-    enabled: true
-    ignore_single_line_blocks: false
-
-  Indentation:
-    enabled: true
-    character: space # or 'tab'
-    width: 4
```
bc-alexsaiannyi, line 23: as I see SCSS guidelines uses 2 spaces for Indentation and this leads to lots of changes in files. Does it worth to rewrite it now?

PR #1987, opened 2021-02-16
fix(storefront): BCTHEME-401 Category pages are creating alt attribute within the span tag

yurytut1993:
#### What?

Due to [w3c](https://www.w3.org/TR/wai-aria-1.1/#img) alt text is acceptable for role image, but if it affects validation we can use second approach - aria-label attribute for element with role image. It has the same result for screen reader and successfully pass validation. Due to [mdn](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Role_Img) alt also acceptable, but all examples using aria-label, so it should be the best solution.
 

#### Tickets / Documentation

 [BCTHEME-401](https://jira.bigcommerce.com/browse/BCTHEME-401)

amp-validation

amp-garden-item-rating

garden-item-rating

jar-rating

### DIFF
--- a/templates/components/amp/products/ratings.html
+++ b/templates/components/amp/products/ratings.html
@@ -1,6 +1,5 @@
 <span role="img"
-      alt="{{lang 'products.reviews.rating_aria_label' current_rating=rating max_rating=5}}"
-      title="{{lang 'products.reviews.rating_aria_label' current_rating=rating max_rating=5}}"
+      aria-label="{{lang 'products.reviews.rating_aria_label' current_rating=rating max_rating=5}}"
 >
     {{#for 1 5}}
         {{#if ../this.rating '>=' $index}}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Draft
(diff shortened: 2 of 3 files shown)


bc-alexsaiannyi (approved):

templates/components/amp/products/ratings.html
```diff
@@ -1,6 +1,5 @@
 <span role="img"
-      alt="{{lang 'products.reviews.rating_aria_label' current_rating=rating max_rating=5}}"
-      title="{{lang 'products.reviews.rating_aria_label' current_rating=rating max_rating=5}}"
+      aria-label="{{lang 'products.reviews.rating_aria_label' current_rating=rating max_rating=5}}"
```
BC-tymurbiedukhin, line 2: Could you please explain why using of title is not enough here?

yurytut1993: I used approch from [mdn](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Role_Img) . As I understood this is the best pattern for such cases

BC-tymurbiedukhin (approved):
LGTM

PR #1997, opened 2021-02-26
Remove AddThis for social sharing, replace with provider sharing links

bookernath:
#### What?

Remove 3rd-party dependency on [AddThis](https://www.addthis.com/) which provides social sharing functionality.

Instead, links have been replaced with the native URL formats for each service.

Google+ sharing functionality has been removed, as Google+ no longer exists.

#### Tickets / Documentation

- [STRF-5798](https://jira.bigcommerce.com/browse/STRF-5798)

FYI @bc-as @bigcommerce/themes-team

### DIFF
--- a/templates/components/common/share.html
+++ b/templates/components/common/share.html
@@ -1,57 +1,90 @@
 {{#if settings.add_this.buttons}}
-    <div class="addthis_toolbox addthis_32x32_style" addthis:url="{{url}}"
-         addthis:title="{{title}}">
+    {{assignVar 'encodedUrl' (encodeURI url)}}
+    {{assignVar 'encodedTitle' (encodeURI head.title)}}
+    <div>
         <ul class="socialLinks">
             {{#each settings.add_this.buttons}}
                 <li class="socialLinks-item socialLinks-item--{{service}}">
-                    <a class="addthis_button_{{service}} socialLinks__link icon icon--{{service}}"
+                    {{#if service '===' 'facebook'}}
+                    <a class="socialLinks__link icon icon--{{service}}"
                        title="{{{capitalize service}}}"
-                       href="#"
+                       href="https://facebook.com/sharer/sharer.php?u={{getVar 'encodedUrl'}}"
+                       target="_blank"
+                       rel="noopener"
                        {{{annotation}}}
                     >
                         <span class="aria-description--hidden">{{{capitalize service}}}</span>
-                        {{#if service '===' 'facebook'}}
-                            <svg>
-                                <use xlink:href="#icon-facebook"/>
-                            </svg>
-                        {{else if service '===' 'email'}}
-                            <svg>
-                                <use xlink:href="#icon-envelope"/>
-                            </svg>
-                        {{else if service '===' 'print'}}
-                            <svg>
-                                <use xlink:href="#icon-print"/>
-                            </svg>
-                        {{else if service '===' 'twitter'}}
-                            <svg>
-                                <use xlink:href="#icon-twitter"/>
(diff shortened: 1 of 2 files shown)


templates/components/products/product-view.html
```diff
@@ -248,9 +248,7 @@ <h2 class="productView-brand"{{#if schema}} itemprop="brand" itemscope itemtype=
                 {{> components/common/wishlist-dropdown}}
             {{/if}}
         </div>
-        {{#unless is_ajax}}
```
bookernath, line 251: No longer needed to hide from AJAX (quick view) as we no longer have a JS dependency, and the sharing URL is passed explicitly from the product object.

templates/components/common/share.html
```diff
@@ -1,57 +1,90 @@
...
-                            </svg>
-                        {{else if service '===' 'twitter'}}
-                            <svg>
-                                <use xlink:href="#icon-twitter"/>
-                            </svg>
-                        {{else if service '===' 'linkedin'}}
-                            <svg>
-                                <use xlink:href="#icon-linkedin"/>
-                            </svg>
-                        {{else if service '===' 'google'}}
```
bookernath, line 33: Completely removed as Google+ is no longer a thing.

BC-tymurbiedukhin (approved):
Nice to see this PR, approved

templates/components/common/share.html
```diff
@@ -1,57 +1,90 @@
...
+                    >
+                        <span class="aria-description--hidden">{{{capitalize service}}}</span>
+                        <svg>
+                            <use xlink:href="#icon-linkedin"/>
+                        </svg>
+                    </a>
+                    {{else if service '===' 'pinterest'}}
+                    <a class="socialLinks__link icon icon--{{service}}"
+                       title="{{{capitalize service}}}"
+                       href="https://pinterest.com/pin/create/button/?url={{getVar 'encodedUrl'}}&amp;media={{getVar 'encodedUrl'}}&amp;description={{getVar 'encodedTitle'}}"
```
bc-as, line 74: @bookernath We should probably remove the media query here and let Pinterest pick up the image automatically, otherwise we're sending the page URL over twice which might cause an issue.

bookernath: Okay, got it. So we should only include `url`? Or `url` and `description`?

templates/components/products/product-view.html
```diff
@@ -248,9 +248,7 @@ <h2 class="productView-brand"{{#if schema}} itemprop="brand" itemscope itemtype=
...
-            {{> components/common/share}}
-        {{/unless}}
+        {{> components/common/share url=product.url}}
```
bookernath, line 251: Passing in `url` explicitly, which makes this component more re-usable in other contexts if desired.

templates/components/common/share.html
bc-as: We should use both `url` and `description` here.

bookernath: @bc-as ♻️

yurytut1993:
We are using [focus trap library](https://www.npmjs.com/package/focus-trap) and it requires iframe not to be the last tabbable item in the modal. In case when product has no description facebook button can be the last and focus trap will avoid to focus it using keyboard. Can we replace facebook button iframe to the top or just hide it in modal?

https://user-images.githubusercontent.com/66325265/109482943-4adb6d00-7a87-11eb-85e7-e63222a1f653.mov

yurytut1993 (approved):

bc-alexsaiannyi (approved):

PR #1998, opened 2021-02-26
fix(storefront): BCTHEME-409 Update focus trap in Modal

yurytut1993:
#### What?

This PR has updates for focus trap in case when iframe appears in modal. In some cases focus can be lost. It is fixed and now focus can be returned from iframe

#### Tickets / Documentation

[BCTHEME-409](https://jira.bigcommerce.com/browse/BCTHEME-409)

https://user-images.githubusercontent.com/66325265/109284620-bb388300-7828-11eb-994c-c80a5c8de376.mov


assets/js/theme/global/modal.js
```diff
@@ -1,49 +1,17 @@
-import 'jquery.tabbable';
 import foundation from './foundation';
+import * as focusTrap from 'focus-trap';
```
BC-tymurbiedukhin, line 2: Does import like { createFocusTrap } will make any profit?

BC-tymurbiedukhin (approved):
LGTM, just keep package-lock.json file updated as well

assets/js/theme/global/modal.js
yurytut1993: Your proposal is working, but if you not insist - I prefer using approach from doc
Screenshot 2021-03-01 at 10 13 59

bc-alexsaiannyi (approved):

PR #2002, opened 2021-03-02
fix(search): ES-2031 add error message for when star search is not av…

bc-nataliya:
## What?
When star search is unavailable for any reasons make sure we have a nice fall back with an appropriate message. 

## Why?
Right now bcapp just throws 500 in storefront that is not a nice way to take care of failed request

## Testing / Proof
For testing run bundle stencil to create new theme with cornerstone changes. Locally stop star search to make requests failed. Enable faceted search experiment. Access Sategory and Brands page and make sure you see message like: "There was a problem retrieving the products on this page. Please try again at a later time"

## How can this change be undone in case of failure?
Turn off star search faceted search experiment

ping @bigcommerce/artemis-dt

### DIFF
--- a/templates/components/category/product-listing.html
+++ b/templates/components/category/product-listing.html
@@ -19,11 +19,15 @@
 
     {{> components/common/paginator pagination.category}}
 {{else}}
-    <p data-no-products-notification
-       role="alert"
-       aria-live="assertive"
-       tabindex="-1"
-    >
-        {{lang 'categories.no_products'}}
-    </p>
+    {{#if category.search_error}}
+        {{lang 'categories.search_error'}}
+    {{else}}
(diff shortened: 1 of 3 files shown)


templates/components/category/product-listing.html
```diff
@@ -19,11 +19,15 @@
...
 {{else}}
-    <p data-no-products-notification
-       role="alert"
-       aria-live="assertive"
-       tabindex="-1"
-    >
-        {{lang 'categories.no_products'}}
-    </p>
+    {{#if category.search_error}}
+        {{lang 'categories.search_error'}}
```
bc-alexsaiannyi, line 23: do we need to wrap it in the tag?

PR #2017, opened 2021-03-19
fix(storefront): IE11 - Clicking on Search Does Not Display Search Bar

yurytut1993:
#### What?

After this PR merged Search input will appear after clicking on top menu Search button 

#### Tickets / Documentation

[BCTHEME-403](https://jira.bigcommerce.com/browse/BCTHEME-403)

search-bar-ie

### DIFF
--- a/assets/scss/components/stencil/navUser/_navUser.scss
+++ b/assets/scss/components/stencil/navUser/_navUser.scss
@@ -291,7 +291,7 @@
     }
 
     &.is-open {
-        display: initial;
+        display: block;
         // scss-lint:disable ImportantRule
         left: 0 !important; // 1
         outline: none;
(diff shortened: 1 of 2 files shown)


assets/scss/components/stencil/navUser/_navUser.scss
```diff
@@ -291,7 +291,7 @@
     }
 
     &.is-open {
-        display: initial;
+        display: block;
```
BC-tymurbiedukhin, line 294: Does it affect other browser behaviours?

yurytut1993: I believe no, because block is init value for div display prop

BC-tymurbiedukhin (approved):
LGTM

PR #2028, opened 2021-04-02
fix(storefront): BCTHEME-447 extends keyboard support for radio buttons

bc-alexsaiannyi:
#### What?

This PR adds extended keyboard support for radio buttons. Such behaviour is based on [W3C](https://www.w3.org/TR/wai-aria-practices/examples/radio/radio-1/radio-1.html) & includes:

- Pressing on right/down arrow keys moves focus and checks the next radio button in the group
- Pressing on left/up arrow keys | moves focus and checks the previous radio button in the group

#### Tickets / Documentation
- [BCTHEME-447](https://jira.bigcommerce.com/browse/BCTHEME-447)

Here you can find a video before and after changes. Moving cross options is made with arrow keys.

https://user-images.githubusercontent.com/67792608/113392996-af4b5e00-939e-11eb-8006-e4ca01ad9eec.mov

https://user-images.githubusercontent.com/67792608/113393127-e1f55680-939e-11eb-9162-1b176a17eff6.mov

### DIFF
--- a/assets/js/theme/common/aria/radioOptions.js
+++ b/assets/js/theme/common/aria/radioOptions.js
@@ -9,6 +9,7 @@ const setCheckedRadioItem = (itemCollection, itemIdx) => {
         }
 
         $item.attr('aria-checked', true).prop('checked', true).focus();
+        $item.trigger('change');
     });
 };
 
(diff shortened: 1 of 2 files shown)


BC-tymurbiedukhin:
Should seal be added into theme as separate dom element?

assets/js/theme/common/aria/radioOptions.js
```diff
@@ -40,12 +41,14 @@ const handleItemKeyDown = itemCollection => e => {
     case ariaKeyCodes.UP: {
         const prevItemIdx = calculateTargetItemPosition(lastCollectionItemIdx, itemIdx - 1);
         itemCollection.get(prevItemIdx).focus();
+        setCheckedRadioItem(itemCollection, itemIdx - 1);
```
BC-tymurbiedukhin, line 44: If we are moving to instant option selection, is there is an ability to remove SPACE and Enter switch case

bc-alexsaiannyi: yes, I believe it's safe to remove them

BC-tymurbiedukhin (approved):
LGTM

yurytut1993 (approved):

PR #2032, opened 2021-04-06
fix(storefront): BCTHEME-449 remove main tag duplicates

bc-alexsaiannyi:
#### What?

This PR removes duplicates for main tag on store pages due to using a main tag with elements that have id="main=content".

#### Tickets / Documentation
- [BCTHEME-449](https://jira.bigcommerce.com/browse/BCTHEME-449)

bctheme-449_404_not_found
bctheme-449_brands
bctheme-449_category
bctheme-449_contact_us
bctheme-449_maintenance
bctheme-449_search

### DIFF
--- a/templates/components/common/body.html
+++ b/templates/components/common/body.html
@@ -1,4 +1,4 @@
-<main class="body" id='main-content' role='main' data-currency-code="{{currency_selector.active_currency_code}}">
+<main class="body" id="main-content" role="main" data-currency-code="{{currency_selector.active_currency_code}}">
     {{#block "hero"}} {{/block}}
     <div class="container">
         {{#block "page"}} {{/block}}
--- a/templates/pages/account/add-return.html
+++ b/templates/pages/account/add-return.html
@@ -4,7 +4,7 @@
 <h1 class="page-heading">{{lang 'account.returns.new_return' }}</h1>
 {{> components/account/navigation account_page='returns'}}
 
-<main class="account account--fixed account--addReturn">
+<div class="account account--fixed account--addReturn">
     <div class="account-body">
         <section class="account-content">
 
@@ -100,7 +100,7 @@
 
         </section>
     </div>
-</main>
+</div>
 
 {{/partial}}
 {{> layout/base}}
--- a/templates/pages/amp/category.html
+++ b/templates/pages/amp/category.html
@@ -28,13 +28,13 @@ category:
             {{category.name}}
         </h1>
     </div>
-    <main class="page-content" id="product-listing-container">
+    <div class="page-content" id="product-listing-container">
         {{#if category.products}}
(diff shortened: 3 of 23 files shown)


yurytut1993 (approved):

templates/components/common/body.html
```diff
@@ -1,4 +1,4 @@
-<main class="body" id='main-content' role='main' data-currency-code="{{currency_selector.active_currency_code}}">
+<main class="body" id="main-content" role="main" data-currency-code="{{currency_selector.active_currency_code}}">
```
BC-tymurbiedukhin, line 1: Do we really need to duplicate main tag and main role?

templates/pages/account/add-return.html
```diff
@@ -4,7 +4,7 @@
 <h1 class="page-heading">{{lang 'account.returns.new_return' }}</h1>
 {{> components/account/navigation account_page='returns'}}
 
-<main class="account account--fixed account--addReturn">
+<div class="account account--fixed account--addReturn">
```
BC-tymurbiedukhin, line 7: Should we use section instead of divs?

templates/pages/amp/category.html
```diff
@@ -28,13 +28,13 @@ <h1 class="categoryView-title">
             {{category.name}}
         </h1>
     </div>
-    <main class="page-content" id="product-listing-container">
+    <div class="page-content" id="product-listing-container">
```
BC-tymurbiedukhin, line 31: Could you please explain why do we need to remove main tag here?

templates/components/common/body.html
bc-alexsaiannyi: it's a [recommendation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main) on MDN to better support IE11.

templates/pages/account/add-return.html
bc-alexsaiannyi: as a [rule](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section) section tag should have a heading inside. And I've switched on section whenever it is possible but prefer to use div in case like this. since here div is used more as a wrapper for styling

templates/pages/amp/category.html
BC-tymurbiedukhin: Could you please double check this page and point to the place with 2nd main tag (I probably skipped it)

bc-alexsaiannyi: duplicated_tags

bc-alexsaiannyi: screen is attached. It's easier to see through it

BC-tymurbiedukhin: See, thanks

BC-tymurbiedukhin (approved):
LGTM

Issue #2038, opened 2021-04-16
price.html


obrien-k:
### Expected behavior
Use https for all links

### Actual behavior

`<meta itemprop="availability" itemtype="http://schema.org/ItemAvailability" 
                	content="http://schema.org/{{#if product.pre_order}}PreOrder{{else if product.out_of_stock}}OutOfStock{{else if product.can_purchase '===' false}}OutOfStock{{else}}InStock{{/if}}">`

Uses http for schema ?

### Steps to reproduce behavior

Review price.html

rowleyaj:
this can be closed as linked PR has been merged

PR #2039, opened 2021-04-16
Use https:// for schema markup

sacr3dc0w:
#### What?

This PR changes http://schema.org for https://schema.org. `http://` will be supported for the foreseeable future, but figured why wait for that to be phased out.

#### Tickets / Documentation

- [Schema.org FAQ (#19 Should we write 'https://schema.org' or 'http://schema.org' in our markup?)](https://schema.org/docs/faq.html#19)
- #2038

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Draft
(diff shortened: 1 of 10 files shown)


BC-tymurbiedukhin (approved):
LGTM

bc-alexsaiannyi (approved):

PR #2044, opened 2021-04-21
fix(storefront): DEV-426 Fix GitHub workflows for default storefront

bookernath:
#### What?

A recent release of stencil-cli broke these GitHub Actions examples, as the CLI arguments now require a channel ID in order to proceed without user interaction.

I am hardcoding this channel ID to `1`, which is the default storefront on every store. This will be the appropriate default for 99% of users for now, and those who wish to update it can adjust their functions accordingly.


bookernath:
ping @junedkazi

junedkazi (approved):

PR #2048, opened 2021-04-30
fix(storefront): Reviews pagination navigation buttons reload page and does not open the Reviews tab

yurytut1993:
#### What?

This PR has fix for reviews pagination - now Reviews tab (not Description tab) opens after click on navigation buttons.

#### Tickets / Documentation

[BCTHEME-350](https://jira.bigcommerce.com/browse/BCTHEME-350)

before fix

https://user-images.githubusercontent.com/66325265/116726540-1c790000-a9ec-11eb-867b-17467eefd22f.mov

after fix

https://user-images.githubusercontent.com/66325265/116726578-269afe80-a9ec-11eb-94ea-978d989fc6dc.mov


assets/js/theme/product/reviews.js
```diff
@@ -12,38 +12,39 @@ export default class {
...
-        });
+        $('#productReview_link').on('click', () => this.showReviews());
     }
 
-    collapseReviews() {
+    setupReviews() {
         // We're in paginating state, do not collapse
         if (window.location.hash && window.location.hash.indexOf('#product-reviews') === 0) {
-            return;
+            return this.showReviews();
```
BC-tymurbiedukhin, line 34: Could we put return after statement call for the reason it does not return anything

assets/js/theme/product/reviews.js
```diff
@@ -12,38 +12,39 @@ export default class {
...
         }
 
         // force collapse on page load
         this.$collapsible.trigger(CollapsibleEvents.click);
     }
 
+    showReviews() {
```
BC-tymurbiedukhin, line 41: Looks like toggleReviews describes this function in a better way

assets/js/theme/product/reviews.js
```diff
@@ -12,38 +12,39 @@ export default class {
...
+        $('.productView-reviewTabLink').trigger('click');
+        if (!this.$reviewsContentList.hasClass('is-open')) {
+            this.$collapsible.trigger(CollapsibleEvents.click);
+        }
```
BC-tymurbiedukhin, line 45: Could you please explain why do we need to trigger 2 click events?

assets/js/theme/product/reviews.js
yurytut1993: but this function can only show reviews, not hide them

assets/js/theme/product/reviews.js
yurytut1993: And it should not return anything. It was combined with `return` to avoid unnecessary line - in case we in `if` parent function must stop its execution

assets/js/theme/product/reviews.js
yurytut1993: np - first click activates tab with reviews (new code which is fix of the bug), second - activates reviews list (was previously working and just replaced to the function to be reusable)

assets/js/theme/product/reviews.js
bc-alexsaiannyi: `expandReviews()` ?..

assets/js/theme/product/reviews.js
bc-alexsaiannyi: btw, comment above says  "... do not collapse" and we execute `this.showReviews()`. Is it correct or we need remove this comment?

yurytut1993: Yes, it is correct. But I have updated it to make it more precise in context of new implementation

assets/js/theme/product/reviews.js
yurytut1993: &#9989;

assets/js/theme/product/reviews.js
BC-tymurbiedukhin: It is a good practice to put return empty if function should return nothing

bc-alexsaiannyi (approved):

assets/js/theme/product/reviews.js
yurytut1993: No, if tabs enabled reviews renders in tab, if not - just separate list

assets/js/theme/product/reviews.js
```diff
@@ -1,49 +1,65 @@
...
+        if ($reviewForm && $reviewForm.length) {
+            this.validator = nod({
+                submit: $reviewForm.find('input[type="submit"]'),
+                tap: announceInputErrorMessage,
+            });
+        }
 
-        this.$reviewsContent = $('#product-reviews');
+        this.$context = $context;
+        this.$reviewTabLink = $('.productView-reviewTabLink', this.$context);
```
BC-tymurbiedukhin, line 17: Do we really need $reviewTabLink and $reviewsContentList class properties or it could be replaced with vars?

assets/js/theme/product/reviews.js
```diff
@@ -1,49 +1,65 @@
...
+        this.$reviewsContent = $('#product-reviews', this.$context);
+        this.$reviewsContentList = $('#productReviews-content', this.$reviewsContent);
         this.$collapsible = $('[data-collapsible]', this.$reviewsContent);
 
-        this.initLinkBind();
+        if (this.context) {
+            collapsibleFactory('[data-collapsible]', { $context });
+        } else {
+            this.initLinkBind();
```
BC-tymurbiedukhin, line 25: Do we really need this method now or it could be replaced with its body?

assets/js/theme/product/reviews.js
yurytut1993: After quick view opens we can have 2 instances of reviews so we need to remember only nodes of the instance

assets/js/theme/product/reviews.js
yurytut1993: This method was in class previously. I think that it is consistent because all actions in the class works via methods. But if you want to remove method wrapping and in this case of one string use it without - also np

BC-tymurbiedukhin: It's a suggestion to remove 1 line body method, decision is up to you

BC-tymurbiedukhin (approved):
LGTM

bc-azvierieva (approved):

PR #2049, opened 2021-05-05
feat(storefront): BCTHEME-476 Scale focus trap for all modals

yurytut1993:
#### What?

For now focus trap exists on the most important theme modal. This PR scales this feature for all modals.

#### Tickets / Documentation

[BCTHEME-476](https://jira.bigcommerce.com/browse/BCTHEME-476)

Here are some new examples

https://user-images.githubusercontent.com/66325265/117110766-4dac5400-ad8f-11eb-9ce3-f4915869d38e.mov

https://user-images.githubusercontent.com/66325265/117110778-53a23500-ad8f-11eb-97cb-c2a7643d12d9.mov

https://user-images.githubusercontent.com/66325265/117110834-66b50500-ad8f-11eb-8c02-7f2dc6de0f95.mov

### DIFF
--- a/assets/js/theme/common/product-details.js
+++ b/assets/js/theme/common/product-details.js
@@ -3,7 +3,7 @@ import ProductDetailsBase, { optionChangeDecorator } from './product-details-bas
 import 'foundation-sites/js/foundation/foundation';
 import 'foundation-sites/js/foundation/foundation.reveal';
 import ImageGallery from '../product/image-gallery';
-import modalFactory, { showAlertModal } from '../global/modal';
+import modalFactory, { alertModal, showAlertModal } from '../global/modal';
 import { isEmpty, isPlainObject } from 'lodash';
 import { normalizeFormData } from './utils/api';
 import { isBrowserIE, convertIntoArray } from './utils/ie-helpers';
@@ -211,6 +211,11 @@ export default class ProductDetails extends ProductDetailsBase {
             this.updateProductAttributes(productAttributesData);
             this.updateView(productAttributesData, productAttributesContent);
             bannerUtils.dispatchProductBannerEvent(productAttributesData);
+
+            if (!this.checkIsQuickViewChild($form)) {
+                const $context = $form.parents('.productView').find('.productView-info');
+                modalFactory('[data-reveal]', { $context });
+            }
         });
     }
 
@@ -231,6 +236,10 @@ export default class ProductDetails extends ProductDetailsBase {
             role: roleType,
             'aria-live': ariaLiveStatus,
         });
+    }
+
+    checkIsQuickViewChild($element) {
+        return !!$element.parents('.quickView').length;
     }
 
     showProductImage(image) {
@@ -369,6 +378,10 @@ export default class ProductDetails extends ProductDetailsBase {
                 const tmp = document.createElement('DIV');
                 tmp.innerHTML = errorMessage;
 
+                if (!this.checkIsQuickViewChild($addToCartBtn)) {
+                    alertModal().$preModalFocusedEl = $addToCartBtn;
+                }
+
                 return showAlertModal(tmp.textContent || tmp.innerText);
             }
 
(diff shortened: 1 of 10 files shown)


assets/js/theme/common/product-details.js
```diff
@@ -233,6 +238,10 @@ export default class ProductDetails extends ProductDetailsBase {
         });
     }
 
+    checkIsInQuickView($element) {
```
bc-alexsaiannyi, line 241: what's your thought to use `isElementInsideQuickView` as method's name ?

assets/js/theme/global/modal.js
```diff
@@ -230,6 +234,10 @@ export class Modal {
     }
 
     onModalOpened() {
+        if (this.pending) {
+            this.$modal.one(ModalEvents.loaded, () => this.setupFocusTrap());
+        } else this.setupFocusTrap();
```
bc-alexsaiannyi, line 239: I would use/remove  curly brackets for both parts of `If statement` to stay consistent

assets/js/theme/common/product-details.js
yurytut1993: Honestly I think it is little bit excess, because function has $element as a parameter and its adding the same context

assets/js/theme/global/modal.js
yurytut1993: It was done to make code 2 strings less. According to you `else` should never be used without curly brackets, because it can not be used without `if` :)

bc-alexsaiannyi: it's not according to me)) Our ESLint extends airbnb style guides and we have no separate rule for blocks. According to [guide](https://github.com/airbnb/javascript#blocks--braces) it's a good practice to "Use braces with all multiline blocks".
ESLint also has its [rule](https://eslint.org/docs/rules/nonblock-statement-body-position) about block (you can check `Examples of correct code for this rule with the default "beside" option`)

assets/js/theme/global/modal.js
```diff
@@ -230,6 +234,10 @@ export class Modal {
     }
 
     onModalOpened() {
+        if (this.pending) {
+            this.$modal.one(ModalEvents.loaded, () => this.setupFocusTrap());
```
BC-tymurbiedukhin, line 238: important: Do we expect one here?

assets/js/theme/global/modal.js
```diff
@@ -204,6 +206,8 @@ export class Modal {
             this.focusTrap = focusTrap.createFocusTrap(this.$modal[0], {
                 escapeDeactivates: false,
                 returnFocusOnDeactivate: false,
+                allowOutsideClick: true,
```
BC-tymurbiedukhin, line 209: Question: why allowOutsideClick has been added?

assets/js/theme/global/modal.js
```diff
@@ -17,6 +17,7 @@ export const ModalEvents = {
     closed: 'closed.fndtn.reveal',
     open: 'open.fndtn.reveal',
     opened: 'opened.fndtn.reveal',
+    loaded: 'loaded.data.custom',
```
BC-tymurbiedukhin, line 20: Question: Why opened event does not satisfy our needs?

assets/js/theme/common/product-details.js
BC-tymurbiedukhin: checkIsQuickViewChild

yurytut1993: &#9989;

assets/js/theme/global/modal.js
yurytut1993: Saw such styling on freecodecamp but as we work with an eye on airbnb guide - agree with you. updated &#9989;

assets/js/theme/global/modal.js
yurytut1993: Yes. Event listener should be removed after execution

assets/js/theme/global/modal.js
yurytut1993: By default focusTrap do not allow outside click, but our modal has such functionality. allowOutsideClick was added to allow user close modal by clicking outside of it

assets/js/theme/global/modal.js
yurytut1993: when opened fired modal opened but content still can not be loaded and there are no element user can tab through

bc-alexsaiannyi (approved):

BC-tymurbiedukhin (approved):
LGTM

bc-alexsaiannyi (approved):

PR #2062, opened 2021-05-18
fix(storefront): BCTHEME-512 add translation for invalid quantity value error on cart

bc-alexsaiannyi:
#### What?

This PR adds a translation instead of hard-coded text. 

#### Tickets / Documentation

- [BCTHEME-512](https://jira.bigcommerce.com/browse/BCTHEME-512)

translation_fix

### DIFF
--- a/assets/js/theme/cart.js
+++ b/assets/js/theme/cart.js
@@ -90,7 +90,7 @@ export default class Cart extends PageManager {
             invalidEntry = $el.val();
             $el.val(oldQty);
             return swal.fire({
-                text: `${invalidEntry} is not a valid entry`,
+                text: this.context.invalidEntryMessage.replace('[ENTRY]', invalidEntry),
                 icon: 'error',
             });
         } else if (newQty < minQty) {
--- a/lang/en.json
+++ b/lang/en.json
@@ -67,6 +67,7 @@
         },
         "label": "Your Cart ({quantity, plural, one {# item} other {# items}})",
         "is_empty": "Your cart is empty",
+        "invalid_entry_message": "[ENTRY] is not a valid entry",
         "coupon_code": "Coupon Code",
         "discount": "Discount",
         "included_in_total": " Included in Total",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Draft
+- Added translation for invalid quantity value error on Cart. [#2062](https://github.com/bigcommerce/cornerstone/pull/2062)
 - Fixed displaying swatch name for multiple swatch options on page. [#2040](https://github.com/bigcommerce/cornerstone/pull/2040)
 - Added settings for payment banners. [#2021](https://github.com/bigcommerce/cornerstone/pull/2021)
 - Use https:// for schema markup. [#2039](https://github.com/bigcommerce/cornerstone/pull/2039)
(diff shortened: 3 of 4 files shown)


assets/js/theme/cart.js
```diff
@@ -90,7 +90,7 @@ export default class Cart extends PageManager {
             invalidEntry = $el.val();
             $el.val(oldQty);
             return swal.fire({
-                text: `${invalidEntry} is not a valid entry`,
+                text: `${invalidEntry} ${this.context.invalidEntryMessage}`,
```
BC-tymurbiedukhin, line 93: **Important**: Could we put string with placeholder into the context and then replace it with invalid entity value?

lang/en.json
```diff
@@ -67,6 +67,7 @@
         },
         "label": "Your Cart ({quantity, plural, one {# item} other {# items}})",
         "is_empty": "Your cart is empty",
+        "invalid_entry_message":"is not a valid entry",
```
BC-tymurbiedukhin, line 70: **Important**: Format the string

assets/js/theme/cart.js
bc-alexsaiannyi: changed.

lang/en.json
```diff
@@ -67,6 +67,7 @@
         },
         "label": "Your Cart ({quantity, plural, one {# item} other {# items}})",
         "is_empty": "Your cart is empty",
+        "invalid_entry_message": "{entry} is not a valid entry",
```
BC-tymurbiedukhin, line 70: Could we custom placeholder here instead of undefined?

bc-alexsaiannyi: I believe `undefined` will suite us better:
-  when we take this localized string from context a value already is set to `undefined` so it's safe to find it in the string and replace.
- using a custom placeholder we will add a risk that it can be accidentally translated later on. In this case we won't be able to find it & replace in js. 

If you still think a custom placeholder is better, I will change it.

BC-tymurbiedukhin: Screenshot 2021-05-19 at 13 45 57

bc-alexsaiannyi: yeah, I know:) we've got  a few of them but I still think they are little bit different cases. 
- firstly, we didn't have a choice there. Depends on language specifics SLIDE_NUMBER or SLIDES_QUANTITY can be placed in string in different order. As a result we can have 2 `undefined` and which of them replace with which value is a question.
- secondly, we've got a little trouble with custom placeholders so we even created a ticket [STRF-8764](https://jira.bigcommerce.com/browse/STRF-8764) to modify lang helper. I'm not sure it will be taken in work or not. Some quotes from there:
After discussion with Localization Team  we found a few downsides:

  - Adding a new kind of brackets will lead to explaining these use cases for Merchants and Dev community. When they should use square brackets and when should not. It can add more complexity for them and for us in a perspective since we need to keep this brackets as they are in en.json without translating.
  
  - New brackets will result in changing documentation (in more extended way comparing to updating lang helper).
  
still change to a placeholder?

BC-tymurbiedukhin: I believe we should follow the consistent approach, and if we have mechanism for similar task we could use it. In case if you want to change it, please change all square brackets logic

BC-tymurbiedukhin (approved):
LGTM

bc-alexsaiannyi:
Added a placeholder instead of  using parameter

BC-tymurbiedukhin (approved):
LGTM

yurytut1993 (approved):

bc-azvierieva (approved):

PR #2063, opened 2021-05-18
fix(storefront): BCTHEME-494 Translation Gap: Checkbox Option selection on Product.

bc-yevhenii-buliuk:
#### What?
This PR fixes translation gap on checkbox option selection (Yes/No) on Product.

#### Tickets / Documentation

- [BCTHEME-494](https://jira.bigcommerce.com/browse/BCTHEME-494)

#### Screenshots
494_fixed_translation_gap
494_cart_preview_translate


bc-alexsaiannyi:
@bc-yevhenii-buliuk why do we need this PR if we do not add anything new to the project?

yurytut1993 (approved):

templates/components/cart/content.html
```diff
@@ -47,7 +47,15 @@ <h2 class="cart-item-name">
                                     {{#if is_file}}
                                         <a href="/viewfile.php?attributeId={{id}}&cartitem={{../id}}">{{value}}</a>
                                     {{else}}
-                                        {{{ sanitize value}}}
+                                        {{#if value '==' 'Yes'}}
```
bc-vlad-dlogush, line 50: Can we use strict equality here?

templates/components/cart/content.html
```diff
@@ -47,7 +47,15 @@ <h2 class="cart-item-name">
...
+                                            {{lang 'common.yes'}}
+                                        {{else}}
+                                            {{#if value '==' 'No'}}
```
bc-vlad-dlogush, line 53: and here

templates/components/cart/content.html
bc-yevhenii-buliuk: @bc-vlad-dlogush we can use strict equality and this will also work, but the stencil-object gives only the string "Yes" or "No" as the checkbox value, and I think that the additional check for the type of the resulting value can be omitted in this case.

templates/components/cart/content.html
```diff
@@ -47,7 +47,15 @@ <h2 class="cart-item-name">
                                     {{#if is_file}}
```
bc-alexsaiannyi, line 47: a question: can we make code more linear?

```
{{#if is_file}}
     code
{{else if value '==' 'Yes'}}
    code
{{else if value '==' 'No'}}
   code
{{else}}
   code
{{/if}}
```

PR #2071, opened 2021-06-03
fix(storefront): BCTHEME-544 fix shift on change options modal on cart page

bc-alexsaiannyi:
#### What?

This PR fixes possible shift on option change popup due to different animation speed. (see attached video)

#### Tickets / Documentation
- [BCTHEME-544](https://jira.bigcommerce.com/browse/BCTHEME-544)

https://user-images.githubusercontent.com/67792608/120655899-cdf7cf00-c48b-11eb-8862-8be90a338cb4.mov

https://user-images.githubusercontent.com/67792608/120655916-d0f2bf80-c48b-11eb-99f6-994c1c986236.mov

### DIFF
--- a/assets/js/theme/cart.js
+++ b/assets/js/theme/cart.js
@@ -4,7 +4,7 @@ import checkIsGiftCertValid from './common/gift-certificate-validator';
 import { createTranslationDictionary } from './common/utils/translations-utils';
 import utils from '@bigcommerce/stencil-utils';
 import ShippingEstimator from './cart/shipping-estimator';
-import { defaultModal } from './global/modal';
+import { defaultModal, ModalEvents } from './global/modal';
 import swal from './global/sweet-alert';
 import CartItemDetails from './common/cart-item-details';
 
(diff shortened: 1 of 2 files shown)


BC-tymurbiedukhin (approved):
LGTM

assets/js/theme/cart.js
```diff
@@ -158,9 +158,11 @@ export default class Cart extends PageManager {
 
         utils.api.productAttributes.configureInCart(itemId, options, (err, response) => {
             modal.updateContent(response.content);
-            const $productOptionsContainer = $('[data-product-attributes-wrapper]', this.$modal);
-            const modalBodyReservedHeight = $productOptionsContainer.outerHeight();
-            $productOptionsContainer.css('height', modalBodyReservedHeight);
+            this.$modal.on(ModalEvents.opened, () => {
```
yurytut1993, line 161: 1. Looks like on every opening of the popup new event handler will be added and on the, for example third opening we will have three executions. Is it expected behaviour?
2. Do we cover a case when response come after popup opening already completed? Looks like now this code will not be executed

bc-alexsaiannyi: thanks. moved it above the code so modal eventListener is added once.

assets/js/theme/cart.js
```diff
@@ -158,9 +167,7 @@ export default class Cart extends PageManager {
 
         utils.api.productAttributes.configureInCart(itemId, options, (err, response) => {
             modal.updateContent(response.content);
-            const $productOptionsContainer = $('[data-product-attributes-wrapper]', this.$modal);
-            const modalBodyReservedHeight = $productOptionsContainer.outerHeight();
-            $productOptionsContainer.css('height', modalBodyReservedHeight);
+            this.$modal.triggerHandler(ModalEvents.opened);
```
yurytut1993, line 170: I believe we do not need to trigger opened event because it will be triggered by lib when popup opened

assets/js/theme/cart.js
yurytut1993: In that case event handler will be triggered when other content will be opened in this modal instance. It is better to add handler and remove it after execution

yurytut1993 (approved):

PR #2076, opened 2021-06-09
feat(storefront): BCTHEME-608 Translation mechanism for config.json values should be implemented

BC-tymurbiedukhin:
#### What?

Values located in config.json should become translatable

#### Tickets / Documentation

- [BCTHEME-291](https://jira.bigcommerce.com/browse/BCTHEME-291)
- [BCTHEME-608](https://jira.bigcommerce.com/browse/BCTHEME-608)


bc-alexsaiannyi (approved):

templates/components/products/card.html
```diff
@@ -1,3 +1,5 @@
+{{inject "isProductCardPresented" true}}
```
yurytut1993, line 1: Maybe we should add the same flag to `list-item component` since there is a possibility it can be used as a `cart component` (without list) in the future

assets/js/theme/common/utils/translations-utils.js
```diff
@@ -26,3 +26,27 @@ export const createTranslationDictionary = (context) => {
...
+/**
+ * defines Translation for values from page builder (locally could be found in config.json)
+ */
+export const translatePageBuilderValues = () => {
+    $('[data-page-builder-key]').each((_, selector) => {
+        const $item = $(selector);
+        const itemText = $item.text().trim();
+        const itemDefaultTranslation = $item.data('default-translation');
+
+        if (itemText === defaultPageBuilderValues[$item.data('page-builder-key')] && itemText !== itemDefaultTranslation) {
```
yurytut1993, line 48: Maybe it will be better to change `itemDefaultTranslation` to `JsonTranslation`

yurytut1993 (approved):

templates/components/products/card.html
BC-tymurbiedukhin: Since it used only in scope of list I prefer to leave as it is

assets/js/theme/common/utils/translations-utils.js
BC-tymurbiedukhin: It's a taste thing, I prefer to use itemDefaultTranslation

bc-azvierieva (approved):
Good job!

PR #2079, opened 2021-06-14
Add missing quotes for Country and State/Province fields of Shipping Estimator to pass functional tests

bc-yevhenii-buliuk:
#### What?

Added missing quotes for Country and State/Province fields of Shipping Estimator to pass functional tests


bc-alexsaiannyi (approved):
Congrats;)

yurytut1993 (approved):

PR #2089, opened 2021-07-12
BCTHEME-693: Update translation mechanism for config.json values

BC-tymurbiedukhin:
#### What?

Current translation mechanism for page builder values should be replaced with the next one:
Make default config.json values empty and replace it with lang file translations on a template level

#### Tickets / Documentation

[Jira](https://jira.bigcommerce.com/browse/BCTHEME-693)


templates/components/products/card.html
```diff
@@ -27,22 +25,34 @@
...
-            }}
-        {{else}}
-            {{#or price.sale_price_with_tax.value price.sale_price_without_tax.value}}
+            {{#if theme_settings.pdp_sold_out_label '===' ''}}
+                {{> components/products/product-badge
+                    badge-type='sold-out'
+                    badge_view=theme_settings.product_sold_out_badges
+                    badge_label=(lang "page_builder.pdp_sold_out_label")
+                }}
+            {{else}}
```
yurytut1993, line 34: can we replace `if` inside `badge_label attr ` to avoid 2x component usage?

yurytut1993, line 34: can we replace` if statement` inside `badge_label attr` to avoid 2x component usage?

BC-tymurbiedukhin: Do you know a way how to implement it?

yurytut1993: Looks like we can only do it inside component. But its up to you

yurytut1993 (approved):

BC-tymurbiedukhin:
@bc-azvierieva Could you please verify from your side?

bc-azvierieva (approved):

bc-alexsaiannyi (approved):

Issue #1990, opened 2021-02-17
Unify date format for release notes in CHANGELOG.md


mike-healy:
The release titles in `CHANGELOG.md` reference the release date, but historically have used both `YYYY-MM-DD` and `MM-DD-YYYY`.

`MM-DD` or `DD-MM` prefixed formats can be ambiguous as different regions use each.

I'd suggest using `YYYY-MM-DD` for all dates in this document.

rowleyaj:
i second this idea and think it might be worthwhile adopting the https://keepachangelog.com/en/1.0.0/ format

@BC-tymurbiedukhin what are you thoughts on this, as I know you're often the one to request changelog entries. I could add the changes to #2080 if you like?

BC-tymurbiedukhin:
We should align all dates to latest date format, it would be great to update previous values @rowleyaj

rowleyaj:
@BC-tymurbiedukhin any chance I can convince you to move towards `YYYY-MM-DD` instead? for the reasons that @mike-healy mentioned and also as mentioned on [keep a changelog](https://keepachangelog.com/en/1.0.0/):

> Confusing Dates
Regional date formats vary throughout the world and it's often difficult to find a human-friendly date format that feels intuitive to everyone. The advantage of dates formatted like 2017-07-17 is that they follow the order of largest to smallest units: year, month, and day. This format also doesn't overlap in ambiguous ways with other date formats, unlike some regional formats that switch the position of month and day numbers. These reasons, and the fact this date format is an [ISO standard](https://www.iso.org/iso-8601-date-and-time-format.html), are why it is the recommended date format for changelog entries.

I'm happy to raise a PR to update all the dates if you agree with this format

mike-healy:
> We should align all dates to latest date format, it would be great to update previous values @rowleyaj

The latest entry in the Changelog is _5.7.1 (07-09-2021)_

In future (after September) it won't be clear to a lot of people if that's 7th September or 9th July.
We have to think about where BigCommerce or the author is based, compare it to other updates, try to asses the likelihood that BC write for a US or international audience, and then make an inference based on all that.

BC-tymurbiedukhin:
I'm fine with updating date format, changes for stencil CLI have been requested, @rowleyaj should update existing changelog entries with a new PR

PR #2092, opened 2021-07-14
Changelog format changes

rowleyaj:
#### What?

Modify changelog date format to match ISO 8601. Also adds a message with link to keep a changelog to explain why this change was made. Changes draft to unreleased as this is clearer what those changes are.

I'll raise further PRs against the changelog at a later date as there are some releases where the date doesn't match the tag creation date / github release date. I'd like to add the links to the version titles so they can be easily diffed. ~Pre-release versions also aren't recorded at all - unsure if we want to add those?~ Pre-releases aren't added see [here](https://github.com/bigcommerce/stencil-cli/blob/d9c0f3d3febdf4b51ae45a32fb534d726e3729f3/lib/release/release.js#L99)

I don't think this PR needs a changelog entry - that would be a little too meta :)

#### Tickets / Documentation

- [1990](https://github.com/bigcommerce/cornerstone/issues/1990)

cc @mike-healy @BC-tymurbiedukhin

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## Draft
+## Unreleased
 
 ## 5.7.1 (2021-07-09)
 - Update lang files for some locales. [#2086](https://github.com/bigcommerce/cornerstone/pull/2086)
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,16 +1,20 @@
 # Changelog
+All notable changes to this project will be documented in this file.
 
-## Draft
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## 5.7.1 (07-09-2021)
+## Unreleased
```
BC-tymurbiedukhin, line 7: 'Draft' should be used here as we are updating changelog with help of CLI

rowleyaj: reverted that commit now @BC-tymurbiedukhin

ref: https://github.com/bigcommerce/stencil-cli/blob/d9c0f3d3febdf4b51ae45a32fb534d726e3729f3/lib/release/release.js#L96

BC-tymurbiedukhin (approved):
LGTM

rowleyaj:
@BC-tymurbiedukhin I don't have write access to `master` so can't merge this or my other PRs that have been approved. Whats is the process for getting them merged after an approval is added?

PR #2094, opened 2021-07-15
fix(storefront): BCTHEME-532 'undefined' is announced with screen reader while changing Product quantity on PDP

BC-tymurbiedukhin:
#### What?

Actual result:
"undefined" is announced instead of Product name with screen reader while changing Product quantity on PDP. Reproduces both on Mac and Windows

Expected result:
Product name is announced with screen reader while changing Product quantity on PDP, as it now works on Cart page.

#### Tickets / Documentation

[Jira](https://jira.bigcommerce.com/browse/BCTHEME-532)

Screenshot 2021-07-15 at 14 19 43


yurytut1993 (approved):

BC-tymurbiedukhin:
@bc-azvierieva Could you please verify from your side?

bc-azvierieva (approved):

PR #2115, opened 2021-09-02
fix(search): ES-2439 change debounce wait time for search

saurabhgedam:
#### What?

The low debounce time on the search query is causing the search api to be hit after every character. Bump up the debounce time to lessen the frequency. 

#### Requirements

#### Tickets / Documentation

https://jira.bigcommerce.com/browse/ES-2439

N/A

### DIFF
--- a/assets/js/theme/global/quick-search.js
+++ b/assets/js/theme/global/quick-search.js
@@ -32,7 +32,7 @@ export default function () {
         }
     };
 
-    // stagger searching for 200ms after last input
+    // stagger searching for 1200ms after last input
     const debounceWaitTime = 1200;
     const doSearch = _.debounce((searchQuery) => {
         utils.api.search.search(searchQuery, { template: 'search/quick-results' }, (err, response) => {
--- a/assets/js/theme/global/quick-search.js
+++ b/assets/js/theme/global/quick-search.js
@@ -33,6 +33,7 @@ export default function () {
     };
 
     // stagger searching for 200ms after last input
+    const debounceWaitTime = 1200;
     const doSearch = _.debounce((searchQuery) => {
         utils.api.search.search(searchQuery, { template: 'search/quick-results' }, (err, response) => {
             if (err) {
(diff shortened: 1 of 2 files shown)


assets/js/theme/global/quick-search.js
```diff
@@ -62,7 +62,7 @@ export default function () {
                 }, 100);
             }
         });
-    }, 200);
+    }, 1200);
```
saurabhgedam, line 65: Settling in on 1200 ms as a middle ground between the average typing speed, slowness while typing on touchscreens and a still acceptable wait time before fetching results.

bc-alexsaiannyi: as a suggestion: can we use a variable for this number which will explain it or just add a comment so it will make sense for future devs

saurabhgedam: Thanks @bc-alexsaiannyi. Added the change.

saurabhgedam: Can this be approved and merged ? I don't see any need for testing on this.

bc-alexsaiannyi:
@saurabhgedam  looks good for me. Please just add a changelog item that describes your change.

saurabhgedam:
added the changelog. Thanks @bc-alexsaiannyi 🙇‍♂️

bc-alexsaiannyi (approved):

assets/js/theme/global/quick-search.js
```diff
@@ -33,6 +33,7 @@ export default function () {
     };
 
     // stagger searching for 200ms after last input
+    const debounceWaitTime = 1200;
```
yurytut1993, line 36: Can you please also update the value in the comment?

saurabhgedam: thanks @yurytut1993 updated the value in changelog.

yurytut1993 (approved):

assets/js/theme/global/quick-search.js
yurytut1993: @saurabhgedam its merged

PR #2117, opened 2021-09-14
fix(storefront): BCTHEME-806 Password complexity error message does not reflect the requirements set in the store settings

jordanarldt:
This is a fix for **BCTHEME-806**

### What?
When typing a password when registering (or resetting your password), the password complexity error message does not reflect the complexity rules in the store settings. The error message uses a static context value set in the **lang** files. 

Example in **/lang/en.json**:
`"invalid_password": "Passwords must be at least 7 characters and contain both alphabetic and numeric characters."`
Screen Shot 2021-09-14 at 4 00 34 PM

Instead, the error message should display the complexity rules set by the store, which is accessible in the **settings.password_requirements.error** handlebars value. This is passed to the jsContext, it is just not being used for the error message. I've corrected this in my code fix.

### Requirements

### Proof of Fix
Screen Shot 2021-09-14 at 4 43 53 PM

@bigcommerce/themes-team @jfasanobc


assets/js/theme/auth.js
```diff
@@ -89,15 +89,15 @@ export default class Auth extends PageManager {
...
+        const { password: enterPassword, password_match: matchPassword } = this.validationDictionary;
         const newPasswordForm = '.new-password-form';
         const newPasswordValidator = nod({
             submit: $(`${newPasswordForm} input[type="submit"]`),
             tap: announceInputErrorMessage,
         });
         const passwordSelector = $(`${newPasswordForm} input[name="password"]`);
         const password2Selector = $(`${newPasswordForm} input[name="password_confirm"]`);
-        const errorTextMessages = createPasswordValidationErrorTextObject(enterPassword, enterPassword, matchPassword, invalidPassword);
+        const errorTextMessages = createPasswordValidationErrorTextObject(enterPassword, enterPassword, matchPassword, this.passwordRequirements.error);
```
bc-alexsaiannyi, line 100: @jordanarldt, I believe there won’t be so easy way. I see the point of these changes but still we need to think about translations.

We’re supposed to provide a localised versions for all validations like invalid password, email or other input to our merchants. As you can see from the code `this.validationDictionary` is based on `validation_messages` object  from `en.json` (injected via **base.html**, line56). 
Each phrase there (for instance, `invalid_password`) is being translated  to all supported languages and can be found in other lang files. Since you’re removing it and start to use `this.passwordRequirements.error` instead, you should find a way to localise it. 
If you can’t find how it can be implemented via theme then it should be found a responsible team with a newly created ticket.
 
P.S. not sure, but eventually the last option can be more realistic since `this.passwordRequirements.error` text is re-generated based on Merchant’s settings.  

One more direction to think about -  we can use `this.passwordRequirements.error` instead of invalid_password within lang files as a template with some sort of Placeholder(s) for min number of characters (`minlength` field from `this.passwordRequirements` obj) and other parts and replace it/them via js (**assets/js/theme/common/utils/form-utils.js** ) on line 213 with `requirements.minlength` for example.

jordanarldt: @bc-alexsaiannyi During my testing, I was surprised to find that the **this.passwordRequirements.error** is actually automatically localized already, which means there's no need to add translation keys for this type of error.

image

Since the passwordRequirements object is automatically generated based on the merchant's password settings, I don't think adding a key in the translation files is a feasible option, which might explain why the message is already localized depending on the store settings.

jordanarldt: @bc-alexsaiannyi As you mentioned, I do believe that the only other option would be to add the placeholder values in the translation files - but as it looks right now, the current localization that happens automatically on the phrase seems to be good enough, so is adding it to the translation files necessary?

bc-alexsaiannyi: @jordanarldt good point! I believe adding a new value to the translation files would be redundant step in this case.

bc-alexsaiannyi:
LGTM

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Draft
+### Fixed
+- Fixed password complexity error not displaying the complexity rules set in the store settings
```
yurytut1993, line 9: Can you please update this to be consisent with other items in the file?

jordanarldt: @yurytut1993 Fixed, let me know if it looks good!

yurytut1993: I mean you should add link to your pr like
`- Fixed password complexity error not displaying the complexity rules set in the store settings [#2117](https://github.com/bigcommerce/cornerstone/pull/2117)`

jordanarldt: @yurytut1993 Ahh okay, sorry about that! Fixed. These PRs are a little different from the other ones I've done :)

yurytut1993: No problem:) resolved

yurytut1993 (approved):

bc-vlad-dlogush (approved):

bc-azvierieva (commented):
@jordanarldt Hey Jordan, could you please squash commits into one and fix merge conflicts on this PR. After that it will be ready for merge!

Issue #2122, opened 2021-09-21
search results heading is not rendering correctly


AnirudhSingal91:
### Expected behavior
on search for keyword "shoe" (for example), div with id = "search-results-heading" should show 3 results for shoe

### Actual behavior
it instead shows 3 results for { search_query }

### Steps to reproduce behavior
 search for keyword "shoe" in the home page search bar

pootle85:
Hi, I am experiencing the same behaviour, seems a bit strange that this has just happened out of the blue.

Thanks,

Neil.

danvisionarybrands:
I can confirm the same is happening as well with all four of our stores, while no changes were made internally to any of the areas that would affect this. I was just about to review some source code to see what is going on when I figured it would be best to check Github first for other complaints. Glad to see it is not just our stores being affected.

**Search Results for { search_query }**

VisionsDeveloper:
Guyes, 

![image](https://user-images.githubusercontent.com/52342724/135303571-8fdfc345-2959-418c-8dbd-535811478346.png)

You can resolve this issue by simply remove the single quote from the en.json file.

![image](https://user-images.githubusercontent.com/52342724/135303790-ceb75fd9-c6be-4883-b6a3-137d96555dde.png)

It's working for us.

Thanks,
VD

enkota:
All good with @VisionsDeveloper's fix, but when we have modified over 100 stores, this is not a quick fix... BC support have logged an engineering ticket with this issue, so hopefully it can be resolved server side.

I have also noticed the issue appears when changing an option set on a product at cart level too.

![Screenshot 2021-10-06 at 10 52 07](https://user-images.githubusercontent.com/13979799/136180913-1548c99d-d16b-4ce1-b3cd-52e58d1eae85.jpg)

danvisionarybrands:
I can confirm on our end the issue is resolved with all four of our stores. We were notified about the BC engineering ticket as well from support and it seems it has been resolved on the server side. While editing the en.json file would have sufficed, as @enkota mentioned, this issue arose on the server side so it would be ideal for BC to fix it as opposed to having to push out new theme versions for all of our stores.

danvisionarybrands:
@enkota Confirm if you are still having the "Configure {name}" issue in the cart configurator modal as we were also affected, I reported it in issue #2128, however it seems the BC engineering ticket fixed both issues as of today.

PR #2127, opened 2021-09-27
StoreFront - Fix search results heading is not rendering correctly #2122

snehal-psdcenter:
#### What?

This PR addresses an issue with search for keyword([search_query](https://cornerstone-light-demo.mybigcommerce.com/search.php?search_query=shoes)) on  Search results page.

**HTML Entities**

It's happening due to the HTML Entity.

Reserved characters in HTML must be replaced with character entities because. sometimes it's creating an issue while using with any language code. 

Some characters are reserved in HTML.

Character entities are used to display reserved characters in HTML.

A character entity looks like this:
`&entity_name;
OR
&#entity_number;`

EXxample
   - To display a less than sign (<) we must write: `&lt;` or `&#60;`
   - To display single quotation mark/apostrophe (') we must write: `&apos;` or `&#39;`

**Note:** Entity names are case sensitive. Browsers may not support all entity names, but the support for entity numbers is good.

#### Error

-- [Error #2122](https://github.com/bigcommerce/cornerstone/issues/2122)
-- [{ search_query }](https://cornerstone-light-demo.mybigcommerce.com/search.php?search_query=shoes)

#### Screenshot
search_query

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -839,8 +839,8 @@
         "results": {
             "form_label": "Search Keyword:",
             "form_button_text": "Search",
-            "count": "{ count, plural, one {# result} other {# results} } for '{ search_query }'",
-            "quick_count": "{ count, plural, one {# product result} other {# product results} } for '{ search_query }'",
+            "count": "{ count, plural, one {# result} other {# results} } for &#39;{ search_query }&#39;",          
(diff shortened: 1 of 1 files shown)


BC-krasnoshapka:
hi @snehal-psdcenter 
does it make sense to fix other files in /lang folder?
please don't forget to update CHANGELOG

Issue #2135, opened 2021-10-21
Invalid object type for field "author"


mattcoy-arcticleaf:
Google has recently started flagging the author field in the default BC reviews system as invalid. I believe this is because the code is trying to set a value for the author as a string, but Google wants author to be an object with a type and a name.

Current code in reviews.html:
`<meta itemprop="author" content="{{name}}">`

Proposed fix:
```
<div itemprop="author" itemtype="https://schema.org/Person">
  <meta itemprop="name" content="{{name}}">
</div>
```

As a side note, the product microdata should probably be removed in favor of the newer LD+JSON setup. I've written this out for some of our clients, and it seems to work nicely.

### Steps to reproduce behavior
https://search.google.com/test/rich-results/result/r%2Freview-snippet?id=T-nYncGb1klfl_oeaUCw3g
![Screen Shot 2021-10-21 at 10 16 19 AM](https://user-images.githubusercontent.com/47044676/138296327-febe5f97-9564-4005-a559-ab40309f034f.png)

mattcoy-arcticleaf:
Note: this is now resolved

PR #2138, opened 2021-11-01
Replace schema microdata with LD+JSON tag

mattcoy-arcticleaf:
#### What?

The LD+JSON approach to schema.org data has been widely used for years now. It offers a much clearer, more concise way of providing the data necessary for Rich Results to search engines, such as Google. 

This change cleans up some of the bugs with the existing microdata, such as the description starting with "Description 1 Review", the author being invalid in product reviews, and a missing "priceValidUntil" field. It also adds the number to the gtin field, making it more specific (IE. "gtin12" instead of just "gtin").

More importantly, it separates the microdata from the product content, allowing for users to edit their site's code without having to worry about breaking the delicate data structure. 

Lastly, it makes editing the actual structured data very easy. For example, if a merchant wanted to list out their variants as separate offers using gql data in the existing Cornerstone setup, they would need to pass this data through to the price component, then go back up, loop through the gql, making it very difficult. With the new setup, they would simply loop through the gql response in this single LD+JSON tag!

#### Requirements

#### Tickets / Documentation

No specific requests here, but it does solve the following issue:
- https://github.com/bigcommerce/cornerstone/issues/2135

##### Before edits:
Screen Shot 2021-11-01 at 6 26 40 PM

##### After edits:
Screen Shot 2021-11-01 at 6 33 22 PM

#### Notes

I am willing to accept a different solution to passing the description through. This way apparently works fine for Google, and does not break, but it does look a bit ugly.

### DIFF
--- a/templates/components/products/schema.html
+++ b/templates/components/products/schema.html
@@ -1,0 +1,61 @@
+<script type="application/ld+json">
+    {
+        "@context": "https://schema.org/",
+        "@type": "Product",
+        "name": {{{JSONstringify product.title}}},
+        {{#if product.sku}}"sku": "{{product.sku}}",{{/if}}
+        {{#if product.mpn}}"mpn": "{{product.mpn}}",{{/if}}
+        {{#if product.gtin}}"gtin{{length product.gtin}}": "{{product.gtin}}",{{/if}}
+        "url" : "{{product.url}}",
+        {{#if product.brand}}
+        "brand": {
+            "@type": "Brand",
+            "url": "{{product.brand.url}}",
+            "name": {{{JSONstringify product.brand.name}}}
+        },
+        {{/if}}
+        "description": "{{encodeURI (sanitize product.description)}}",
+        "image": "{{getImage product.main_image 'zoom_size' (cdn theme_settings.default_image_product)}}",
+        {{#and settings.show_product_reviews product.reviews.list.length}}
+        "aggregateRating": {
+            "@type": "AggregateRating",
+            "ratingValue": "{{product.rating}}",
+            "reviewCount": "{{product.num_reviews}}"
+        },
+        "review": [
+        {{#each product.reviews.list}}
+            {
+                "@type": "Review",
+                "author": {
+                    "@type": "Person",
+                    "name": {{{JSONstringify name}}}
+                },
+                "datePublished": "{{date}}",
+                "reviewBody": {{{JSONstringify text}}},
+                "name": {{{JSONstringify title}}},
+                "reviewRating": {
+                    "@type": "Rating",
+                    "bestRating": "5",
+                    "ratingValue": "{{rating}}",
+                    "worstRating": "1"
+                }
+            }{{#unless @last}},{{/unless}}
+        {{/each}}
+        ],
+        {{/and}}
+        "offers": {
+            "@type": "Offer",
+            "priceCurrency": "{{currency_selector.active_currency_code}}",
+            {{#if product.price.price_range}}
+            "minPrice": "{{#if product.price.price_range.min.with_tax}}{{product.price.price_range.min.with_tax.value}}{{else}}{{product.price.price_range.min.without_tax.value}}{{/if}}",
+            "maxPrice": "{{#if product.price.price_range.max.with_tax}}{{product.price.price_range.max.with_tax.value}}{{else}}{{product.price.price_range.max.without_tax.value}}{{/if}}",
+            {{else}}
+            "price": "{{#if product.price.with_tax}}{{product.price.with_tax.value}}{{else}}{{product.price.without_tax.value}}{{/if}}",
+            {{/if}}
+            "itemCondition" : "https://schema.org/{{#if product.condition}}{{product.condition}}{{else}}New{{/if}}Condition",
+            "availability" : "https://schema.org/{{#if product.pre_order}}PreOrder{{else if product.out_of_stock}}OutOfStock{{else if product.can_purchase '===' false}}OutOfStock{{else}}InStock{{/if}}",
+            "url" : "{{product.url}}",
+            "priceValidUntil": "{{moment (moment add='31536000000') 'YYYY-MM-DD' }}"
+        }
+    }
+    </script>
--- a/templates/components/products/schema.html
+++ b/templates/components/products/schema.html
@@ -58,4 +58,4 @@
             "priceValidUntil": "{{moment (moment add='31536000000') 'YYYY-MM-DD' }}"
         }
     }
-    </script>
+</script>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Draft
 - Fixed Quick View modal "X" button focus bubble being slightly off center [#2130](https://github.com/bigcommerce/cornerstone/pull/2130)
 - Apply dependency updates (jest & lighthouse). [#2132](https://github.com/bigcommerce/cornerstone/pull/2132)
+- Replace schema microdata with LD+JSON tag. [#2138](https://github.com/bigcommerce/cornerstone/pull/2138)
 
 ## 6.1.1 (10-01-2021)
 - Fix product images on PDP has clipped outline. [#2124](https://github.com/bigcommerce/cornerstone/pull/2124)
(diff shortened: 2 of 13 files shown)


Tiggerito:
Looks nice. A few comments:

I've seen a few json-ld solutions where they place all the code in its own file and reference it in the head section. This would make it easier to maintain/modify and reduce duplication (like the amp version).

Variables could be used at the top of the structured data section to make it easier to modify behaviour. 

Some sites use third party review systems, which means marking up the BC reviews could end up being deceptive. Is there a way to detect if the BC reviews are showing? 

I'm not keen on adding a fake priceValidUntil value just to remove a warning. At some point people might make sites honour the price from a year ago!

Test with empty values for author name, product description etc.

mattcoy-arcticleaf:
Thanks for taking a look, Tony. Much appreciated!

Regarding the code being placed in a separate component, I could certainly do that. When I originally built this, I ignored the amp files altogether. However, I realized I had to update them because they rely on the same microdata, so I kind of added the same snippet to the amp template as an afterthought. I'm not even 100% sure if it works there (untested), but it should be fine. I do like the idea of placing it into a component so long as the data being passed through should be identical.

Which properties would you like to see as variables? I know that variables add to the memory usage, and there is a cap, and I don't want to compete with variable-heavy themes for that memory.

There is a check in place to ensure the reviews portion is not included if the store settings for reviews is off:
`{{#and settings.show_product_reviews product.reviews.list.length}}`

I'm not super keen on a fake priceValidUntil either. I am open to suggestions here. Google seems to really want it, and products do score better when they have it. Should I just make it shorter, like a week or 2 in the future?

Tiggerito:
With AMP Google recommends adding the exact same Structured Data but I don't think it is critical. AMP pages are canonicalized to the original page so for indexing they are of little significance.

Interesting to know about variable usage. I use them quite a bit in my [social tag system](https://bigcommerce.websiteadvantage.com.au/rich-snippets/articles/advanced-social-meta-tags-for-bigcommerce/) to make it easier for store owners to customise the code.

At the moment you would not be using any variables. But you may want to give developers more control in the future. Like setting how priceValidUntil works.

That review check should work.

With the priceValidUntil, how do you determine that products "score better"? Would be interesting to know if it makes a difference (or is even used).

mattcoy-arcticleaf:
I guess you are right, I don't really know that the products score better. However, we had a client who swore by it!

I will go ahead and put the script tag in a component, and include the component on both pages.

Tiggerito:
I'd love to see an example of priceValidUntil being used. I don't like it when Google recommends a property but does not tell us what it is for.

In our solution we don't add priceValidUntil by default, but do give the store owner options to set it (fixed data or rolling date based on number of days), so that it is their choice to add it and suppress the warning.

p.s. great initiative here with the move to json-ld.

BC-Chkalov-Valerii (approved):

MisterOctober:
Howdy y'all! 

I have a question about the use of `encodeURI()` in the `description` key of schema.html (line 17) : 

After implementing these changes in our store's theme and validating the structured data with Google's tools, it appears that the spaces within the description text are escaped (`%20`) -- 

I believe this is the expected behavior of `encodeURI()` generally, but what is the benefit of its use here? 
Anywhere the description schema data is displayed, the escape sequences are shown literally (e.g., `this%20is%20the%20description`) instead of correctly parsed, and although it is not flagged as an error by the Google tools, we're concerned that this may be detrimental. 

The behavior we're observing is identical to that described here: [https://stackoverflow.com/questions/29862371/can-text-be-encoded-with-uricomponent-instead-of-escaped](url)

When we remove the `encodeURI()` statement, the description text still validates and appears readable and correctly-parsed. 

Is there any contraindication to omitting this (`encodeURI()`) statement from schema.html?

mattcoy-arcticleaf:
Without encodeURI, a quote character (", ') in the product description would break the entire LD+JSON tag. As I mentioned in the initial PR message, I am open to other suggestions here.

templates/components/products/schema.html
```diff
@@ -0,0 +1,61 @@
...
+        ],
+        {{/and}}
+        "offers": {
+            "@type": "Offer",
+            "priceCurrency": "{{currency_selector.active_currency_code}}",
+            {{#if product.price.price_range}}
+            "minPrice": "{{#if product.price.price_range.min.with_tax}}{{product.price.price_range.min.with_tax.value}}{{else}}{{product.price.price_range.min.without_tax.value}}{{/if}}",
+            "maxPrice": "{{#if product.price.price_range.max.with_tax}}{{product.price.price_range.max.with_tax.value}}{{else}}{{product.price.price_range.max.without_tax.value}}{{/if}}",
+            {{else}}
+            "price": "{{#if product.price.with_tax}}{{product.price.with_tax.value}}{{else}}{{product.price.without_tax.value}}{{/if}}",
```
yurytut1993, line 53: Hey @mattcoy-arcticleaf 👋 
Looks like `minPrice` and `maxPrice` fields are not recognized by [schema.org](https://validator.schema.org/?hl=en-US#url=https%3A%2F%2Fyurytutstore.mybigcommerce.com%2Fsport%2Ffootball-shorts%2F) They may be placed in `priceSpecification` inside `offers`
I am thinking about creating the pr with the fix, what do you think? Or I am missing something and everything is correct? 🤔
Also according to [Rich Results Test](https://search.google.com/test/rich-results/result/r%2Fmerchant-listings?id=MPIhxH1oSr8Xz-47ql8ybA) `price` field should be represented even with price range

Tiggerito: Google currently has no documented features based around minPrice and maxPrice. They do support AggregateOffer however they consider that as a set of prices from different merchants, like on an affiliate site. 

https://developers.google.com/search/docs/appearance/structured-data/product#merchant-listings_offer 

Maybe Google will add support if more sites use it :-)

yurytut1993: @mattcoy-arcticleaf thanks for clarification and contribution 👍 
@Tiggerito Google totally fine with `minPrice` and `maxPrice` on `offers` level, but https://schema.org/ validator does not recognize those fields unfortunately. So from my side it is better to have them inside `priceSpecification` to satisfy both validators

mattcoy-arcticleaf: @Tiggerito @yurytut1993 ,

Realistically, the correct way to display a range of prices for different variants is to make an offers object and specify the price (and other data) of each. However, with BigCommerce, this is only possible through using GraphQL, since the variant data is not available via the Stencil context. I've actually custom-coded this for a few sites, but didn't want to add it to Cornerstone for a number of reasons.

See the LD+JSON tag with offers here: https://www.fishusa.com/Berkley-Deep-Hit-Stick-Premium-Lure/

Tiggerito: Cool. I found an example that is showing a price range: https://www.fishusa.com/McFly-Foam/

Google does not support variants in a usable way at the moment (They say you have to have a unique URL for each one). They are looking into improving that.

The multiple offer looks to be a good undocumented workaround to get the price range without affecting merchant listings by using aggregateOffer.

Tiggerito: > @Tiggerito Google totally fine with `minPrice` and `maxPrice` on `offers` level, but https://schema.org/ validator does not recognize those fields unfortunately. So from my side it is better to have them inside `priceSpecification` to satisfy both validators

The Rich Result Test tool focuses on errors and warnings with markup used to get rich results within Google search. It ignores properties it does not care about. While the schema.org validator is a lot stricter with the spec saying those properties have to be in a PriceSpecification.

Adding `minPrice` and `maxPrice` to the PriceSpecification is valid, but at the moment it does not look like Google looks at it.

mattcoy-arcticleaf: @Tiggerito , the code I wrote on fishUSA provides a unique URL to each variant using the sku parameter. Does it actually work on that site? Tough to say. It would, however, work on a standard BC site since going to that URL would select the correct options. I believe this satisfies Google "unique URL" requirement.

Tiggerito: @mattcoy-arcticleaf I see several issues:

The variant URLs are canonicalised out via the canonical tag, so they will not be in the index or show in search.

Even if they are not canonicalised out, Google will most likely consider them duplicates of the main product page, and canonicalise them out anyhow.

Typically, there are no real links to the variant pages, making it unlikely that search engines will find or consider the URLs worth indexing. I see you provide links via the structured data, I suspect that is not enough of a signal for Google to decide to index them, especially when they have not documented support for URLs in offers.

This is why I think their previous solution for variants was unworkable on most e-commerce websites.

They've currently removed most of their documentation that defined how to mark up variants. All that is left is the requirement that they have their own page. The merchant centre documentation had different requirements, but they have gone.

Tiggerito: @mattcoy-arcticleaf saying all that, I think the multiple offer solution is the best option at the moment as it causes price ranges.

PR #2161, opened 2022-01-20
Add api host to github action examples

steve-ross:
#### What?
Fixes github workflow/actions by bypassing api host prompt on stencil init job

#### Tickets / Documentation

- closes  #2160 closes #2159

PR #2170, opened 2022-01-28
fix(storefront): Specify default API Host when initialising stencil

bobbyshaw:
#### What?

Automatic deployment workflow fails due to missing config files because without a specified API host this step fails silently waiting for interaction. 

See https://support.bigcommerce.com/s/question/0D74O000007xZEs/detail

#### Requirements

If you would like me to add a changelog entry, let me know and I'll update the PR.


bobbyshaw:
Closing, now I've spotted https://github.com/bigcommerce/cornerstone/pull/2161

PR #2171, opened 2022-01-28
add zh locale

imgarylai:
#### What?

Add zh and zh-TW locale

#### Requirements

N/A

#### Tickets / Documentation

N/A

N/A


BC-krasnoshapka:
hi @imgarylai 
thanks for your effort with translations. BigCommerce provides official translations for Cornerstone, storefront and other services. We're still working on supporting Chinese languages across storefront and other services, so for now we can't accept your PR because translations won't be complete across all interfaces. But we'll definitely use your work as a base.

imgarylai:
> hi @imgarylai thanks for your effort with translations. BigCommerce provides official translations for Cornerstone, storefront and other services. We're still working on supporting Chinese languages across storefront and other services, so for now we can't accept your PR because translations won't be complete across all interfaces. But we'll definitely use your work as a base.

Thank you @BC-krasnoshapka . Please let me know if you have any helps regrading the Mandarin translations. We have been using this translation in our store for two years. (But I forgot the reason why I didn't create the PR.) The store theme we are using is cornerstone. And we really want to keep the theme up to date with the upstream which is here. 

Looking forward to see Big Commerce supports Mandarin translations. 

Happy Lunar new year!

Issue #2178, opened 2022-02-14
Blog Post Titles Should be h1 tags


mattcoy-arcticleaf:
### Expected behavior
As the title of the blog post on the post page is the title of the page, it should be an h1 tag.

This is an ADA violation, as well as bad practice for SEO.

### Actual behavior
It is currently an h2 tag.

### Steps to reproduce behavior
There is no h1 tag on this page:
https://cornerstone-light-demo.mybigcommerce.com/journal/your-third-blog-post/

sacr3dc0w:
Opened PR with change. https://github.com/bigcommerce/cornerstone/pull/2179

mattcoy-arcticleaf:
Looks like this has been resolved now. Closing.

PR #2179, opened 2022-02-22
Update blog component to use H1 tags on posts

sacr3dc0w:
#### What?

Blog posts should use an H1 tag for a page header.

#### Tickets / Documentation

https://github.com/bigcommerce/cornerstone/issues/2178


mattcoy-arcticleaf:
@sacr3dc0w , Unfortunately, the change is slightly more complicated than this, since the post.html component is used to generate the posts on the listing page. So, either a variable should be passed through to make it an h2 on the listing page, or probably better to just not use the same component for both instances.

BC-krasnoshapka (approved):
Looks good

bc-yevhenii-buliuk (approved):

PR #2190, opened 2022-03-29
fix(storefront): BCTHEME-1006 When price list price is set for currency, the cart does not respect product's price

bc-vlad-dlogush:
#### What?

This PR fixes a bug that did not update the currency in the cart the first time changed currency.

#### Tickets / Documentation

- [BCTHEME-1006](https://jira.bigcommerce.com/browse/BCTHEME-1006)

https://user-images.githubusercontent.com/83779098/160631189-23c52ddf-10df-42b9-8654-ed42de2ab68d.mov

https://user-images.githubusercontent.com/83779098/160631163-1d71eef6-4c20-4789-845f-e8c7c1ea288d.mov

### DIFF
--- a/assets/js/theme/global/currency-selector.js
+++ b/assets/js/theme/global/currency-selector.js
@@ -1,7 +1,17 @@
 import swal from './sweet-alert';
 import utils from '@bigcommerce/stencil-utils';
 
+let currencySelectorCalled = false;
+
 export default function (cartId) {
+    if (!cartId) return;
+
+    if (!currencySelectorCalled) {
+        currencySelectorCalled = true;
+    } else {
+        return;
+    }
+
     function changeCurrency(url, currencyCode) {
         $.ajax({
             url,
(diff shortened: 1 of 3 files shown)


assets/js/theme/global/currency-selector.js
```diff
@@ -2,6 +2,17 @@ import swal from './sweet-alert';
...
+    if (!cartId) return;
+
+    let countOfHandlers;
+    const events = $._data($('[data-cart-currency-switch-url]')[0], 'events');
+
+    $.each(events, (type, handlers) => {
+        countOfHandlers = handlers.length;
+    });
+
+    if (countOfHandlers >= 1) return;
```
BC-krasnoshapka, line 14: Even though this method of using internal JQuery attribute to get handlers is suggested in many places, I wonder why can't we just do something like this?
```
let currencySelectorCalled = false;

export default function (cartId) {
    if (!cartId) {
        return;
    }

    if (!currencySelectorCalled) {
        currencySelectorCalled = true;
    } else {
        return;
    }
```

BC-krasnoshapka (changes requested):

assets/js/theme/global/currency-selector.js
bc-vlad-dlogush: You're right, thanks.

BC-krasnoshapka (approved):

yurytut1993 (approved):

bc-alexsaiannyi (approved):

bc-yevhenii-buliuk (approved):

bc-azvierieva (approved):

PR #2193, opened 2022-04-07
fix(storefront): BCTHEME-1072 Stored XSS within Wishlist creation

bc-vlad-dlogush:
#### What?

In the **add-wishlist.html** file, there should be `{{{}}}` because we should not escape, but should display the data as it is and send it as a form to the server.
In the **wish-list.html** file, there should be `{{}}` to prevent an XSS attack.

#### Tickets / Documentation

- [BCTHEME-1072](https://jira.bigcommerce.com/browse/BCTHEME-1072)


bc-alexsaiannyi:
Previously there was a [ticket](https://jira.bigcommerce.com/browse/BCTHEME-277) regarding that some special characters like `& ' " < > `render as Unicode. Can you add please screenshots or videos that everything works as expected

PR #2198, opened 2022-04-11
fix(storefront): BCTHEME-709 "Out of Stock" banner is duplicated and overlaps "Add to cart" button on PDP

bc-vlad-dlogush:
#### What?

This PR removes the duplicate "Out of Stock" banner.

#### Tickets / Documentation

- [BCTHEME-709](https://bigcommercecloud.atlassian.net/browse/BCTHEME-709)

before 709
after  709
before  AMP  709
after  AMP  709

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Draft
+- "Out of Stock" banner is duplicated and overlaps "Add to cart" button on PDP.[#2198](https://github.com/bigcommerce/cornerstone/issues/2198)
 
 ## 6.4.0 (05-11-2022)
 - Remove adminBar translations from da/no lang files [#2209](https://github.com/bigcommerce/cornerstone/issues/2209)
--- a/templates/components/amp/products/product-view.html
+++ b/templates/components/amp/products/product-view.html
@@ -26,8 +26,8 @@
             {{/each}}
         </amp-carousel>
     </section>
-    <div class="productView-action">
-        {{#or customer (if theme_settings.restrict_to_login '!==' true)}}
+    {{#or customer (if theme_settings.restrict_to_login '!==' true)}}
+        <div class="productView-action">
             <amp-iframe height="400" width="250"
                 layout="responsive"
                 sandbox="allow-forms allow-pointer-lock allow-popups allow-same-origin allow-scripts allow-top-navigation allow-modals"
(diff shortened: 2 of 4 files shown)


bc-yevhenii-buliuk (approved):

BC-krasnoshapka (approved):

yurytut1993 (approved):

bc-alexsaiannyi (approved):

BC-krasnoshapka (approved):

bc-jz (changes requested):
I do not think this is a proper way to address this issue of the double banner being displayed as it does not work in all cases. Specifically if you have a complex product that does not have any default option values set, see below on my test store where I copied your changes in `product-view.html`:
Screen Shot 2022-05-20 at 9 47 44 AM
The add to cart button is gone but the out of stock message is not displayed. If I were to set at least one default option value then the message would display when loading this page.

Which brings me to my second point. The behavior being depended on to display the out of stock message with this code block removed is an ajax call to the remote product-attributes endpoint. When that call is made, if the response indicates that the product is out of stock then we will display the out of stock message. I do not think we should rely on that ajax call for displaying out of stock on an initial page load and really should try to eliminate use of that call altogether.

I will add some more details shortly about where that call is being made and how I think we should address this issue of the double error messages.

bc-jz:
This logic here is what is triggering on initial page load whenever there are default option values set. It fires the remote product-attributes ajax call and is using the results to update the option data on the page including the OOS state of the product:
https://github.com/bigcommerce/cornerstone/blob/master/assets/js/theme/common/product-details.js#L90-L94

This code is what is adding the 2nd error message when the selected option values are OOS:
https://github.com/bigcommerce/cornerstone/blob/master/assets/js/theme/common/product-details-base.js#L394-L397

Traditionally this was needed because the context that we provided to the page was incomplete and incorrect for certain situations:
- partial default option selections would not indicate OOS status or price correctly
- products tracking inventory by variant did not indicate OOS state
- the "in_stock_attributes" data provided on initial page load could be wrong in certain situations

I believe we are at a point now with the stencil context where these items have been corrected and we can move away from making this ajax call on initial page load altogether. The data received in the stencil context should correctly match up with any default selections if there are any and take partial selections into account. The last piece of this was merged yesterday with this fix to OOS labeling when products are tracking inventory by variant:
https://github.com/bigcommerce/bigcommerce/pull/46317

So rather than remove the checks to `{{product.out_of_stock}}` that is being done in this PR we should leave those and instead remove this ajax call that can fire on initial page load.

bc-jz:
In summation, the reasons I recommend going this more complex route for a fix are:
- we should be able to load the initial state of the page purely from stencil context which is more performant than needing follow up calls
- we should strive to eliminate making additional bcapp calls wherever possible for the overall health of our platform
- ajax calls should ideally only be needed for changes in state after page load or for pulling dynamic data that we don't make available in stencil context

jordanarldt:
Hey @bc-vlad-dlogush I had some free time so I made a draft PR to fix the issue while also addressing Jason's points. Feel free to take a look here:
https://github.com/jordanarldt/cornerstone/pull/3/files

bc-as:
@BC-krasnoshapka @bc-vlad-dlogush can you check out @jordanarldt 's solution to see if this hits all of the points that @bc-jz mentioned?

PR #2230, opened 2022-06-28
fix(storefront): BCTHEME-1092 Make screen reader say all errors then …

huntario:
…each error while tabbing

Watching - https://github.com/bigcommerce/cornerstone/pull/2235 - before moving forward

#### What?

Currently screen readers do not read out the errors until after the second submission, this seems to be do to the error not being defined within 'announceInputErrorMessage' assets/js/theme/common/utils/form-utils.js:132 

So the role alert is not attached as expected. Once attached (after the second submission), the alert is too intrusive. The alerts interrupt each other causing the reading to not be audible.  

Also, once the user is tabbing through each input, it should remind them of the specific error for that input field. 

announceInputErrorMessage Does not seem to be firing as expected on any of the forms and should likely be replaced with different solution, but we will just replace it for the account signup for now; and remove it further from other forms on separate tickets. 

#### Requirements

#### Tickets / Documentation

- [BCTHEME-1092](https://bigcommercecloud.atlassian.net/browse/BCTHEME-1092)

Before:

https://user-images.githubusercontent.com/5630999/176069042-13fea788-9ce6-4539-9bc1-a23a86649a05.mp4

https://user-images.githubusercontent.com/5630999/176070511-bbabfb7b-17b9-4b0b-8998-56035d7d013d.mp4

After:

https://user-images.githubusercontent.com/5630999/176071029-ed65e5be-7de4-4104-9f62-3b1e66b8da32.mp4

### DIFF
--- a/templates/components/common/forms/date.html
+++ b/templates/components/common/forms/date.html
@@ -13,6 +13,8 @@
                     data-label="month"
                     data-input
                     aria-required="{{required}}"
+                    aria-labelledby="{{id}}"
+                    aria-live="polite"
                     {{#if private_id}}data-field-type="{{private_id}}"{{/if}}
             >
             {{> components/common/forms/date-options month isRequired=required}}
@@ -25,6 +27,8 @@
                     data-label="day"
                     data-input
                     aria-required="{{required}}"
+                    aria-labelledby="{{id}}"
+                    aria-live="polite"
                     {{#if private_id}}data-field-type="{{private_id}}"{{/if}}
             >
             {{> components/common/forms/date-options day isRequired=required}}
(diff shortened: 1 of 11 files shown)


bc-yevhenii-buliuk:
@huntario it would be nice to implement logic to set focus to the first non-valid input field when submitting the form.

templates/components/common/forms/date.html
```diff
@@ -37,6 +38,7 @@
                     data-label="year"
                     data-input
                     aria-required="{{required}}"
+                    aria-labelledby="{{id}}"
                     {{#if private_id}}data-field-type="{{private_id}}"{{/if}}
```
bc-yevhenii-buliuk, line 42: do we need to add the `aria-live` and `aria-labelledby` attributes to the day selection as well?

huntario: Good call

huntario:
Thanks for the review @bc-yevhenii-buliuk . I've addressed the notes you mentioned. Let me know if you have any remaining concerns.

bc-yevhenii-buliuk (approved):

BC-krasnoshapka (approved):

PR #2231, opened 2022-07-01
Add global region at the bottom of the navigation

bobbyshaw:
#### What?

Currently, it's not possible to add a widget to a global region within the header. The only available region is `header_bottom--global`. This is in the header template so is on all pages but is actually outside of the <header> element so may be styled differently resulting it feeling like it's below the header, not at the bottom of the header.

As an app developer, I would like to request at least one global region be added inside the header.

Happy for this to be a discussion on the best way to do this and it's appropriate to add further global regions too, e.g. top of this file as an alternative to banners.

#### Requirements

TBC

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Bump webpack-bundle-analyzer [#2229]https://github.com/bigcommerce/cornerstone/pull/2229
 - Make screen reader say all errors then each error while tabbing. [#2230]https://github.com/bigcommerce/cornerstone/pull/2230
 - Clarify customer order pagination. [#2241]https://github.com/bigcommerce/cornerstone/pull/2241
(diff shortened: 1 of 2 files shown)


BC-krasnoshapka:
@bc-as what do you think about adding such region to header?

bobbyshaw:
Hi @bc-as. We launched [a mega menu app](https://www.bigcommerce.com/apps/mega-menu-builder-by-space-48/). The current global header region mostly works but does mean that technically it's below (and outside of) the header, not "bottom" as described. As such, it can cause some styling issues.

bc-as:
@bobbyshaw this looks fantastic!  I'm glad to see a creative use of regions!  @BC-krasnoshapka this seems like a great addition to Cornerstone, can we get this merged and released with the next version?

BC-krasnoshapka:
Sure, no problem. @bobbyshaw can you please update CHANGELOG also?

bobbyshaw:
@BC-krasnoshapka I have updated the CHANGELOG.

BC-krasnoshapka (approved):

BC-krasnoshapka:
Thanks. LGTM

PR #2235, opened 2022-07-12
fix(storefront): BCTHEME-1128 Form Input Error Message and Accessibility.

bc-yevhenii-buliuk:
#### What?
This Pr fixes the accessibility of form fields. By adding the `role="status"` and `aria-live="polite"` attributes to the error messages, as well as the `"aria-describedby"` attribute associated with the `"id"`, the screen reader is able to read form fields and error messages on those fields.
Also added the ability to set focus to the first field with an error when submitting an empty form.

#### Tickets / Documentation
[BCTHEME-1128](https://bigcommercecloud.atlassian.net/browse/BCTHEME-1128)

#### Screen Recording

**before fixes**

https://user-images.githubusercontent.com/82589781/178459948-6d4810ad-b739-4508-9fc2-63985bd1287e.mov

**after fixes**

https://user-images.githubusercontent.com/82589781/178458834-bdb65d61-a4f2-4b5e-bb1e-deb97f51eb49.mov

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Draft
+- Form Input Error Message and Accessibility. [#2235](https://github.com/bigcommerce/cornerstone/issues/2235).
 
 ## 6.5.0 (06-24-2022)
 - Category icons do not appear in Search Form [#2221]https://github.com/bigcommerce/cornerstone/pull/2221
--- a/assets/js/theme/auth.js
+++ b/assets/js/theme/auth.js
@@ -112,6 +112,7 @@ export default class Auth extends PageManager {
         const createAccountValidator = nod({
             submit: `${this.formCreateSelector} input[type='submit']`,
             tap: announceInputErrorMessage,
+            delay: 0,
         });
         const $stateElement = $('[data-field-type="State"]');
         const emailSelector = `${this.formCreateSelector} [data-field-type='EmailAddress']`;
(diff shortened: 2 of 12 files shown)


huntario:
The focus on the latest element is a nice touch. 

I've been looking into this and related issues on the PRs below. In my opinion, we don't seem to get much for the added complexity of  'announceInputErrorMessage' and we should work towards removing it. Assuming the HTML elements have appropriate Aria attributes, it should work without 'announceInputErrorMessage'  which appears to not have been working for some time now (maybe never?). Is it ideal for code adding accessibility attributes to be tied to the form validation library?

That's the approach I've taken in these:
BCTHEME-1092 - https://github.com/bigcommerce/cornerstone/pull/2230
BCTHEME-1093 - https://github.com/bigcommerce/cornerstone/pull/2234
BCTHEME-1094 - https://github.com/bigcommerce/cornerstone/pull/2233

I will of course defer to whatever you and your team wish to do. I'll watch this PR and adapt mine above based on how this one moves forward.

Whatever approach we take we'll want to also consider the methods within - assets/js/theme/account.js - related to the forms on the various logged in customer pages, but not necessarily in this PR.

bc-yevhenii-buliuk:
@huntario After discussing this task within the team, we decided to support the idea of ​​separating the form validation logic from the accessibility logic. Indeed, we can avoid using the "announceInputErrorMessage" function which adds attributes to improve accessibility during form validation and set the required accessibility attributes for input fields as implemented in your [PR](https://github.com/bigcommerce/cornerstone/pull/2230).
It would also be nice to implement logic to set focus to the first non-valid input field when submitting the form.
We expect that this approach will be implemented in all your other PRs that solve this problem.
Thus, I close this PR so as not to duplicate another solution. Thank you for discussing and paying attention to this issue!

PR #2241, opened 2022-07-30
fix(storefront): BCTHEME-1077 clarify customer order pagination

huntario:
#### What?

The pagination enumeration text at the bottom of the customer account orders page displays innacurately once there are 7 or more pages of orders. This is because the pagination.links does not always return the total list of pages which is expected, so we cannot rely on the length of that array as indicator for the total pages. Instead we will return a horizontal list of page numbers which are links to the corresponding page of orders. 

#### Requirements

#### Tickets / Documentation

- [BCTHEME-1077](https://bigcommercecloud.atlassian.net/browse/BCTHEME-1077?filter=-3)

Before: Notice the unexpected page count behavior

https://user-images.githubusercontent.com/5630999/181863853-8dfc2673-189f-41d8-8849-b16abd58f167.mp4

After: More control and clarity for the user

https://user-images.githubusercontent.com/5630999/181863874-48d6db87-d003-4410-946e-ac8ad94a3922.mp4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - On customer message page, screen reader should say each error [#2234]https://github.com/bigcommerce/cornerstone/pull/2234
 - Bump webpack-bundle-analyzer [#2229]https://github.com/bigcommerce/cornerstone/pull/2229
 - Make screen reader say all errors then each error while tabbing. [#2230]https://github.com/bigcommerce/cornerstone/pull/2230
+- Clarify customer order pagination. [#2241]https://github.com/bigcommerce/cornerstone/pull/2241
 
 ## 6.5.0 (06-24-2022)
 - Category icons do not appear in Search Form [#2221]https://github.com/bigcommerce/cornerstone/pull/2221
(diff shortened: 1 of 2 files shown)


BC-krasnoshapka (approved):

bc-vlad-dlogush (approved):

mattcoy-arcticleaf:
@BC-krasnoshapka @bc-vlad-dlogush , it looks like this PR adds invalid HTML structure to the page. An <li> tag cannot be inside an <a> tag and an <a> tag cannot be inside a <ul> tag.

huntario:
Thanks for taking a look at this PR @mattcoy-arcticleaf. Are you seeing a direct user impact related to this?  Historically you're certainly correct but there may be some debate about this with the current wording of the standard - https://stackoverflow.com/questions/12086453/wrap-anchor-tag-around-li-element .

In either case with the standard's handling of the list owner being determined by the closest ul, ol, or menu ancestor (not necessarily the direct parent), I'd anticipate that supported browsers will handle this correctly - https://html.spec.whatwg.org/multipage/grouping-content.html#the-li-element . It's also a non-public facing page so there shouldn't be any SEO concerns

For now this is the smallest change available that will allow for the orders pagination mechanism to work in a predictable/desirable way for the user, so unless there is some direct user impact with this change, we will need to treat the validity concern as a separate issue and weigh any additional changes against the risk of negative user impact.

mattcoy-arcticleaf:
@huntario it will be flagged by accessibility scanners. BigCommerce claims Cornerstone to be AA WCAG compliant. This change goes against that claim.

Issue #2213, opened 2022-05-17
Unparsable structured data > Incorrect value type - Google Search Console


LanetheGreat:
### Expected behavior
When the new LD+JSON is rendered (added from PR #2138), the Review Author's `name` field is expected to render the same as the template when anonymous users post reviews, which by default should be "Unknown" from the /lang/en.json file (under products.reviews.anonymous_poster).

### Actual behavior
I believe this was an issue similar to #2135 before the PR was merged. But post-merge when a product's schema is rendered, that contains a Review made by an anonymous user, it renders nothing for the `name` field (not even an empty JSON string), creating invalid JSON that Google Search Console fails to parse; categorized under "Unparsable structured data > Incorrect value type" with the following example of invalid JSON:

![search_console_json_screenshot](https://user-images.githubusercontent.com/15223575/168872581-06915fd6-0b4e-4107-81dc-e36742df7ea6.png)

I noticed this issue in Google Search Console for my company's site after merging that PR into our template.

### Steps to reproduce behavior
Create an anonymous review for any product and validate the LD+JSON produced by that product's page.

### Recommended Fix
Replace line 31 in /templates/components/products/schema.html from:
```
27            {
28                "@type": "Review",
29                "author": {
30                    "@type": "Person",
31                    "name": {{{JSONstringify name}}}
32                },
33                "datePublished": "{{date}}",
```
to:
```
27            {
28                "@type": "Review",
29                "author": {
30                    "@type": "Person",
31                    "name": {{#if name}}{{{JSONstringify name}}}{{else}}{{{JSONstringify (lang 'products.reviews.anonymous_poster')}}}{{/if}}
32                },
33                "datePublished": "{{date}}",
```

sacr3dc0w:
Fixed in #2255

PR #2255, opened 2022-09-05
Fix "incorrect value type" for anonymous reviews in Google Search Console

sacr3dc0w:
#### What?

> When the new LD+JSON is rendered (added from PR https://github.com/bigcommerce/cornerstone/pull/2138), the Review Author's name field is expected to render the same as the template when anonymous users post reviews, which by default should be "Unknown" from the /lang/en.json file (under products.reviews.anonymous_poster).

#### Tickets / Documentation

https://github.com/bigcommerce/cornerstone/issues/2213

**BEFORE**
1 BEFORE

**AFTER**
2 AFTER


BC-krasnoshapka (approved):
LGTM, thanks

bc-vlad-dlogush (approved):

PR #2261, opened 2022-09-12
Update key assets with preload headers

bookernath:
#### What?

Update the key render-blocking resources in `base.html` to be tagged as preloadable; this will result in [preload headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Link) being sent for these resources.

Furthermore, when the Stencil storefront is behind Cloudflare, this will result in [HTTP 103 Early Hints](https://blog.cloudflare.com/early-hints/) being sent when appropriate, which can result in significant performance improvements for some clients.

Limiting this to only the most important, render-blocking resources for right now, as there is probably a point of diminishing returns or even nullifying the benefit if too many resources are tagged.

Outside of the theme, some Scripts in the Script Manager will also be preloaded.

#### Requirements

#### Tickets / Documentation

- [BCTHEME-1219](https://bigcommercecloud.atlassian.net/browse/BCTHEME-1219)

#### Proof

Here's a WebPageTest result showing the effect of Early Hints (resources start downloading before HTML is delivered):

https://www.webpagetest.org/result/220912_BiDcW5_D8J/1/details/#waterfall_view_step1

### DIFF
--- a/templates/layout/base.html
+++ b/templates/layout/base.html
@@ -24,12 +24,11 @@
             window.lazySizesConfig = window.lazySizesConfig || {};
             window.lazySizesConfig.loadMode = 1;
         </script>
-        <script async src="{{cdn 'assets/dist/theme-bundle.head_async.js'}}"></script>
+        <script async src="{{cdn 'assets/dist/theme-bundle.head_async.js' resourceHint='preload' as='script'}}"></script>
         
         {{getFontsCollection font-display='block'}}
         
-        <link rel="preload" href="{{cdn 'assets/dist/theme-bundle.font.js'}}" as="script">
-        <script async src="{{cdn 'assets/dist/theme-bundle.font.js'}}"></script>
+        <script async src="{{cdn 'assets/dist/theme-bundle.font.js' resourceHint='preload' as='script'}}"></script>
 
         {{{stylesheet '/assets/css/theme.css'}}}
 
@@ -48,9 +47,6 @@
         {{~inject 'carouselArrowAndDotAriaLabel' (lang 'carousel.arrow_and_dot_aria_label')}}
         {{~inject 'carouselActiveDotAriaLabel' (lang 'carousel.active_dot_aria_label')}}
         {{~inject 'carouselContentAnnounceMessage' (lang 'carousel.content_announce_message')}}
-
-        {{!-- Get this loading ASAP --}}
-        <link rel="preload" href="{{cdn 'assets/dist/theme-bundle.main.js'}}" as="script">
     </head>
     <body>
         <svg data-src="{{cdn 'img/icon-sprite.svg'}}" class="icons-svg-sprite"></svg>
@@ -66,7 +62,7 @@
                 window.stencilBootstrap("{{page_type}}", {{jsContext}}).load();
             }
         </script>
-        <script async defer src="{{cdn 'assets/dist/theme-bundle.main.js'}}" onload="onThemeBundleMain()"></script>
+        <script async defer src="{{cdn 'assets/dist/theme-bundle.main.js' resourceHint='preload' as='script'}}" onload="onThemeBundleMain()"></script>
 
         {{{footer.scripts}}}
     </body>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add logic to collect Product Details data and send it to the BC App stencil template through custom event [#2270](https://github.com/bigcommerce/cornerstone/pull/2270)
 - Allow quantity of "0" in cart to remove item [#2266](https://github.com/bigcommerce/cornerstone/pull/2266)
 - Fix the issue with getting product details data if the product details form is valid on page load [#2271](https://github.com/bigcommerce/cornerstone/pull/2271)
+- - Update key render-blocking resources to be preloaded via HTTP headers/Early Hints [#2261](https://github.com/bigcommerce/cornerstone/pull/2261)
+
 
 ## 6.6.1 (09-14-2022)
 


rafa-avila-bc (approved):

Tiggerito:
This is great.

In the webpage test, I see the fonts CSS is loaded twice. An encoding issue makes the preload not match the actual CSS URL. This also means the file is still render-blocking and causing an extra delay on all resources found after it.

https://fonts.googleapis.com/css?family=Karla:400%7CMontserrat:400,500,700&display=block
https://fonts.googleapis.com/css?family=Karla:400|Montserrat:400,500,700&display=block

I'd also test the idea of preloading the font js bundle. With that, the browser may start to preload the web fonts if it has the time. I'm not 100% on if the current setup will preload before it sees the fonts in use.

The final big win I see is in the preload of the LCP image. The ability to add image preloads would make that a possible option. In your test, it would potentially cut 250ms off the LCP value.

bookernath:
Thanks for your feedback @Tiggerito!

> In the webpage test, I see the fonts CSS is loaded twice. An encoding issue makes the preload not match the actual CSS URL. This also means the file is still render-blocking and causing an extra delay on all resources found after it.

Nice catch. Looking into it!

> I'd also test the idea of preloading the font js bundle. With that, the browser may start to preload the web fonts if it has the time. I'm not 100% on if the current setup will preload before it sees the fonts in use.

I've made this update, it seems like a sensible change in any case.

> The final big win I see is in the preload of the LCP image. The ability to add image preloads would make that a possible option. In your test, it would potentially cut 250ms off the LCP value.

I think this is a good idea, and it may add some benefit. In practice, though, we already have a pretty good image loading strategy between responsive images + LQIP + the progressive loading nature of JPGs (regardless of whether or not we get "credit" for this in the LCP metric).

I also wonder if the carousel will always be above-the-fold content, it may not be on all storefront designs all the time, especially with Widgets in the mix for example. Or to phrase it another way, I wanted to put more thought into understanding when an image can benefit from preloading vs when that will be a waste of bandwidth, and I kept it out of this PR because I didn't have what felt like a perfect answer. Interested in feedback on this.

Tiggerito:
Good point on the responsive images. Currently, HTML preload can deal with them, but HTTP headers can't.

However, many images in a theme end up being none responsive. Or responsive with one size. So could be HTTP preloaded. 

Maybe it makes sense to let HTML preloads optionally generate the HTTP preload. It could block those with imgsrcset. That would be flexible. And/or let img tags optionally generate the HTTP preload as long as they do not have a srcset or data-srcset. It will not work 100% as I often see srcset lists with a fixed 'sizes', meaning only one image will ever be chosen.

Obviously more complex, so for another day. 

At the moment, I manually add image preloads and often have to fix lazyloading of prominent images. Our Core Web Vitals repost tells us which images are most often the LCP making them a strong candidate for preloading.

I did some testing of things like LQIP and preloads:

https://bigcommerce.websiteadvantage.com.au/tag-rocket/articles/improving-image-loading-without-javascript/

bookernath:
Thanks, I'll take a look at this!

I should mention that as part of this update we've made updates to [several handlebars helpers](https://github.com/bigcommerce/paper-handlebars/releases) (documentation coming soon) which will allow you to tag your own assets for resourcing hinting in custom themes, based on your own understanding of what's important on a particular page in your own theme. Cornerstone, naturally, needs to think very generically and "one-size-fits-all".

So if you're already putting images into preload `<link>` tags in the page, you can consider moving those to `{{cdn}}` helpers using the `resourceHint` argument to emit them as HTTP headers/Early Hints.

templates/layout/base.html
```diff
@@ -24,14 +24,13 @@
             window.lazySizesConfig = window.lazySizesConfig || {};
             window.lazySizesConfig.loadMode = 1;
         </script>
-        <script async src="{{cdn 'assets/dist/theme-bundle.head_async.js'}}"></script>
+        <script async src="{{cdn 'assets/dist/theme-bundle.head_async.js' resourceHint='preload'}}"></script>
         
-        {{getFontsCollection font-display='block'}}
+        {{getFontsCollection font-display='block' resourceHint='preload'}}
         
-        <link rel="preload" href="{{cdn 'assets/dist/theme-bundle.font.js'}}" as="script">
```
BC-krasnoshapka, line 31: @bookernath why do you think link with rel=preload is not needed anymore? this technique makes browser to load font bundle faster. what about [browsers that doesn't support ](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/103#browser_compatibility)early hints yet?

Tiggerito: It's basically Chrome for now. Not sure if others support HTTP header link preloads, which would be enough.

Another option is fetchpriority="high" on the async script. That would do the same thing on supported browsers. It also helps with images. You can also add it to link preloads, which are, by default, a lower priority.

I'd keep the preloads for now. It also means you can control sequencing. e.g. the way we preload main js towards at the end of the head so that other resources get first dibs. And request it for real in the footer to guarantee it runs late.

I personally load the fonts script earlier as it's required to preload the fonts, which we want to happen ASAP.

bookernath: My assumption was that this is duplicative when we are already sending Link headers. But even if it's duplicative, I have no reason to believe it's _harmful_ (browsers will likely identify that resource is already being downloaded and not start downloading it twice... I hope). Maybe this is worth a quick test.

Cloudflare will choose when to send HTTP 103 Early Hints vs just allowing the normal Link headers to go through based on their client detection and a few other factors.

Tiggerito: I'd assume the browser is smart enough not to make multiple requests. That is the case for other scenarios. I'd still test.

One gotcha is that the crossorigin attribute has to be the same. If it is different, it is considered a different resource to load.

I try and add crossorigin="anonymous" if it is allowed, as it lets me gather more error data. e.g. on theme-bundle.main.js

Tiggerito: I have mixed views on how to deal with the main bundle with regard to loading speed. It is CPU intensive and blocking (especially with sliders) which stops user interaction (FID, INP),  and delays LCP if it starts running before the page renders. So loading it quickly increases the chance of Web Vitals issues.

For one client I deliberately delayed the main bundle from running for a short period (100ms I think) and spilt up the slider initialisation code into separate delayed tasks (setTimeout per slider). That was enough to let the CPU spend time rendering and responding to user inputs in a reasonable time (<100ms).

The thing is, does it contain stuff that is required for the page to render properly, or for interaction to work? If so, it should load and run quickly. It would be nice if the non-critical work is delayed and split into mini tasks (like my slider solution). This delay typically needs to be longer than DCL as it can happen before rendering so tasks sitting on it can delay rendering.

In some cases, my slider solution did need some CSS adjustments to make sure the sliders looked correct before they were built. This is a good move anyhow as it stops layout shifts (CLS).

Tiggerito:
I just spotted some preload header issues on a client's store. Error in the Chrome console:

> \<link rel=preload\> must have a valid \`as\` value

I checked the headers and they look like this:

`<https://www.domain.com/script.js?external-type=bigcommerce>; rel=preload; type=application/javascript`

When I think they should be:

`<https://www.domain.com/script.js?external-type=bigcommerce>; rel=preload; as=script`

Tiggerito:
While I'm here. Here's another header:

` link: <https://fonts.gstatic.com>; rel=dns-prefetch, <https://fonts.googleapis.com>; rel=dns-prefetch, <https://cdn11.bigcommerce.com/s-qfy5foqczc>; rel=dns-prefetch`

A DNS prefetch is at the domain level. So I don't think including s-qfy5foqczc has any value.

Any reason why dns-prefetch is used and not preconnect?

rafa-avila-bc:
> I just spotted some preload header issues on a client's store. Error in the Chrome console:
> 
> > <link rel=preload> must have a valid `as` value
> 
> I checked the headers and they look like this:
> 
> `<https://www.domain.com/script.js?external-type=bigcommerce>; rel=preload; type=application/javascript`

This behavior and Chrome's warning also happen in the https://cfearlyhintstest.cc/ test store.

image

However, the following might not be completely right.

> When I think they should be:
> 
> `<https://www.domain.com/script.js?external-type=bigcommerce>; rel=preload; as=script`

Although MDN `Link` header docs state that "the header is semantically equivalent to the HTML `link` tag", the [`Link` header standard](https://www.rfc-editor.org/rfc/rfc5988#section-5) does not specify an `as` parameter. `Link` headers generation is limited by the standard.

CC @Tiggerito

Tiggerito:
Maybe because it is quite new/experimental and only on Chrome. These contain examples using 'as':

https://httpwg.org/specs/rfc8297.html
https://html.spec.whatwg.org/multipage/semantics.html#early-hints
https://blog.cloudflare.com/early-hints/
https://developer.chrome.com/blog/early-hints/

bookernath:
We've fixed the encoding issue which was causing URLs to not match and cause a cache miss.

We're fixing the `as` value next to fix Chrome's errors.

templates/layout/base.html
```diff
@@ -48,9 +47,6 @@
         {{~inject 'carouselArrowAndDotAriaLabel' (lang 'carousel.arrow_and_dot_aria_label')}}
         {{~inject 'carouselActiveDotAriaLabel' (lang 'carousel.active_dot_aria_label')}}
         {{~inject 'carouselContentAnnounceMessage' (lang 'carousel.content_announce_message')}}
-
-        {{!-- Get this loading ASAP --}}
-        <link rel="preload" href="{{cdn 'assets/dist/theme-bundle.main.js'}}" as="script">
```
bookernath, line 53: Also removing this because it becomes redundant; the headers will be parsed before this HTML.

templates/layout/base.html
```diff
@@ -24,14 +24,13 @@
             window.lazySizesConfig = window.lazySizesConfig || {};
             window.lazySizesConfig.loadMode = 1;
         </script>
-        <script async src="{{cdn 'assets/dist/theme-bundle.head_async.js'}}"></script>
+        <script async src="{{cdn 'assets/dist/theme-bundle.head_async.js' resourceHint='preload' as='script'}}"></script>
         
-        {{getFontsCollection font-display='block'}}
+        {{getFontsCollection font-display='block' resourceHint='preload'}}
```
rafa-avila-bc, line 29: this update is not needed. this helper and the `stylesheet` one will, by default, produce preload hints.

templates/layout/base.html
```diff
@@ -24,14 +24,13 @@
...
         
-        <link rel="preload" href="{{cdn 'assets/dist/theme-bundle.font.js'}}" as="script">
-        <script async src="{{cdn 'assets/dist/theme-bundle.font.js'}}"></script>
+        <script async src="{{cdn 'assets/dist/theme-bundle.font.js' resourceHint='preload' as='script'}}"></script>
 
-        {{{stylesheet '/assets/css/theme.css'}}}
+        {{{stylesheet '/assets/css/theme.css' resourceHint='preload'}}}
```
rafa-avila-bc, line 33: this update should not be needed. `stylesheet` helper, by default, produce preload hint.

templates/layout/base.html
```diff
@@ -24,14 +24,13 @@
             window.lazySizesConfig = window.lazySizesConfig || {};
             window.lazySizesConfig.loadMode = 1;
         </script>
-        <script async src="{{cdn 'assets/dist/theme-bundle.head_async.js'}}"></script>
+        <script async src="{{cdn 'assets/dist/theme-bundle.head_async.js' resourceHint='preload' as='script'}}"></script>
```
rafa-avila-bc, line 27: this looks correct to me. `cdn` helper support optimal `as` attribute and `script` is one of its valid values.

bookernath:
All the known issues are fixed; results on WPT mobile test are looking great:

https://www.webpagetest.org/result/221026_AiDcKK_19X5/

@bigcommerce/themes-team this is ready for final review; please merge if it looks good.

BC-krasnoshapka (approved):

Tiggerito:
I've been tracking the effect of early hints on some of the stores using our Tag Rocket app. That is, I'm assuming early hints was enabled around Sept 23rd when we saw some big TTFB improvements. [103 early hint triggers the TTFB moment](https://github.com/GoogleChrome/web-vitals/issues/299#issuecomment-1365028734).

image

As only Chrome actually supports early hints, most of these improved TTFB scores have no effect on the page's performance. This will improve as more browsers support it.

I'm yet to see early hints showing an effect on FCP or LCP.

PR #2266, opened 2022-09-20
Allow quantity of "0" in cart to remove item

sacr3dc0w:
#### What?

Currently, when you type in a quantity 0, you receive an error alert letting you know that's not an accepted value. This PR allows 0 so the customer can remove the line item from the cart without taking a separate action like clicking the minus button or the delete link.

#### Screenshots

**BEFORE**
https://user-images.githubusercontent.com/5056945/191312543-849e56ee-ff1d-4839-9e7a-5916c7f67abf.mov

**AFTER**
https://user-images.githubusercontent.com/5056945/191312063-5be8f468-c113-447e-a194-17c37c551df8.mov


BC-krasnoshapka (approved):
Code looks good. @bc-as what do you think about this feature?

PR #2272, opened 2022-10-13
fix(storefront): CHECKOUT-6665 Add Shipping Info in cart provide a clear indication of fields in error for information that is submitted

bc-NikitaPuzanenko:
#### What?

A description about what this pull request implements and its purpose. Try to be detailed and describe any technical details to simplify the job of the reviewer and the individual on production support.

#### Requirements

#### Tickets / Documentation

- ...

Before:
before
Note that focus stays on the submit button.

After:
after
Focus is moved to the select input element responsible for error.


bc-alexsaiannyi:
Hi @bc-NikitaPuzanenko. I’ve got a concern regarding this change. Currently, we have a flow for announcement  such errors - User can go through all form’s fields using keyboard and screen reader assistance and he/she will get announcement regarding an error on attempt to submit a form.  After you put focus on element User is fixed to this field and can’t go further. So we actually limit User's actions. Is it ok @bc-as?  And also we use this method `announceInputErrorMessage` on 9 forms - authorisation, wishlist, product review, search, contact us etc..

As an option we can add a condition and add focus for specific form (like shipping estimation etc). Another option - to implement this solution specifically in the merchant’s theme. What do you think about this?

bc-NikitaPuzanenko:
@bc-alexsaiannyi Hey thanks for the feedback, I agree that this would limit the users action. Will implement this solution specifically for the merchants theme instead.

PR #2304, opened 2023-01-11
feat(storefront): STRF-10309 Refactor Cornerstone for show_cart_action behavior

jordanarldt:
#### What?
This is a minor refactor of handlebars logic when `show_cart_action` is used. Previously, I had an additional `#or` block for when a customer is logged out and `hide_price_from_guests` is true, but since `show_cart_action` is directly impacted by this in our back-end logic, the `#or` was redundant.

Using `show_cart_action` alone will suffice in this circumstance, as this value will return `false` anyways if a customer is not logged in while `hide_price_from_guests` is true.

#### Requirements

#### Tickets / Documentation

- Related to [STRF-10309](https://bigcommercecloud.atlassian.net/browse/STRF-10309)
- Depends on: https://github.com/bigcommerce/storefront/pull/2092

No screenshots needed, as this PR will not make any difference in the UI or behavior.

[STRF-10309]: https://bigcommercecloud.atlassian.net/browse/STRF-10309?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


templates/components/products/list-item.html
```diff
@@ -95,23 +95,23 @@ <h4 class="listItem-title">
                     {{#if price}}
                         <div class="listItem-price">{{> components/products/price price=price}}</div>
                     {{/if}}
-                    {{#if show_cart_action}}
```
bc-yevhenii-buliuk, line 98: @jordanarldt why do we need to move this condition `{{#if show_cart_action}}...{{/if}} ` outside of `{{#or customer (unless settings.hide_price_from_guests)}}...{{/or}}`?

jordanarldt: Hey @bc-yevhenii-buliuk 
the reason for this is because the `show_cart_action` will already return false in the situation where a customer is guest and prices are restricted, the `{{#or customer (unless settings.hide_price_from_guests)}}...{{/or}}` logic around the `#if show_cart_action` is just redundant.

Wrapping `{{#or customer (unless settings.hide_price_from_guests)}}...{{/or}}` around the `#if price` makes more sense, as the only reason this would be needed is to show the `Login for pricing` message 👍 

In summary, this change in condition will have 0 impact on behavior/functionality, but is slightly more optimized.

BC-krasnoshapka (approved):

templates/components/products/list-item.html
bc-yevhenii-buliuk: If I understand correctly, if the condition `{{#or customer (unless settings.hide_price_from_guests)}}` returns false, then we will see a login message and a comparison button opposite the product.

10309_before

But after transferring the logic, we will see a slightly different display in front of the product, namely the login message, the option selection button (or the add to cart button) and the compare button.

10309_after

Is the appearance of additional option buttons (or an add to cart button) consistent with the design?

jordanarldt: Hey @bc-yevhenii-buliuk that is correct, but in your second screenshot, there would be no `Choose Options`, `Add to Cart`, or `Compare` buttons.

The `show_cart_action` logic will also return `false` in the scenario that `{{#or customer (unless settings.hide_price_from_guests)}}` returns false: 
https://github.com/bigcommerce/bigcommerce/blob/master/includes/classes/class.product.php#L3196-L3202

Essentially, I only moved the logic because now, when `{{#or customer (unless settings.hide_price_from_guests)}}` resolves false, we can also be certain that `show_cart_action` will also return false :)

So there is no reason to wrap two conditions that will return the same thing, since we only need the "Log in for pricing" message where the price would normally be. I hope this makes sense!

bc-yevhenii-buliuk: @jordanarldt As far as I can see, there are no problems in the code itself, and based on this, I give my approval.
But what catches my eye is that in the `templates/components/products/card.html` and `templates/pages/compare.html` files you remove the condition `{{#or customer (unless settings.hide_price_from_guests)}}`
and leave the `{{#if show_cart_action}}` condition, and in the `templates/components/products/list-item.html` file you leave both conditions, moving `{{#if show_cart_action}}` condition outside the `{{#or customer (unless settings .hide_price_from_guests)}}`. This looks a bit inconsistent from my side.
If the backend logic has been changed and the `show_cart_action` key can replace the` {{#or customer (unless settings.hide_price_from_guests)}} `condition, can we do this in the `templates/components/products/list-item.html` file as well?

bc-yevhenii-buliuk (approved):

templates/components/products/list-item.html
jordanarldt: @bc-yevhenii-buliuk Essentially, the reason I've done that is because of the changes I made in storefront service, the `show_cart_action` will now always return `false` during price restriction (user is guest, and prices are hidden for guests), but it also returns `false` under other conditions as well.

The only places we use `{{#or customer (unless settings.hide_price_from_guests)}}` now, are where we need to display the `Login for pricing` message 👍 

I'll resolve this conversation so I can merge. Thanks for the approval!

PR #2357, opened 2023-05-12
Use ISO 8601 DateTime format for datePublished in schema

sacr3dc0w:
#### What?

Noticed this warning came up in the schema visualizer. Simple use of the `{{moment}}` helper resolves it.

> Schema.org datePublished value is not in a valid format or type
> The value needs to be a type or sub-type of:
>
> [DateTime](https://schema.org/DateTime) which needs to comply with the [ISO 8601 DateTime format](http://en.wikipedia.org/wiki/ISO_8601). e.g. '2020-10-24T23:30:05'
> [Date](https://schema.org/Date) which needs to comply with the [ISO 8601 Date format](http://en.wikipedia.org/wiki/ISO_8601). e.g. '2020-10-24'

https://schema.org/Date
https://schema.org/DateTime

**BEFORE**
Screenshot 2023-05-12 at 10 21 09 AM

**AFTER**
Screenshot 2023-05-12 at 10 24 18 AM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added css classes for ApplePay Button [[#2344]](https://github.com/bigcommerce/cornerstone/pull/2344)
 - Added styling config for the Bolt smart payment button [[#2356]](https://github.com/bigcommerce/cornerstone/pull/2356)
 - Remove default whitespace from multiline input [#2355](https://github.com/bigcommerce/cornerstone/pull/2355)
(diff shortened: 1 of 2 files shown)


Tiggerito:
moment can't cope with all possible date formats that can be used, including the default one. It would be nice if a computer-readable format is also provided in the first place. I raised this as an idea in 2017:

https://support.bigcommerce.com/s/ideas#0871B000000ivWuQAI

And this year:

https://support.bigcommerce.com/s/ideas#0874O000000LeuzQAC

Nice to see you are using my Classy Schema tool for validating your schema 😁

PR #2370, opened 2023-06-29
fix(storefront): BCTHEME-1378 fix add product to cart on iphone x (iphone version 11)

yurytut1993:
#### What?

This pr is a fix for correct processing of invalid formData in iOS Safari 11.2. Two issues were covered here:
1. When formData is invalid but exists - polifill do not work, so additional checks were added to detect such browsers, like iOS 11.2 Safari
2. FormData polyfill should run after whatwg-fetch to correctly update fetch that should convert formData like object into blob. That is the reason why correctly collected formData object was unclear for server. So formData polyfill should run after whatwg-fetch declared in stencil-utils. It goes with main bundle so after main loaded formData polyfill can be added to the browser 

#### Tickets / Documentation

- [BCTHEME-1678](https://bigcommercecloud.atlassian.net/browse/BCTHEME-1378)

https://github.com/bigcommerce/cornerstone/assets/66325265/48600629-6a2a-41b4-8704-b1f0b04cd281

[BCTHEME-1678]: https://bigcommercecloud.atlassian.net/browse/BCTHEME-1678?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/templates/layout/base.html
+++ b/templates/layout/base.html
@@ -60,6 +60,23 @@
             {{!-- Exported in app.js --}}
             function onThemeBundleMain() {
                 window.stencilBootstrap("{{page_type}}", {{jsContext}}).load();
+
+                function browserSupportsFormData() {
+                    return typeof FormData !== 'undefined' 
+                        && !!FormData.prototype.keys;
+                }
+                function loadFormDataPolyfillScript(src) {
+                    var formDataPolyfillScript = document.createElement('script');
+                    formDataPolyfillScript.src = src;
+                    formDataPolyfillScript.onerror = function () {
+                        console.error('Failed to load formData polyfill script ' + src);
+                    };
+                    document.body.appendChild(formDataPolyfillScript);
+                }
+
+                if (!browserSupportsFormData()) {
+                    loadFormDataPolyfillScript('{{cdn 'assets/dist/theme-bundle.polyfill_form_data.js'}}');
+                }
             }
         </script>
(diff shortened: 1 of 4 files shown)


templates/layout/base.html
```diff
@@ -60,6 +60,27 @@
...
+                    };
+                    document.body.appendChild(formDataPolyfillScript);
+                }
+
+                if (!browserSupportsFormData()) {
+                    console.log('formData polyfill works');
+                    loadFormDataPolyfillScript('{{cdn 'assets/dist/theme-bundle.polyfill_form_data.js'}}');
+                } else {
+                    console.log('formData polyfill NOT work');
+                }
```
BC-krasnoshapka, line 83: @yurytut1993  does it make sense to have this in the separate js bundle, same as polyfills.js? I know it's just  ~20 lines, but with separate bundle it will be cached in browser and not loaded on every request as html response.

yurytut1993: We can not guarantee that polyfills.js will be ready before main bundle. 99% that it will be, but still. This is the reason I put it here in base. When I remove extra consoles it will be less lines 😃

bc-alexsaiannyi (approved):
I can't find better way of implementation as well. 
Just a quick notice - such amount of changes, is it expected?

yurytut1993:
> I can't find better way of implementation as well. Just a quick notice - such amount of changes, is it expected?
I believe lock file changes should not be added in this pr, but its just for not to forget to discuss

BC-krasnoshapka (approved):

bc-azvierieva (approved):

PR #2395, opened 2023-10-11
Improve German plural strings

2ndkauboy:
#### What?

In German, when referring to a single item, we use "ein" ("one") instead of using the number 1 before the term.

Example: "ein Produkt" instead of "1 Produkt"

This PR will change all translations to a more natural way of writing those strings in German.

### DIFF
--- a/lang/de.json
+++ b/lang/de.json
@@ -10,7 +10,8 @@
         "info": "Info",
         "categories": "Kategorien",
         "call_us": "Rufen Sie uns an unter {phone_number}",
-        "powered_by": "Bereitgestellt von"
+        "powered_by": "Bereitgestellt von",
+        "test": "Hello, Berlin!"
     },
     "home": {
         "heading": "Startseite"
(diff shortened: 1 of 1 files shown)


2ndkauboy:
If you decide not to go forward with this PR, please update at least this line to a capitalized word: https://github.com/bigcommerce/cornerstone/pull/2395/files#diff-b856eeca8165aed4012d1d98ca6e927f20366c9c1272932ad9d8004358c1f38dR343

BC-krasnoshapka:
hi @2ndkauboy , since BigCommerce provides translations for all officially supported languages, we have internal translation flow that gets them from translation agency and pushes to github, thus usually we don't accepts PRs to lang files as they could be overwritten later. Instead we'll pass your suggestion to language experts and hopefully will get it back as an update. Thanks.

2ndkauboy:
@BC-krasnoshapka all right! If that's your workflow, please feel free to send my suggestions to that external agency.

PR #2396, opened 2023-10-11
Load other font weights and styles for the body-font

2ndkauboy:
#### What?

Using a font name like `Google_Karla_400` for `body-font`, it would only load the 400 font weight. When using `Google_Karla_700` for another property, it would then load both the 400 and 700 fon weights. But since in CSS you could make something bold or italic, the font weight and/or style for this text might be missing, as it was not loaded automatically by using it in another property.

This PR will make sure that the most common font weights and font styles will be loaded for the body font.

#### Screenshots

The Karla font with `font-weight: 400` rendered in bold - looks squashed:
Karla-Font

The Karla font with `font-weight: 700` rendered in bold - looks perfect:
Karla-Font-700

### DIFF
--- a/config.json
+++ b/config.json
@@ -103,7 +103,7 @@
     "default_image_brand": "img/BrandDefault.gif",
     "default_image_product": "img/ProductDefault.gif",
     "default_image_gift_certificate": "img/GiftCertificate.png",
-    "body-font": "Google_Karla_400",
+    "body-font": "Google_Karla_400,400i,700,700i",
     "headings-font": "Google_Montserrat_400",
     "fontSize-root": 14,
     "fontSize-h1": 28,


2ndkauboy:
I wasn't able to (easily) find out, if the Montserrat font was also used in `italic` somewhere in CSS. But it might be a good idea to also load it in different styles, like this:
```
"headings-font": "Google_Montserrat_400,400i,500,500i,700,700i"
```

bc-vlad-dlogush (approved):

yurytut1993 (approved):

BC-krasnoshapka:
hi @2ndkauboy , looks reasonable. can you please update CHANGELOG file also? thanks

bc-alexsaiannyi (approved):
lgtm

BC-krasnoshapka (approved):

mattcoy-arcticleaf:
This is not a valid value (since it's not an option in the schema) and this will slow down the PageSpeed, potentially for no reason.

I would recommend reconsidering this change.

@BC-krasnoshapka

bc-alexsaiannyi:
Hi @mattcoy-arcticleaf. Thank you for bringing that up. Obviously, you're right! What do you think if we can follow up the original contributor and extend `Google_Karla_400`  with 700 version and delete 2 other italic versions (with respectful  changes in `schema`)? It will probably still slow down the PageSpeed a bit, but not as much as having 4 fonts.

PR #2484, opened 2024-09-17
Draft PR - Example for Supporting Missing Prices in Stencil Context Data

bc-jz:
#### What?
This is an example of some template adjustments that will allow the PDP to support new pricing behavior where `price` can be excluded from variant data if not defined in an "Inclusive" or "Exclusive" price list. These changes specifically address issues brought up in BCTHEME-1941, namely:

- If PDP page loads without a `price` then it breaks the HTML of the page and a price will never be shown even when selecting variants that have a price set
- on PDP when selecting a variant that does not have a price set, we do not hide the price on the page but instead keep showing the previous selections price

#### Requirements

#### Tickets / Documentation

- ...

Example Image

### DIFF
--- a/assets/js/theme/common/product-details-base.js
+++ b/assets/js/theme/common/product-details-base.js
@@ -15,13 +15,13 @@ const optionsTypesMap = {
     PRODUCT_LIST: 'product-list',
 };
 
-export function optionChangeDecorator(areDefaultOtionsSet) {
+export function optionChangeDecorator(areDefaultOptionsSet) {
     return (err, response) => {
         const attributesData = response.data || {};
         const attributesContent = response.content || {};
 
         this.updateProductAttributes(attributesData);
-        if (areDefaultOtionsSet) {
+        if (areDefaultOptionsSet) {
             this.updateView(attributesData, attributesContent);
         } else {
             this.updateDefaultAttributesForOOS(attributesData);
@@ -145,8 +145,14 @@ export default class ProductDetailsBase {
      */
     getViewModel($scope) {
         return {
-            $priceWithTax: $('[data-product-price-with-tax]', $scope),
-            $priceWithoutTax: $('[data-product-price-without-tax]', $scope),
+            priceWithTax: {
+            	$div: $('div.price--withTax', $scope),
+            	$span: $('[data-product-price-with-tax]', $scope)
+            },
+            priceWithoutTax: {
+            	$div: $('div.price--withoutTax', $scope),
+            	$span: $('[data-product-price-without-tax]', $scope)
+            },
             rrpWithTax: {
                 $div: $('.rrp-price--withTax', $scope),
                 $span: $('[data-product-rrp-with-tax]', $scope),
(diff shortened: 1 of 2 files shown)


bc-yevhenii-buliuk:
@bc-jz thanks for the suggested implementation, it looks good. Also, we need to consider the case when none of the product options are selected by default. Do you think we can similarly add an additional `price--withoutTax / price--withTax` classes for `price-range.html` that you use for `price.html`? 

https://github.com/user-attachments/assets/d101cb4a-684d-4cb7-8b04-473897ae9bd9

bc-jz:
> @bc-jz thanks for the suggested implementation, it looks good. Also, we need to consider the case when none of the product options are selected by default. Do you think we can similarly add an additional `price--withoutTax / price--withTax` classes for `price-range.html` that you use for `price.html`?

Yes, I think the adjustments we make to `price-range.html` will be comparable. We mainly just need to make sure that the `div` that wraps around the "price now" section is given the same `price--withTax` or `price--withoutTax` class that was added into `price.html`, for example:
https://github.com/bigcommerce/cornerstone/pull/2484/files#diff-377f95bae5a6b0973d0ab2f44655a0bae14e006e651ee3c48e96b1e00f73d59aR32

We probably don't need to worry about adding the logic for `display: none` since if there is a price range provided it should always have a price. The output of `price-range.html` just needs to be comparable to `price.html` and we should validate they both play appropriately with the changes to the JS.

----

Side note, what do you think about removing the `price--withTax` and `price--withoutTax` classes from their existing placement on the `span` elements wrapping the price to reduce the chance to accidentally target them? For example here:
https://github.com/bigcommerce/cornerstone/blob/98d2e4032607621303c2b385ac494cc16b7aa13d/templates/components/products/price.html#L42

I wrote my JS to specifically target the `div` elements and was not sure if those classes on the `span` are used anywhere else so left them in place but maybe that could be cleaned up.

bc-yevhenii-buliuk:
`Side note, what do you think about removing the price--withTax and price--withoutTax classes from their existing placement on the span elements wrapping the price to reduce the chance to accidentally target them?`

@bc-jz I agree with you and I also paid attention to this and I didn't see anywhere in the styles or JS logic where the `price--withoutTax / price--withTax` classes are used for the `span` element, so I think we can remove these unused classes. As a result, we won't need to specify this in the `getViewModel` method on the [`div`](https://github.com/bigcommerce/cornerstone/pull/2484/files#diff-129774752f85c699dae74c4d8cdbe34d3f2bfad7b885f5d33fa9e0da46e84f5cR149) element.

Indeed, at the first rendering of the price range, the price always exists and is taken only from the stencil object without an additional AJAX request for price clarification. Therefore, it is probably [worth adding](https://github.com/bigcommerce/cornerstone/pull/2484/files#diff-377f95bae5a6b0973d0ab2f44655a0bae14e006e651ee3c48e96b1e00f73d59aR5) the `price--withoutTax / price--withTax` classes to the corresponding `div` elements to display/hide the price, just like in `price.html`. At the same time, we can also remove unused classes from `span` in `price-range.html`.

bc-jz:
Superseded by https://github.com/bigcommerce/cornerstone/pull/2486

PR #2526, opened 2024-12-19
fix(storefront): remove escaping of = symbol for <head>

bc-yevhenii-buliuk:
#### What?

This PR removes escaping of `=` symbol for `<head>`, namely for meta prev/next URLs on the Category Page and favicon.

#### Tickets / Documentation

- [SD-10779](https://bigcommercecloud.atlassian.net/browse/SD-10779)

#### Screenshots (if appropriate)

**before:**

https://github.com/user-attachments/assets/f79e6871-e9ff-4175-9a27-84d386861e10

**after:**

https://github.com/user-attachments/assets/f6776a66-6b1e-4521-85c3-bee660536f40

[SD-10779]: https://bigcommercecloud.atlassian.net/browse/SD-10779?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2528, opened 2024-12-25
Remove escaping of "=" symbol for <head> for blog and brand

sacr3dc0w:
#### What?

While merging #2526 into our codebase, I noticed that `blog.html` and `brand.html` didn't get the update that removes escaping of = symbol for `<head>` for meta prev/next URLs.

#### Screenshots

We don't utilize the blog functionality, so I can't provide BEFORE / AFTER screenshots for that change, but here are screenshots for changes to `brand.html`

**BEFORE**
Screen Shot 2024-12-24 at 4 09 47 PM

**AFTER**
Screen Shot 2024-12-24 at 4 09 23 PM


rtalvarez (approved):

bc-alexsaiannyi (approved):
Thanks👍

PR #2554, opened 2025-06-13
fix(storefront): SD-11213 restore indentation and quote rules to match BC Sass Style Guide

bc-yevhenii-buliuk:
## What?
This PR restores indentation and quote styles to follow company [Sass Style Guide](https://github.com/bigcommerce/sass-style-guide).

This PR implements the official Stylelint migration recommendation by using [@stylistic/stylelint-plugin](https://www.npmjs.com/package/@stylistic/stylelint-plugin), which restores deprecated stylistic rules as a plugin.
It allows us to enforce indentation, quotes, and other formatting rules directly through Stylelint without relying on Prettier or conflicting legacy configs.

##
Modern versions of Stylelint (v16+) removed core stylistic rules like `indentation` and `string-quotes`. These rules were previously handled directly or via legacy config `stylelint-config-sass-guidelines`, which is no longer compatible with the new architecture. As a result, enforcing basic code style (4-space indentation, double quotes) became impossible without errors like `Unknown rule`.

#### Tickets / Documentation

[SD-11213](https://bigcommercecloud.atlassian.net/browse/SD-11213)

**BEFORE:**
```
assets/scss/common/_aria.scss
  1:1  ✖  Unknown rule string-quotes     string-quotes
  1:1  ✖  Unknown rule scss/indentation  scss/indentation
```
11213_CI_err_before

##
**AFTER:**
_CI error:_
11213_bug_in_CI

_fixed CI Stylelint error by reproducing and resolving it locally._
11213_fix

[SD-11213]: https://bigcommercecloud.atlassian.net/browse/SD-11213?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bc-alexsaiannyi (approved):
👍

package.json
```diff
@@ -56,8 +57,8 @@
     "lighthouse": "^12.1.0",
     "load-grunt-config": "^3.0.1",
     "lodash-webpack-plugin": "^0.11.6",
+    "postcss-scss": "^4.0.9",
```
jkanive, line 60: Why do we need this package? Had we removed it before?

bc-yevhenii-buliuk: @jkanive We need `"postcss-scss"` since our styles are written in SCSS, which has features like `//` comments and `$variables` that regular CSS doesn't support. Without this parser, Stylelint doesn't understand the syntax and throws errors like
```
assets/scss/checkout.scss
  1:1  ✖  Unknown word //  CssSyntaxError.
```
So it helps Stylelint read our SCSS files properly.
We didn’t need `postcss-scss` before because we were using the `stylelint-config-sass-guidelines` config, which internally handled the SCSS parsing.

PR #2557, opened 2025-07-03
refactor(orders): ORDERS-7036 refactor orders/invoice.html pickup_details section to better align with existing pickup_address output

matt-evangelidis:
#### What?

This PR changes the `pickup_address` fields accessed and displayed in `orders/invoice.html` to match the fields used in `orders/details.html`. 

This change, along with another PR solves an issue where the pickup address would not display on the order invoice or order details pages, if an order with a pickup consignment had only one pickup consignment (which is technically all of them based on current platform limitations).

This is also being done as part of simplifying the backend code that produces the object for `order.pickup_address`, as there is no reason to use two different output formats for the same data.

#### Requirements

#### Tickets / Documentation

- [ORDERS-7036](https://bigcommercecloud.atlassian.net/browse/ORDERS-7036)
- https://github.com/bigcommerce/bigcommerce/pull/63822

<details>
<summary>Invoice Page</summary>

Without UI or backend changes
image

With backend changes only
image

With UI and backend changes
image

</details>

### DIFF
--- a/templates/pages/account/orders/invoice.html
+++ b/templates/pages/account/orders/invoice.html
@@ -61,13 +61,15 @@
                     <div class="ShippingAddress">
                         <h3>{{lang 'account.orders.details.pickup_details'}}</h3>
                         <ul>
-                            <li><strong>{{order.pickup_address.line1}}</strong></li>
-                            <li>{{order.pickup_address.line2}}</li>
-                            <li>{{order.pickup_address.city}}, {{order.pickup_address.state}} {{order.pickup_address.zip}}</li>
-                            <li>{{order.pickup_address.country}}</li>
+                            {{#each order.pickup_address}}
+                            <li><strong>{{location_line_1}}</strong></li>
+                            <li>{{location_line_2}}</li>
+                            <li>{{location_city}}, {{location_state}} {{location_zip}}</li>
+                            <li>{{location_country_name}}</li>
                             <li>&nbsp;</li>
-                            <li>{{lang 'invoice.phone' number=order.pickup_address.phone}}</li>
-                            {{#if order.pickup_address.email}}<li>{{lang 'invoice.email' email=order.pickup_address.email}}</li>{{/if}}
+                            <li>{{lang 'invoice.phone' number=location_phone}}</li>
+                            {{#if location_email}}<li>{{lang 'invoice.email' email=location_email}}</li>{{/if}}
+                            {{/each}}
                         </ul>
                     </div>
                 {{/unless}}


donald-nguyen-bc (approved):
👍

bc-vincent-zhao (commented):
For context: I think the code look like this is due to the MLI phasing, strictly speaking, we've only delivered the "single pickup" phase, so you would see assumption made throughout the codebase that there's one pickup address to simplify things. 

Some engineers find it more future proof to handle multiple pickup case to start with so we don't have to come back and revisit this, which led to the current inconsistency.

For rolling out this change:

This change depends on a new data structure - a list of pickup address instead of a single one, so for it to work, the new data structure needs to be always there, so given we don't have experiment control in cornerstone, the best way to introduce this change is to make the **additional** data available from bcapp side.

So for the default experience for any new stores or existing store, they will benefit from this change that now will support multiple pickup addresses.

However, for existing stores that might have customised this template, for backward compatible reason, I think we still need to expose the single pickup object format.

So we should really have 2 data available at the same time:
- order.pickup_address -- existing 1st pickup address
- order.pickup_addresses -- new additive change that expose the collection

jkanive:
@matt-evangelidis Could you point me to the different in `before` and `after` screenshots? I might be missing something.

bc-alexsaiannyi:
Hi @matt-evangelidis, changes themselves look fine. Just small comments:

1.  Please add item to `changelog` regarding your change 💅
2. I like Vincent idea about backward compatibility. Can you expand where we are with this? and do I understand this correctly that new data `order.pickup_addresses` covered with experiment `ORDERS-7036.SF_order_details_page_unify_pickup_address` and if not we back to `order.pickup_address`?

matt-evangelidis:
> Hi @matt-evangelidis, changes themselves look fine. Just small comments:
> 
> 1. Please add item to `changelog` regarding your change 💅
> 2. I like Vincent idea about backward compatibility. Can you expand where we are with this? and do I understand this correctly that new data `order.pickup_addresses` covered with experiment `ORDERS-7036.SF_order_details_page_unify_pickup_address` and if not we back to `order.pickup_address`?

Hi @bc-alexsaiannyi, 

I've added a `changelog` entry under `Draft`.

To explain the backwards compatibility concept, the existing payloads that the Order Details and Order Invoice pages get from BCApp contains a `pickup_address` field. Previously, this object would have different keys depending on the [condition evaluated here](https://github.com/bigcommerce/bigcommerce/blob/0b0633112280376ea11eca30a976f634024378bb/includes/resources/AccountOrderDetails.php#L614), so depending on the order, `pickup_address` would have the correct keys for one page but not the other.

To solve this, we've added a net-new `pickup_addresses` field which has no conditional behaviour and will have a guaranteed set of keys that both pages can safely read from (we're no longer covering this with an experiment, given this is completely net-new behaviour with no existing usage). To keep previous versions of Cornerstone working, we'll keep the `pickup_address` field around and with its current behaviour.

matt-evangelidis:
Would appreciate any reviews from @bigcommerce/team-store-design

bc-alexsaiannyi:
> ... we'll keep the `pickup_address` field around and with its current behaviour.

@matt-evangelidis, thanks for giving more context on this. It sounds good to me.

bc-alexsaiannyi (approved):
lgtm

PR #2558, opened 2025-07-07
feat(checkout): Prerender cart checkout

richapatel1510:
#### What?

Add speculation rules for prerendering cart and checkout page. This will prerender cart page and checkout page (based on version to bust the cache) and result in slow rendering times for checkout page. 
Faster storefront experience leads to a better checkout experience which give our merchants with a better chance conversion.

**Note:** This depends on bcapp PR to pass the version for cart update, apply coupon and apply gift certificate endpoints.
https://github.com/bigcommerce/bigcommerce/pull/64102

#### Requirements

#### Tickets / Documentation

- [CHECKOUT-9357](https://bigcommercecloud.atlassian.net/browse/CHECKOUT-9357)
- [PROJECT-6882](https://bigcommercecloud.atlassian.net/browse/PROJECT-6882)

**Experiment OFF**

https://github.com/user-attachments/assets/3576eacb-760a-4fb2-90e2-57780fc2e1dc

**Experiment ON**

https://github.com/user-attachments/assets/cb1d7662-e1ac-439a-92b4-ce95156c5efd

[CHECKOUT-9357]: https://bigcommercecloud.atlassian.net/browse/CHECKOUT-9357?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PROJECT-6882]: https://bigcommercecloud.atlassian.net/browse/PROJECT-6882?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/js/theme/cart.js
+++ b/assets/js/theme/cart.js
@@ -60,8 +60,10 @@ export default class Cart extends PageManager {
                 // if the quantity is changed "1" from "0", we have to remove the row.
                 const remove = (newQty === 0);
 
-                const checkoutUrl = urlUtils.replaceParams(this.$cartPrimaryCheckoutAction.attr('href'), { version: response.data.version });
-                this.$cartPrimaryCheckoutAction.attr('href', checkoutUrl);
+                if (this.context.prerenderCheckoutEnabled) {
+                    const checkoutUrl = urlUtils.replaceParams(this.$cartPrimaryCheckoutAction.attr('href'), { version: response.data.version });
+                    this.$cartPrimaryCheckoutAction.attr('href', checkoutUrl);
+                }
 
                 this.refreshContent(remove);
             } else {
@@ -103,8 +105,10 @@ export default class Cart extends PageManager {
                 // if the quantity is changed "1" from "0", we have to remove the row.
                 const remove = (newQty === 0);
 
-                const checkoutUrl = urlUtils.replaceParams(this.$cartPrimaryCheckoutAction.attr('href'), { version: response.data.version });
-                this.$cartPrimaryCheckoutAction.attr('href', checkoutUrl);
+                if (this.context.prerenderCheckoutEnabled) {
+                    const checkoutUrl = urlUtils.replaceParams(this.$cartPrimaryCheckoutAction.attr('href'), { version: response.data.version });
+                    this.$cartPrimaryCheckoutAction.attr('href', checkoutUrl);
+                }
 
                 this.refreshContent(remove);
             } else {
@@ -119,8 +123,10 @@ export default class Cart extends PageManager {
         this.$overlay.show();
         utils.api.cart.itemRemove(itemId, (err, response) => {
             if (response.data.status === 'succeed') {
-                const checkoutUrl = urlUtils.replaceParams(this.$cartPrimaryCheckoutAction.attr('href'), { version: response.data.version });
-                this.$cartPrimaryCheckoutAction.attr('href', checkoutUrl);
+                if (this.context.prerenderCheckoutEnabled) {
+                    const checkoutUrl = urlUtils.replaceParams(this.$cartPrimaryCheckoutAction.attr('href'), { version: response.data.version });
+                    this.$cartPrimaryCheckoutAction.attr('href', checkoutUrl);
+                }
 
                 this.refreshContent(true);
             } else {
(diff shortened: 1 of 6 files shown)


templates/pages/checkout.html
```diff
@@ -32,6 +32,20 @@ <h2 class="checkoutHeader-heading">
 
 {{{ footer.scripts }}}
 
+<!-- Script to remove the version parameter from url without reloading the page -->
```
bc-peng, line 35: Just for the record, could you please explain why do we have a `version` here?
I can see the `speculationrules` object is matching `&version=*`, which indicates the version value does not matter.

bc-peng (approved):

templates/pages/checkout.html
richapatel1510: Version is basically to bust the cache. Whenever there is update in the cart we get the updated cart version from API and append it to the checkout url, which will now refetch the latest version of checkout. If we don't do that, it will fetch the checkout on very first load and it may then result in outdated information when we go to checkout, hence the implementation.

And this script is to remove that version from url parameters when we actually redirect to checkout since we do not need that information there.

richapatel1510: > I can see the speculationrules object is matching &version=*, which indicates the version value does not matter.

It doesn't actually indicate the version value does not matter, it will trigger speculation rule again for every new version when appended to checkout url.

bc-peng: Thanks for the very clear explaination.

Tiggerito: See if I understand.

Version is like a hash of the cart contents which changes as the cart changes. It's a way to ignore rendered caches of the checkout that have old cart information.

The version removal is just cleaning up the address bar on the checkout page.

animesh1987: Hey @Tiggerito you are correct. We added `version` field to checkout API to as part of implementing optimistic locking. You can read more about version here https://developer.bigcommerce.com/docs/rest-storefront/checkouts.

> The current version of the checkout increments with each successful update. You can use it to enable optimistic concurrency control for subsequent updates.

assets/js/theme/common/product-details.js
```diff
@@ -519,6 +519,36 @@ export default class ProductDetails extends ProductDetailsBase {
 
             modal.updateContent(response);
 
+            // Check if prerender is enabled from the template data
+            const isPrerenderCheckoutEnabled = $('.previewCart').find('[data-checkout-prerender-enabled="true"]').length > 0;
+
+            // Add speculative loading after cart template is loaded
+            if (isPrerenderCheckoutEnabled) {
```
davidchin, line 526: I'm not sure if the browser would throw any error if we add speculation rules for browsers that don't support them. Perhaps we need to add a feature [check](https://developer.mozilla.org/en-US/docs/Web/API/Speculation_Rules_API#speculation_rules_api_feature_detection) before adding the rules? e.g.:

```js
if (HTMLScriptElement.supports && HTMLScriptElement.supports('speculationrules')) {
  // ...
}
```

assets/js/theme/cart.js
```diff
@@ -58,6 +60,11 @@ export default class Cart extends PageManager {
                 // if the quantity is changed "1" from "0", we have to remove the row.
                 const remove = (newQty === 0);
 
+                if (this.context.prerenderCheckoutEnabled) {
+                    const checkoutUrl = urlUtils.replaceParams(this.$cartPrimaryCheckoutAction.attr('href'), { version: response.data.version });
+                    this.$cartPrimaryCheckoutAction.attr('href', checkoutUrl);
```
davidchin, line 65: We should be able to expire the pre-rendered cache whenever the shopper updates their cart on the same page. However, if the shopper has two tabs open, and updates their cart in the first tab before switching to the second tab to proceed to checkout, they will see an outdated cart. This happens because the links (hrefs) in the second tab still contain the old version number. You can find similar examples listed under "[unsafe](https://developer.mozilla.org/en-US/docs/Web/API/Speculation_Rules_API#server-rendered_varying_state) speculative loading conditions."

Looking through the docs a bit more, I wonder instead of trying to expire the cache on the client side using version numbers, it might be easier to just return Clear-Site-Data: "prerenderCache" in the [response header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Clear-Site-Data#clearing_speculations) whenever a cart mutation request is made. We don’t have as much control over what gets cleared, but it might be a safer and simpler way to handle it. Another benefit is that if we don’t need to handle cache expiration on the client side, we could inject speculation rules on the server side - either through the head.scripts Stencil variable or via a response header. This would allow us to apply the rules across all Stencil stores, not just a specific version of Cornerstone. 🤔

PR #2579, opened 2025-12-04
feat(checkout): CHECKOUT-9692 show the display name and coupon code under line item

richapatel1510:
#### What?

- Show the display name and code under the product so it is clear to shoppers that the product coupon was applied successfully. Attach coupon to free items as well.
- Behind the experiment `CHECKOUT-9674.multi_coupon_cart_checkout`

#### Requirements

#### Tickets / Documentation

- [CHECKOUT-9692](https://bigcommercecloud.atlassian.net/browse/CHECKOUT-9692)

**BEFORE - Experiment OFF**
Screenshot 2025-12-04 at 5 00 48 pm

**AFTER - Experiment ON**
Screenshot 2025-12-04 at 4 48 51 pm

[CHECKOUT-9692]: https://bigcommercecloud.atlassian.net/browse/CHECKOUT-9692?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Update Cornerstone documentation url [#2575](https://github.com/bigcommerce/cornerstone/pull/2575)
 - Fix keyboard navigation on the swatch options [#2576](https://github.com/bigcommerce/cornerstone/pull/2576)
+- CHECKOUT-9688 Introduce new coupon management section on cart page [#2578](https://github.com/bigcommerce/cornerstone/pull/2578)
+- CHECKOUT-9692 show the display name and coupon code under line item [#2579](https://github.com/bigcommerce/cornerstone/pull/2579)
 
 ## 6.17.0 (10-01-2025)
 - Add net-new "order.pickup_addresses" to unify objects used on Order Details and Order Invoice pages [#2557](https://github.com/bigcommerce/cornerstone/pull/2557)


templates/components/cart/coupons.html
```diff
@@ -19,7 +19,15 @@
...
-                        <span class="cart-coupon-name">{{display_name}} ({{code}})</span>
+                        {{#if display_name}}
+                            {{#compare display_name "===" "Coupon promotion"}}
+                                <span class="cart-coupon-name">{{code}}</span>
+                            {{else}}
+                                <span class="cart-coupon-name">{{display_name}} ({{code}})</span>
+                            {{/compare}}
+                        {{else}}
+                            <span class="cart-coupon-name">{{code}}</span>
+                        {{/if}}
```
animesh1987: can we try something like this


```suggestion
        {{#if (and display_name (compare display_name "!==" "Coupon promotion"))}}
    <span class="cart-coupon-name">{{display_name}} ({{code}})</span>
{{else}}
    <span class="cart-coupon-name">{{code}}</span>
{{/if}}
```

richapatel1510: I tried this, somehow it doesn't work using `and` and `compare` in same block. It ends up throwing `TypeError: options.inverse is not a function` error.

richapatel1510: Found another way to optimise it https://github.com/bigcommerce/cornerstone/pull/2579/commits/0c9e9a3a2cfff49795415e93459dd886e6a55a87

animesh1987 (approved):

assets/scss/components/stencil/cart/_cart.scss
```diff
@@ -517,6 +517,10 @@ $card-preview-zoom-bottom-offset:       6rem;
     }
 }
 
+.cart-item-discount, .cart-item-value.price--discounted {
+    color: stencilColor("color-greyLight");
```
bc-alexsaiannyi, line 521: hi @richapatel1510 👋 I might be missing something, but are we considering the fact that the theme has three style variations? Will this colour still match  rest of the theme if the user switches to a different variation? For example:
image

If not, maybe we can take an example from `cart-total-label`. Here are  [scss](https://github.com/bigcommerce/cornerstone/blob/master/assets/scss/components/stencil/cart/_cart.scss#L368) settings and here is a [config](https://github.com/bigcommerce/cornerstone/blob/master/config.json#L142) where it applies across all versions.

PR #2649, opened 2026-05-04
feat(orders): ORDERS-7706 add new return page layout

jpajar:
#### What?

A description about what this pull request implements and its purpose. Try to be detailed and describe any technical details to simplify the job of the reviewer and the individual on production support.

This PR introduces the layout as per designs in ORDERS-7706 for the shopper facing new returns page.

This PR currently uses the same shape as the `returns-ui` merchant facing repo and has hardcoded objects until we have decided on the shape and passing it to the stencil context.

This PR currently does not call any backend when submitting the form, and has conditional rendering, if conditional rendering is not the pattern or we do not want to introduce this, we may have to figure out how to shape the backend response so that the front-end can remail purely handlebars / presentational.

#### Requirements

#### Tickets / Documentation

- [ORDERS-7706(https://bigcommercecloud.atlassian.net/browse/ORDERS-7706
- ...

https://github.com/user-attachments/assets/11f41d62-137f-466c-a78a-e1c3abf14cec

### DIFF
--- a/assets/js/theme/add-return-new.js
+++ b/assets/js/theme/add-return-new.js
@@ -1,0 +1,245 @@
+import { escape } from 'lodash';
+import PageManager from './page-manager';
+
+export default class AddReturnNew extends PageManager {
+    onReady() {
+        const $form = $('[data-new-return-form]');
+        if (!$form.length) return;
+
+        // ---------------------------------------------------------------------------
+        // Hardcoded page data which mirrors the TypeScript returns-ui interface.
+        // Replace with Stencil context when it is available.
+        // ---------------------------------------------------------------------------
+        this.pageData = {
+            order: {
+                id: '101',
+                datePlaced: new Date('2024-01-01'),
+                status: 'Completed',
+                currency: 'AUD',
+                isTaxInclusive: true,
+                shipping: {
+                    address: {
+                        fullName: 'Jane Smith',
+                        address1: '1-3 Smail Street',
+                        city: 'Ultimo',
+                        state: 'NSW',
+                        zip: '2007',
+                        country: 'Australia',
+                    },
+                    method: 'Australia Post',
+                    dateShipped: 'May 15, 2024',
+                    trackingNumber: '7E27315406641',
+                },
+                items: [
+                    {
+                        id: 'item-1',
+                        name: 'Product Name',
+                        variant: 'Blue / Black / Green',
+                        sku: 'SKU-001',
+                        quantity: 1,
+                        returnableQuantity: 1,
+                        thumbnailUrl: '',
+                        totalIncTax: 123.99,
+                        totalExTax: 110.71,
+                    },
+                    {
+                        id: 'item-2',
+                        name: 'Product Name',
+                        variant: 'Blue / Black / Green',
+                        sku: 'SKU-002',
+                        quantity: 1,
+                        returnableQuantity: 1,
+                        thumbnailUrl: '',
+                        totalIncTax: 123.99,
+                        totalExTax: 110.71,
+                    },
+                    {
+                        id: 'item-3',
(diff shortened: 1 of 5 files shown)


assets/js/theme/add-return-new.js
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    escHtml(str) {
+        return String(str)
+            .replace(/&/g, '&amp;')
+            .replace(/</g, '&lt;')
+            .replace(/>/g, '&gt;')
+            .replace(/"/g, '&quot;')
+            .replace(/'/g, '&#39;');
+    }
```
matt-evangelidis: Is there anything pre-existing in cornerstone we can make use of? Needing to hand-roll any kind of HTML escaping tends to be a very bad time.

package-lock.json
```diff
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "bigcommerce-cornerstone",
-      "version": "6.19.0",
+      "version": "6.19.1",
```
matt-evangelidis, line 9: What actually caused the version bump here?

assets/js/theme/add-return-new.js
jpajar: I'll have another look

package-lock.json
jpajar: I actually do not know 😅 

The only commands I've run are 
`stencil init`
`stencil start`
`npm install`

assets/js/theme/add-return-new.js
jpajar: I think as a stopgap, we can use `lodash escape` as `lodash` is already used in the repo. We don't do HTML escaping as everything is done via handlebars but atm I'm building the HTML manually via JS as I don't have the Stencil context yet. Ideally we do the `{{#each}}` approach to render the HTML

assets/js/theme/add-return-new.js
```diff
@@ -0,0 +1,245 @@
+import { escape } from 'lodash';
+import PageManager from './page-manager';
+
+export default class AddReturnNew extends PageManager {
+    onReady() {
+        const $form = $('[data-new-return-form]');
+        if (!$form.length) return;
```
matt-evangelidis, line 7: To clarify about the JS here, the only thing that really gets run is `onReady`, and the rest is more or less private functions that we use to manipulate what we stored in `pageData`?

assets/js/theme/add-return-new.js
```diff
@@ -0,0 +1,245 @@
...
+
+        // ---------------------------------------------------------------------------
+        // Hardcoded page data which mirrors the TypeScript returns-ui interface.
+        // Replace with Stencil context when it is available.
```
denis-bc, line 11: Are there any limitations on how big stencil context can be? I assume for orders and their returns it will be quite chunky blob 🙂

P.S.
Is it the only way how we can pass this data to the template?

package-lock.json
bc-bryan: `npm install` bumps it for me locally as well, likely just matching the `bigcommerce-cornerstone` package version.

assets/js/theme/add-return-new.js
jpajar: Yeah that's the way we pass data onto templates in the context of cornerstone unfortunately 😅 

is that correct @bc-bryan ?

assets/js/theme/add-return-new.js
jpajar: Pretty much, and at the end of the day if we can structure the Stencil Context similar to other templates, we could cut down on what we have to manipulate via private functions if we want as well.

matt-evangelidis (approved):
Seems fine for merging to the feature branch

PR #2673, opened 2026-06-04
Orders 7764 render create return page with context

bc-vivekaggarwal:
#### What?

Massive refactoring for new return flow to use updated stencil page context(order details, reasons and resolutions) and use the new /create-return route

#### Requirements

#### Tickets / Documentation

Jira: [ORDERS-7764](https://bigcommercecloud.atlassian.net/browse/ORDERS-7764)

All pages load fine:

###Orders listing page:
image

###Order details page: 
image

###Add returns page:

#### When both flags(legacy `returns_enabled` and new `returns_v2_enabled`) off:
Orders list and Order details page don't show returns items and new_return route redirects to orders list page:
image
image

#### Existing add Return page (when only legacy returns_enabled):
image

#### Add return page when only new flag `returns_v2_enabled` enabled:
Redirects to orders list page. Currently BC app doesn't support redirection when legacy returns is disabled. We need to fix this as part of [ORDERS-7825](https://bigcommercecloud.atlassian.net/browse/ORDERS-7825)

#### Add Return page with new content(when returns_v2_enabled or both enabled):
image

New Design after consultation with JW and JH:
image

[ORDERS-7764]: https://bigcommercecloud.atlassian.net/browse/ORDERS-7764?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Changes which account routes load which return UI and how v2 is reached relative to feature flags; submission is not wired yet, but mis-routing could affect shoppers on the wrong template.
> 
> **Overview**
> Moves the **returns v2** UI off `account/add-return` onto a dedicated **`create-return`** route and page type, wired to stencil context (`order`, per-line `returnableQuantity`/`price`, and shared `reasons`/`resolutions`).
> 
> **Routing & JS:** `app.js` loads `create-return.js` for `create_return` and routes `account_new_return` back to the generic account bundle. `add-return-new.js` is removed in favor of `CreateReturn`, which validates submit from **select** controls (qty/resolution/reason), drops the quantity stepper, and exposes `buildReturnInput()` for the upcoming Storefront GQL submit (still a TODO).
> 
> **Templates:** `add-return-new.html` is deleted; `add-return.html` no longer branches on `returns_v2_enabled` and only renders the **legacy** POST return form (with small a11y/table copy tweaks). `create-return.html` implements the new layout: column headers on large screens, qty dropdowns capped by `returnableQuantity`, and dynamic reason/resolution options.
> 
> **Styles & copy:** `.newReturn` styles move from `addReturn` into a new `createReturn` SCSS component (header buttons, column alignment, select widths). `en.json` adds/updates strings for the new return request UX.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4ee19ec39b80bf89163db75e282c47b2c8247ba4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

[ORDERS-7825]: https://bigcommercecloud.atlassian.net/browse/ORDERS-7825?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bc-vivekaggarwal:
cursor review

matt-evangelidis:
>Add return page when only new flag returns_v2_enabled enabled:
Redirects to orders list page. Currently BC app doesn't support redirection when legacy returns is disabled. We need to fix this as part of ORDERS-7764

For context, what change do we need to make on the BCApp side to get redirection to work?

assets/js/theme/create-return.js
```diff
@@ -0,0 +1,72 @@
+import PageManager from './page-manager';
+
+export default class CreateReturn extends PageManager {
+    onReady() {
+        const $form = $('[data-new-return-form]');
```
BC-SEven: Oh we have jquery in cornerstone! 😅 

Im not really sure what our position is with jquery. But from my own experience I thinks its better to use native browser functionality over jquery where possible. Could we just use `document. querySelector()` here instead?

templates/pages/account/add-return.html
```diff
@@ -1,118 +1,106 @@
...
+            {{#if forms.return.order_products.length}}
+                <form
+                    action="{{urls.account.orders.save_new_return}}"
+                    class="form"
+                    method="post"
+                    data-account-return-form-error="{{lang 'account.returns.error_no_qty'}}"
+                    data-account-return-form
+                >
+                    <fieldset class="form-fieldset">
+                        <input type="hidden" name="order_id" value="{{forms.return.order_id}}">
```
BC-SEven, line 22: 🍹 Would it be better to just add order_id into the payload via js. Happy to keep this as a hidden input tho if its just simpler

templates/pages/account/add-return.html
```diff
@@ -1,118 +1,106 @@
...
+                        <table class="table table--line">
```
BC-SEven: Should we include a `<caption>` in the table for accessibility?

https://www.w3.org/WAI/tutorials/tables/caption-summary/

Either that or we should at-least include an `aria-description`

bc-vivekaggarwal:
> > Add return page when only new flag returns_v2_enabled enabled:
> > Redirects to orders list page. Currently BC app doesn't support redirection when legacy returns is disabled. We need to fix this as part of ORDERS-7764
> 
> For context, what change do we need to make on the BCApp side to get redirection to work?

@matt-evangelidis Good catch, updated description. The redirection fix is dependant on: [ORDERS-7825](https://bigcommercecloud.atlassian.net/browse/ORDERS-7825)
Everything else is already there:
1. returns_v2_enabled already works now. But it comes in settings context which is currently behind STRF-13960.Settings_resource_migration. Either you turn it ON for your store or hack it in Storefront to bypass that FF.
2. Redirection: Currently all GraphQL order related APIs return only legacy URL in API response, we need to update that as part of ORDERS-7825 to the legacy/new URL based on which returns flow is enabled. Meanwhile, use hardcoded URL /create-return/{orderId} to land on the page

[ORDERS-7825]: https://bigcommercecloud.atlassian.net/browse/ORDERS-7825?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

templates/pages/account/add-return.html
```diff
@@ -1,118 +1,106 @@
...
-                                        <th class="return-itemName">{{lang 'account.orders.return.order_item'}}</th>
-                                        <th class="return-itemPrice">{{lang 'account.orders.return.item_price'}}</th>
-                                        <th class="return-itemQuantity">{{lang 'account.orders.return.quantity'}}</th>
+                                        <td class="return-itemName">
+                                            <span class="return-itemTitle">{{name}}</span>
+                                            {{#if options}}
+                                            <dl class="definitionList">
+                                                {{#each options}}
+                                                <dt class="definitionList-label">{{name}}:</dt>
+                                                <dd class="definitionList-description">{{{sanitize value}}}</dd>
```
BC-SEven, line 40: For security I believe this is still safer: 
```suggestion
                                                <dd class="definitionList-description">{{value}}</dd>
```
Just stripping the tags is still easy to get around if we don't escape the string properly 
https://dev.to/rodrigojavornik/xss-attack-why-striptags-is-not-enough-5gmo

templates/pages/account/add-return.html
```diff
@@ -1,118 +1,106 @@
...
+                                                {{/each}}
+                                            </dl>
+                                            {{/if}}
+                                        </td>
+                                        <td class="return-itemPrice">{{price.formatted}}</td>
+                                        <td class="return-itemQuantity">
+                                            <label class="form-label">{{lang 'account.orders.return.quantity'}}:</label>
```
BC-SEven: For accessibility please include a `for="select-input-id"` in the label

templates/pages/account/add-return.html
```diff
@@ -1,118 +1,106 @@
...
+                        </table>
+                    </fieldset>
+                    <div class="account--addReturn-row">
+                        <fieldset class="form-fieldset account--addReturn-column">
+                            <div class="form-field">
+                                <label for="return_reason" class="form-label">
+                                    {{lang 'account.orders.return.reason'}}
+                                    <small>{{lang 'common.required'}}</small>
+                                </label>
+                                <select name="return_reason" id="return_reason" class="form-select">
```
BC-SEven, line 66: Since this is a giant loop I think we need to make the `id` for this select unique on the page.  Can we append the `index` to every id in here?

templates/pages/account/add-return.html
bc-vivekaggarwal: {{{sanitize value}}} is a general way in this repo. And the link rather tell why strip_tags is not enough

templates/pages/account/add-return.html
bc-vivekaggarwal: Legacy returns file, not part of recent changes. Good to fix. Done.

templates/pages/account/add-return.html
bc-vivekaggarwal: Legacy returns file, not part of recent changes. Good to fix. Done.

templates/pages/account/add-return.html
bc-vivekaggarwal: Also, Legacy returns file, not part of recent changes. Just reverted to existing changes here.

templates/pages/account/add-return.html
bc-vivekaggarwal: Legacy returns file, not part of recent changes. I was just reverting previous changes. Good to fix. Done.

templates/pages/account/add-return.html
bc-vivekaggarwal: Legacy returns file, not part of recent changes. I was just reverting previous changes. Good to fix. Done.

assets/js/theme/create-return.js
bc-vivekaggarwal: Done.

templates/pages/create-return.html
```diff
@@ -1,9 +1,110 @@
...
+                        lazyload=../../theme_settings.lazyload_mode
+                        default_image=../../theme_settings.default_image_product
+                    }}
+                </div>
+
+                <div class="newReturn-orderLineItemInfo">
+                    <p class="newReturn-orderLineItemName">{{name}}</p>
+                    {{#if productAttributesList}}
+                        <p class="newReturn-orderLineItemVariant">
+                            {{#each productAttributesList}}{{name}}: {{{sanitize value}}}{{#unless @last}} &middot; {{/unless}}{{/each}}
```
matt-evangelidis: Is everything in `productAttributesList` merchant-controlled? If so, is there any reason to sanitize the value here?

templates/pages/create-return.html
```diff
@@ -1,9 +1,110 @@
...
+                            <option value="{{$index}}">{{$index}}</option>
+                        {{/for}}
+                    </select>
+                    <select class="form-select form-select--small" id="resolution-{{id}}" aria-label="{{lang 'account.returns.select_resolution'}}">
+                        <option value="">{{lang 'account.returns.select_resolution'}}</option>
+                        {{#each ../../resolutions}}
+                            {{#if active}}
+                                <option value="{{resolutionType}}">{{label}}</option>
+                            {{/if}}
+                        {{/each}}
```
matt-evangelidis, line 76: For resolutions and reasons here, is there a particular reason we're doing the active filtering here rather than filtering that earlier on?

templates/pages/create-return.html
BC-SEven:
```suggestion
                            {{#each productAttributesList}}{{name}}: {{ value }}{{#unless @last}} &middot; {{/unless}}{{/each}}
```
Please avoid usage of `sanitize`.  All it does is strip html tags. Which isn't sufficient to prevent injection vulnerabilities

templates/pages/create-return.html
BC-SEven: I suppose it gives merchants freedom to still display disabled options if they wish? 😅

templates/pages/create-return.html
```diff
@@ -1,9 +1,110 @@
...
+                    </select>
+                    <select class="form-select form-select--small" id="reason-{{id}}" aria-label="{{lang 'account.returns.select_reason'}}">
+                        <option value="">{{lang 'account.returns.select_reason'}}</option>
+                        {{#each ../../reasons}}
+                            {{#if active}}
+                                <option value="{{id}}">{{nameForMerchant}}</option>
```
BC-SEven: `nameForMerchant` is only intended for CP and admin apis. We shouldn't be exposing this in our storefront apis. Storefront should always use `nameForShopper`

templates/pages/create-return.html
bc-vivekaggarwal: Good call — switching to {{value}}

templates/pages/create-return.html
bc-vivekaggarwal: I'll move it Storefront. I don't think merchants would need it later.

bc-vivekaggarwal: Before:
image


After:
image

Same with resolutins as well

BC-SEven (approved):

PR #2687, opened 2026-06-23
feat(checkout): show backorder prompts for bundled items in the my account->order detail page

bc-shawnwang:
#### What?

- Display backorder prompts (`quantity_ready_to_ship`, `quantity_backordered`, `backorder_message`) for bundled picklist options on the account order details page, using the new `linked_bundled_item` field on each option
- Switch parent item ready-to-ship quantity to use the new `quantity_ready_to_ship` field directly instead of computing it via subtraction

Some minor styling/design changes:
- Move parent item backorder prompts to appear directly below the product name, above options
- Reduce backorder prompt font size from `smaller` to `smallest` for both parent and bundled items
- Add spacing between parent backorder prompts and the options list, and between each bundled item's backorder prompts and the next option row

> [!NOTE]
> All gating and show/hide logic for backorder fields (quantity_backordered, quantity_ready_to_ship, backorder_message, linked_bundled_item) is handled server-side; the theme renders fields purely based on their presence in the data, no need of extra gating.  

#### Requirements

#### Tickets / Documentation

- [BACK-692](https://bigcommercecloud.atlassian.net/browse/BACK-692) [FE] Render backorder prompts for picklist items on My orders

Screenshot 2026-06-23 at 6 32 55 PM

Screenshot 2026-06-23 at 6 47 03 PM

[BACK-692]: https://bigcommercecloud.atlassian.net/browse/BACK-692?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Low Risk**
> Template and SCSS-only changes that render optional order fields; no checkout, auth, or client-side inventory logic.
> 
> **Overview**
> **My Account order details** now surfaces backorder copy for **picklist bundled options**, not just the parent line item.
> 
> Parent prompts use **`quantity_ready_to_ship`**, **`quantity_backordered`**, and **`backorder_message`** when the API sends them (replacing the old block that only showed when `quantity_backordered > 0` and derived on-hand from `quantity - quantity_backordered`). Those prompts sit **under the product title**, before the options list. Each option can show the same three fields from **`linked_bundled_item`** under its value.
> 
> **Styling:** backorder text uses **`fontSize("smallest")`**, with extra margin before the options **`definitionList`** and before the next option row when prompts precede **`dt`/`dd`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bd424a7fbdeef75bf6cc0943d2585ef3ffc0c991. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Draft
 - Validate picklist option `available_to_sell` on PDP so Add to Cart is disabled and a "maximum purchasable quantity for {picklist} is {qty}" error is shown when the requested quantity exceeds a selected picklist item's stock, even when the main product still has enough stock [#2684](https://github.com/bigcommerce/cornerstone/pull/2684)
+- Render backorder prompts for bundled picklist options on the account order details page, using `linked_bundled_item` data per option; move parent item backorder prompts above the options list; reduce backorder prompt font size and add spacing between prompts and adjacent option rows [#2687](https://github.com/bigcommerce/cornerstone/pull/2687)
 - Hide option values on the PDP that would complete a "disable and hide" rule for the current selection, and steer the shopper off a default selection that is itself a forbidden combination [#2678](https://github.com/bigcommerce/cornerstone/pull/2678)
 - Refine backorder layout and colors on the account order details page [#2677](https://github.com/bigcommerce/cornerstone/pull/2677)
 - Make the cart Shipping row label span the full width of the totals row so the shipping expectation prompt is no longer constrained to the narrow label column
--- a/assets/scss/components/stencil/account/_account.scss
+++ b/assets/scss/components/stencil/account/_account.scss
@@ -155,8 +155,17 @@ $account-backorder-textColor: #757575;
 
 .account-product-backorder-prompts {
     color: $account-backorder-textColor;
-    font-size: fontSize("smaller");
+    font-size: fontSize("smallest");
     margin-bottom: 0;
+
+    + .definitionList {
+        margin-top: spacing("half");
+    }
+
+    + dt,
+    + dt + dd {
+        margin-top: spacing("half");
+    }
 }
 
 .account-order-backorder-expectation {
--- a/templates/components/account/order-contents.html
+++ b/templates/components/account/order-contents.html
@@ -54,11 +54,29 @@
                 {{#if brand}}
                     <h6>{{brand.name}}</h6>
                 {{/if}}
+                {{#if quantity_ready_to_ship}}
+                    <p class="account-product-backorder-prompts">{{lang 'products.quantity_on_hand' quantity=quantity_ready_to_ship}}</p>
+                {{/if}}
+                {{#if quantity_backordered}}
+                    <p class="account-product-backorder-prompts">{{lang 'products.quantity_backordered' quantity=quantity_backordered}}</p>
+                {{/if}}
+                {{#if backorder_message}}
+                    <p class="account-product-backorder-prompts">{{backorder_message}}</p>
+                {{/if}}
                 {{#if options}}
                 <dl class="definitionList">
                     {{#each options}}
                         <dt class="definitionList-key">{{name}}:</dt>
                         <dd class="definitionList-value">{{> components/common/product-options}}</dd>
+                        {{#if linked_bundled_item.quantity_ready_to_ship}}
+                            <dd class="definitionList-value account-product-backorder-prompts">{{lang 'products.quantity_on_hand' quantity=linked_bundled_item.quantity_ready_to_ship}}</dd>
+                        {{/if}}
+                        {{#if linked_bundled_item.quantity_backordered}}
+                            <dd class="definitionList-value account-product-backorder-prompts">{{lang 'products.quantity_backordered' quantity=linked_bundled_item.quantity_backordered}}</dd>
+                        {{/if}}
+                        {{#if linked_bundled_item.backorder_message}}
+                            <dd class="definitionList-value account-product-backorder-prompts">{{linked_bundled_item.backorder_message}}</dd>
+                        {{/if}}
                     {{/each}}
                 </dl>
                 {{/if}}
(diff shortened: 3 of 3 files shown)


bc-shawnwang:
Hi @JessicaHu51 

The design intends for the main product's backorder prompts to appear below all regular options and above all picklist options.

image

However, this is difficult to implement because it requires breaking the original option ordering. Picklist options are still product options and exist at the same level as all other options. to match the original design, we need to group all options to two parts based on if they are picklist options, and first display:
- all non-picklist options 
- main product's backorder prompts if exists
- all picklist options

Unlike Checkout JS, where ordering can be adjusted more easily on the client side, this template is rendered server-side, so any reordering and grouping would need to happen on the backend.

More details of the tech problem:  

The template engine only supports basic flow control (if/else and for loops), so to accommodate the new design, the server would need to restructure the stencil data from a single options array into two separate arrays: non_picklist_options and picklist_options.
This approach has two significant problems. First, it's a messy data structure change driven purely by a UI rendering constraint rather than any logical separation of the data. Second, and more critically, this is a breaking change — themes are customizable, and any custom theme currently iterating over options would break if that field were split or removed.
The only workaround would be to keep all three fields in the payload simultaneously (options, non_picklist_options, and picklist_options), which is redundant, harder to maintain, and even messier than the original problem.

-----

Even putting the technical challenge aside, I see a couple of issues with the design:

**Issue 1: Visibility**

If a product has many options, the main product's backorder prompts may end up far below the product name and quantity, making them easy to miss.

For example:

```
5 × Sample

Color: xxx
Size: xxx
Custom 1: xxx
Custom 2: xxx
...
Custom N: xxx

4 ready to ship
1 will be backordered
```

The backorder information is pushed too far down the page. 

**Issue 2: Option ordering no longer matches the merchant's configuration**

This is the more serious concern.

Suppose the merchant configures the options in this order:

* Custom Message
* Bundle 1
* Bundle 2
* Include Insurance?

image

> [!CAUTION]
> Merchants can intentionally arrange options by dragging them up and down in Product settings UI so that means the ordering is meaningful and sensitive (they don't really always want picklist options all under non-picklist options). Some merchants really want the options to be displayed like the ordering they setup. At least before the backorder changes, from PDP -> cart/checkout -> order they always follow it.

With the original backorder design, we would break the ordering and render them like this:

```
Custom Message
Include Insurance?

(main item backorder prompts)

Bundle 1
(bundle item backorder prompts)

Bundle 2
(bundle item backorder prompts)
```

This changes the original ordering defined by the merchant.

image

I think we should respect the merchant's configured order rather than rearranging them.

So in my change,   I display the main product's backorder prompts directly below the product quantity and name, before all options, for example:

```
5 × Sample

4 ready to ship
1 will be backordered

Color: xxx
Size: xxx
...
```

image

This keeps the message highly visible and avoids modifying the merchant's option ordering.

WDYT? 

---
FYI, the PDP/manual order and order emails don't have this issue, because for those @JessicaHu51 you have another design idea, to display parent + bundled items' backorder prompts under qty independently, not within options list.

image

JessicaHu51:
> Hi @JessicaHu51
> 
> The design intends for the main product's backorder prompts to appear below all regular options and above all picklist options.
> 
> image
> However, this is difficult to implement because it requires breaking the original option ordering. Picklist options are still product options and exist at the same level as all other options. to match the original design, we need to group all options to two parts based on if they are picklist options, and first display:
> 
> * all non-picklist options
> * main product's backorder prompts if exists
> * all picklist options
> 
> Unlike Checkout JS, where ordering can be adjusted more easily on the client side, this template is rendered server-side, so any reordering and grouping would need to happen on the backend.
> 
> More details of the tech problem:
> 
> The template engine only supports basic flow control (if/else and for loops), so to accommodate the new design, the server would need to restructure the stencil data from a single options array into two separate arrays: non_picklist_options and picklist_options. This approach has two significant problems. First, it's a messy data structure change driven purely by a UI rendering constraint rather than any logical separation of the data. Second, and more critically, this is a breaking change — themes are customizable, and any custom theme currently iterating over options would break if that field were split or removed. The only workaround would be to keep all three fields in the payload simultaneously (options, non_picklist_options, and picklist_options), which is redundant, harder to maintain, and even messier than the original problem.
> 
> Even putting the technical challenge aside, I see a couple of issues with the design:
> 
> **Issue 1: Visibility**
> 
> If a product has many options, the main product's backorder prompts may end up far below the product name and quantity, making them easy to miss.
> 
> For example:
> 
> ```
> 5 × Sample
> 
> Color: xxx
> Size: xxx
> Custom 1: xxx
> Custom 2: xxx
> ...
> Custom N: xxx
> 
> 4 ready to ship
> 1 will be backordered
> ```
> 
> The backorder information is pushed too far down the page.
> 
> **Issue 2: Option ordering no longer matches the merchant's configuration**
> 
> This is the more serious concern.
> 
> Suppose the merchant configures the options in this order:
> 
> * Custom Message
> * Bundle 1
> * Bundle 2
> * Include Insurance?
> 
> image
> Caution
> 
> Merchants can intentionally arrange options by dragging them up and down in Product settings UI so that means the ordering is meaningful and sensitive (they don't really always want picklist options all under non-picklist options). Some merchants really want the options to be displayed like the ordering they setup. At least before the backorder changes, from PDP -> cart/checkout -> order they always follow it.
> 
> With the original backorder design, we would break the ordering and render them like this:
> 
> ```
> Custom Message
> Include Insurance?
> 
> (main item backorder prompts)
> 
> Bundle 1
> (bundle item backorder prompts)
> 
> Bundle 2
> (bundle item backorder prompts)
> ```
> 
> This changes the original ordering defined by the merchant.
> 
> image
> I think we should respect the merchant's configured order rather than rearranging them.
> 
> So in my change, I display the main product's backorder prompts directly below the product quantity and name, before all options, for example:
> 
> ```
> 5 × Sample
> 
> 4 ready to ship
> 1 will be backordered
> 
> Color: xxx
> Size: xxx
> ...
> ```
> 
> image
> This keeps the message highly visible and avoids modifying the merchant's option ordering.
> 
> WDYT?
> 
> FYI, the PDP/manual order and order emails don't have this issue, because for those @JessicaHu51 you have another design idea, to display parent + bundled items' backorder prompts under qty independently, not within options list.
> 
> image

thanks @bc-shawnwang for the detailed context. I'm aligned with the proposed approach to display in the order of:
- product name
- sku (if applicable)
- backorder info
- options in the order of merchant's choice

animesh1987 (approved):

bc-yevhenii-buliuk (approved):
LGTM!

PR #2688, opened 2026-06-24
feat(orders): ORDERS-7681 add Cancel return button to return-detailsjs page

jpajar:
#### What?

This PR introduces the ability for shoppers to Cancel a return that is in `OPEN` status via the new shopper facing returns portal return details page

#### Requirements

#### Tickets / Documentation

https://bigcommercecloud.atlassian.net/browse/ORDERS-7681

https://github.com/user-attachments/assets/1064bc34-ad2a-4cee-b224-5afb6d32b532

The return ends up cancelled
image

When the cancelation errors

https://github.com/user-attachments/assets/9f84edd0-3cf5-4335-b014-e7feaffa6cb5

---

> [!NOTE]
> **Medium Risk**
> Mutates order return state through authenticated GraphQL; follows the same pattern as create-return but incorrect authorization or status handling could let shoppers cancel returns they should not.
> 
> **Overview**
> Shoppers with **OPEN** returns can **cancel** them from the account return-details page instead of only viewing static details.
> 
> Return detail routes (`account_return_detail`, `return_detail`) now load a dedicated **`return-details`** module (extending account behavior) that wires a sidebar **Cancel return** button to a confirmation modal, then calls the Storefront GraphQL **`cancelReturn`** mutation using the injected API token and `return_id` from the URL. Success reloads the page; failures show a page-level error with loading/disabled state while the request runs.
> 
> The template adds the cancel action (OPEN only), loading overlay, error region, and context injections; **en.json** adds cancel copy; SCSS styles the action block and error alert.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5b4f55fb337eceda715d44bc45433e3b2579dbfa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/assets/js/theme/return-details.js
+++ b/assets/js/theme/return-details.js
@@ -1,8 +1,12 @@
-import PageManager from './page-manager';
+import Account from './account';
 import { showAlertModal } from './global/modal';
 
-export default class ReturnDetails extends PageManager {
+export default class ReturnDetails extends Account {
     onReady() {
+        // Run the shared account-page behavior (this page is mapped to this module
+        // instead of Account in app.js).
+        super.onReady();
+
         this.root = document.querySelector('[data-return-details]');
         if (!this.root) return;
 
@@ -10,9 +14,12 @@ export default class ReturnDetails extends PageManager {
         const cancelBtn = this.root.querySelector('[data-return-cancel]');
         if (!cancelBtn) return;
 
-        // The return's entity ID is the `return_id` query param the page is reached with
-        // (/account.php?action=view_return&return_id={id}).
+        // The return's ID is the `return_id` query param the page is reached with
+        // (/account.php?action=view_return&return_id={id}). It's a UUID string.
         this.returnId = new URLSearchParams(window.location.search).get('return_id');
+
+        // Without a return ID the cancel mutation can't run, so leave the button inert.
+        if (!this.returnId) return;
 
         cancelBtn.addEventListener('click', () => this.openCancelModal(cancelBtn));
     }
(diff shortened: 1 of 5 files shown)


assets/js/theme/return-details.js
```diff
@@ -0,0 +1,124 @@
...
+        // Run the shared account-page behavior (this page is mapped to this module
+        // instead of Account in app.js).
+        super.onReady();
+
+        this.root = document.querySelector('[data-return-details]');
+        if (!this.root) return;
+
+        // The cancel action is only server-rendered for open returns.
+        const cancelBtn = this.root.querySelector('[data-return-cancel]');
+        if (!cancelBtn) return;
```
matt-evangelidis, line 15: Just to refresh my memory, what do the early returns here do if I open this page and it returns early without setting up the `returnId` and without adding the event listener to the cancel button?

assets/js/theme/return-details.js
```diff
@@ -0,0 +1,124 @@
...
+        this.clearError();
+
+        try {
+            const response = await this.cancelReturn(this.returnId);
+            const errorMessages = this.getErrorMessages(response);
+
+            if (errorMessages.length) {
+                this.showError(errorMessages.join(' '));
+                return;
+            }
```
matt-evangelidis: This should probably be an actual length check rather than just the presence of length, given that any calls to `join`, even with an empty array, will always return a string, even if it's just empty.

assets/js/theme/return-details.js
```diff
@@ -0,0 +1,124 @@
...
+
+        const messageEl = errorBox.querySelector('[data-return-error-message]');
+        if (messageEl) messageEl.textContent = message || '';
+        errorBox.style.display = '';
+    }
+
+    clearError() {
+        const errorBox = document.getElementById('return-cancel-error');
+        if (errorBox) errorBox.style.display = 'none';
+    }
```
matt-evangelidis, line 87: Should we also wipe the actual content of the error message to prevent anything on the page from picking up on it (as it still exists in the DOM and all that)?

jpajar: Yeah thats a fair concern I can do that

assets/js/theme/return-details.js
jpajar: With the guards it will never try to attach an event listener if the cancel button doesn't exist (for non-open returns), and that handler is the only thing that uses the returnId so it should be fine I think 🤔

bc-bryan (approved):

denis-bc:
cursor review

matt-evangelidis (approved):

PR #2735, opened 2026-09-02
fix(storefront): TRAC-1466 Show option-set rule message on rule-blocked options

bc-yevhenii-buliuk:
#### What?
Cornerstone versions after 6.19.0 stopped showing an option-set rule's custom "unavailable" message (e.g. "This Size Temporarily Sold Out"), even though Add to Cart and the quantity input still correctly
  disabled. This affects legacy-catalog (v2) products with no SKUs, an option-set rule set to "unavailable for purchase" with a custom message, and product-level inventory tracking off.

**Root cause:** in product-details-base.js, updateView() called showMessageBox() (which renders the rule's message) before updateDefaultAttributesForOOS(). The latter calls toggleSoldOutAlert(), which
  unconditionally hides that same message box whenever the product itself is sellable (instock: true) - regardless of whether the specific option combination is blocked by a rule (purchasable: false). The backend
  sends exactly this combination (instock: true + purchasable: false) when inventory isn't tracked at the product level, so the rule's message was shown and then immediately hidden on every option change.

**Fixes:**
  1. Moved the showMessageBox() call to run after updateDefaultAttributesForOOS(), so the rule message is set last and isn't overwritten by the stock check.
  2. showMessageBox() now clears a stale data-qty-limit flag when it sets its own message, so a later quantity-limit check (updateAddToCartForQty) doesn't mistake this message for its own and delete it.
  3. updateAddToCartForQty() now skips overwriting the box with a "maximum purchasable quantity" message when Add to Cart/quantity are already disabled for a more fundamental reason (a rule or true out-of-stock) - a customer who can't buy an item at all shouldn't see a quantity-limit message instead of the actual reason. This check is based on the button's disabled state (set earlier by the OOS/rule check), so it stays correct for backorderable items that can still show a stale stock_message while remaining purchasable (see BACK-633 / commit 083e45f4).

**Testing:**
  Verified both with unit tests and manually on two live stores, since option-set rules (no-SKU products) and backorder limits (SKU-level) can't coexist on the same product:
  - Legacy v2 catalog store, rule-blocked option: the custom rule message now shows and stays visible (**fix 1**).
  - Backorder-enabled store, switching between a qty-limited variant and an out-of-stock/unavailable variant with a leftover quantity in the field: the correct message (rule/unavailable/OOS) is preserved instead of being wiped by a stale data-qty-limit flag (**fix 2**), and instead of being overwritten by a "maximum purchasable quantity" message when the variant is separately marked unpurchasable (**fix 3**).
  - Confirmed no regression for normal backorder quantity limits (over-limit quantity on a purchasable backorderable variant still shows the correct "maximum purchasable quantity" message and disables Add to Cart).

#### Requirements

#### Tickets / Documentation

- Jira: [TRAC-1466](https://bigcommercecloud.atlassian.net/browse/TRAC-1466)

**FIX 1**
Tested using bundle with the fixes in staff sandbox that has legacy product catalogue (v2)

Screenshot 2026-09-02 at 14 35 12

**before:**

https://github.com/user-attachments/assets/2e2bbe7c-a091-415f-bd75-bee21726f064

**after:**

https://github.com/user-attachments/assets/862b33ed-9261-4232-8ce5-1e22e2d320f1

---

**FIX 2**

Screenshot 2026-09-02 at 20 49 14
Screenshot 2026-09-02 at 20 49 41

**before:**

https://github.com/user-attachments/assets/6bd0fe4c-a4e4-4d45-9789-15e1b9c1a99d

**after:**

https://github.com/user-attachments/assets/1ac5a1d3-385a-4991-bed6-20e40fae6b2b

---

**FIX 3**

Screenshot 2026-09-02 at 20 47 34
Screenshot 2026-09-02 at 20 47 13

**before:**

https://github.com/user-attachments/assets/3cabceb7-b850-45eb-8558-e8973e913bb1

**after:**

https://github.com/user-attachments/assets/eef05b33-2348-4fa6-846b-01fb114d65a2

[TRAC-1466]: https://bigcommercecloud.atlassian.net/browse/TRAC-1466?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Touches PDP option-change messaging and add-to-cart/qty-limit ordering; regression risk for backorder quantity limits and OOS flows, though changes are narrowly scoped to message box timing and flags.
> 
> **Overview**
> Restores **custom option-set rule messages** on the PDP when a combination is unpurchasable but the product still looks sellable (e.g. legacy v2, no SKUs, inventory not tracked at product level). Add to Cart and quantity were already disabled; the alert text was missing or wrong.
> 
> In `product-details-base.js`, **`showMessageBox`** now runs **after** `updateDefaultAttributesForOOS`, so `toggleSoldOutAlert` no longer hides the variant alert whenever `instock` / ATS says the product can sell while the selected options are rule-blocked.
> 
> **`showMessageBox`** clears **`data-qty-limit`** when it sets a rule/stock message so **`updateAddToCartForQty`** does not treat that box as its own and clear or replace it.
> 
> **`updateAddToCartForQty`** skips the “maximum purchasable quantity” message when quantity controls are already disabled (rule or true OOS); it only keeps Add to Cart disabled instead of swapping the shopper’s real reason for a qty-cap message.
> 
> CHANGELOG Draft entry added for [#2735].
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 781a01418ac781e78f4f22418bd9ba909da1e3e1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/assets/js/theme/common/product-details-base.js
+++ b/assets/js/theme/common/product-details-base.js
@@ -502,10 +502,18 @@ export default class ProductDetailsBase {
             ? Infinity
             : parseInt(this.context.availableToSell, 10) || 0;
 
+        // If increment buttons are already disabled for a reason unrelated to quantity
+        // (e.g. out-of-stock), don't override that with a quantity limit.
+        const alreadyBlocked = viewModel.$increments.prop('disabled');
+
         // The main product's sell-limit takes priority over any picklist limit.
         if (availableToSell > 0 && qty > availableToSell) {
-            const template = this.context.quantityMaxMessage || 'The maximum purchasable quantity is __QTY__';
-            this.showQtyLimitMessage(viewModel, $variantMessage, template.replace('__QTY__', availableToSell));
+            if (alreadyBlocked) {
+                viewModel.$addToCart.prop('disabled', true);
+            } else {
+                const template = this.context.quantityMaxMessage || 'The maximum purchasable quantity is __QTY__';
+                this.showQtyLimitMessage(viewModel, $variantMessage, template.replace('__QTY__', availableToSell));
+            }
             return;
         }
 
@@ -515,12 +523,16 @@ export default class ProductDetailsBase {
             && this.picklistBackorder.getSellLimitViolation(qty);
 
         if (picklistLimit) {
-            const template = this.context.quantityMaxPicklistMessage
-                || 'The maximum purchasable quantity for __NAME__ is __QTY__';
-            const message = template
-                .replace('__NAME__', picklistLimit.name)
-                .replace('__QTY__', picklistLimit.availableToSell);
-            this.showQtyLimitMessage(viewModel, $variantMessage, message);
+            if (alreadyBlocked) {
+                viewModel.$addToCart.prop('disabled', true);
+            } else {
+                const template = this.context.quantityMaxPicklistMessage
+                    || 'The maximum purchasable quantity for __NAME__ is __QTY__';
+                const message = template
+                    .replace('__NAME__', picklistLimit.name)
+                    .replace('__QTY__', picklistLimit.availableToSell);
+                this.showQtyLimitMessage(viewModel, $variantMessage, message);
+            }
             return;
         }
 
@@ -607,8 +619,6 @@ export default class ProductDetailsBase {
     updateView(data, content = null) {
         const viewModel = this.getViewModel(this.$scope);
 
-        this.showMessageBox(data.stock_message || data.purchasing_message);
-
         if (data.price instanceof Object) {
             this.updatePriceView(viewModel, data.price);
         } else {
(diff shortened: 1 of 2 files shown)


assets/js/theme/common/product-details-base.js
```diff
@@ -502,10 +502,16 @@ export default class ProductDetailsBase {
             ? Infinity
             : parseInt(this.context.availableToSell, 10) || 0;
 
+        // If another message is already showing, something more important already
```
jairo-bc: what's "`something more important`"?

bc-yevhenii-buliuk: Rephrased the comment to spell this out directly.
**The idea is**: this banner can already be showing a "Sold Out"-message (from an option-set rule or out-of-stock) - that's a higher-priority message than "you can buy it, just not in this quantity". So we shouldn't let the quantity-limit check overwrite a "can't purchase" banner with a "quantity limit" banner, I think the inability to purchase at all is more important to communicate than a quantity restriction.

jairo-bc (approved):

